PR #125, opened 2019-11-12
Oc 27 bis

youhou1515:
OC-27


src/docs/asciidoc/architecture/02_business_concepts/03_business_services/01_user.adoc
```diff
@@ -29,15 +29,14 @@ Description:: Create User
...
+ . Automation creation user when an user succeed in authenticating and connecting for the first time (See <<To Authenticate>>)
 
 .Alternative scenario
- . Administrator creates a user directly
+ . Administrator creates an user directly
 
 .Error scenario
- * If user is not registred in authentication provider when an Administrator creates it, raise an error
- * If user already exist when an Administrator creates it, raise an error
+ * If an user is not registered in authentication provider when an Administrator creates it, raise an error
```
rlg-rte: English language is tricky about the "a/an" usage and it seems that for nouns beginning with a "u" which sounds like "y" in you or a "o" which sounds like "w" in "won", a "a" should be used (cf [grammaticality - Is it "a user" or "an user" - English Language & Usage Stack Exchange](https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user))

tools/spring/spring-oauth2-utilities/src/main/java/org/lfenergy/operatorfabric/springtools/configuration/oauth/jwt/groups/roles/RoleClaimStandardList.java
```diff
@@ -0,0 +1,55 @@
...
+	public List<String> computeNodeElementRole(JsonNode jsonNodeElement) {
+		List<String> listGroupsResult = new ArrayList<>();
+		listGroupsResult.addAll(Arrays.asList((jsonNodeElement.asText()).split(separator)));	
+		return listGroupsResult;
+	}
+	
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append("RoleClaimStandardList(path="+path+", separator=" + separator+")");
```
rlg-rte: You might have used the `StringBuilder` for the whole composed String and maybe have something like:
```
sb.append("RoleClaimStandardList(path=);
sb.append(path);
sb.append(" separator=");
sb.append(  separator);
sb.append(")");
```

PR #127, opened 2019-11-14
[OC-536] reuse the color style of the theme-color

youhou1515:
OC-536


youhou1515:
the feature is to add a color pastil, not anymore the color text

PR #133, opened 2019-11-26
[OC-515] 

agigox:
Fixing Timeline initial view of W, M and Y and code refactoring


youhou1515:
Hi @agigox , (to @AlexGuironnetRTE  to @rlg-rte and to @vlo-rte )

there are some improvement (for my opinion) in the code, it needs to be discussed with the team on the fact that I see a structure repeated a lot of time.
This structure : 
  **: {
    year: 0,
    month: 0,
    week: 1,
    day: 0,
    hour: 0,
    minute: 0,
    second: 0,
  },** 

should be in one structure and call with it.

And you can also use the typescript functionality "..." as it is used in the @ngrx/store to change the attribut who need to be updated.

agigox:
Hello @youhou1515 

A very good point.

I created a function that accepts three parameters and return a moment object as discussed earlier in the day meeting

Amine,

PR #134, opened 2019-11-27
Oc 537

youhou1515:
*Please check if the PR fulfills these requirements* _(please use `&#39;[x]&#39;` to check the boxes, or submit the PR and then click the checkboxes)_

* [X] The commit message follows our guidelines
* [X] Tests for the changes have been added (for bug fixes / features)
* [X] Docs have been added / updated (for bug fixes / features)


ui/main/src/app/components/navbar/navbar.component.html
```diff
@@ -7,7 +7,7 @@
 <nav class="navbar navbar-expand-lg fixed-top navbar-dark bg-dark ">
   <a class="navbar-brand">
     <div *ngIf="this.customLogo; else opFabLogo">
-      <of-icon icon="custom-logo" medium [base64]="customLogo"></of-icon>
+      <of-icon icon="custom-logo" [limitSize]="limitSize" [height]="height" [width]="width" [base64]="customLogo"></of-icon>
```
AlexGuironnetRTE: Why do we need to pass "custom-logo" as icon and as base64 ?

AlexGuironnetRTE:
As discussed, we are integrating these changes because projects are waiting for this functionality for a production deployment but there are issues with it that must be addressed first thing in the next sprint (see OC-558):
Split icon component in two

Due to the developments for the custom logo (OC-537), the code from icon component is currently very complex due to the fact that it combines an “icon” mode and a “logo” mode that work quite differently (different sizing parameters, one takes its size from the html template, the other from config, one has a light/dark mode and not the other, etc.). The resulting code is difficult to read and will make maintenance difficult, it would be better to split this into two components “icon” and “logo”.

2. Default size for the logo is not really sensible

If property operatorfabric.logo.base64 is set to define a custom logo but operatorfabric.logo.height and operatorfabric.logo.width aren’t set, the default is a 16px by 16 px square, which is way too small.

32 px by 32 px (the size of the OperatorFabric icon displayed by default) would be more sensible in my opinion.

3. The limit doesn’t seem to be applied if limitSize is not set:

logo:  

base64:  take logo from unit tests

height: 100px

width: 400px  

limitSize: true

→ Result: Resulting logo is 400px wide and 100px high.

Expected result: 200px wide and height =64px 

(I said doesn’t seem because I thought it had worked the first time I tested).

4. The limits don’t work if only width or only height is set. For exemple:

logo:  

base64:  take logo from unit tests

width: 400px  

limitSize: true

→ Result: Resulting logo is 400px wide and about 91px high (apparently if it is not set the height is calculated so as to preserve the height/width ratio of the svg)

Expected result: 200px wide and height =64px 

Also, if the width or the height is above the limit, setting them to the max value without taking into account the width/height ratio distorts the logo.

We should either fix it or remove this property entirely.

5. Property values

When I tested, from the documentation I understood that width and height properties should be provided as plain numbers representing the value in pixels, for example: height: 200, when actually judging from the tests it is height: 200px that is expected. 

@All What would you expect?

The weird part is that it seemed to work if I used plain numbers but the limit didn’t work. Wouldn’t using plain numbers make the code simpler?

PR #136, opened 2019-11-29
[OC-554] Move timeline documentation to web ui reference doc

AlexGuironnetRTE:


AlexGuironnetRTE:
Checks failed because of the issue with the sonar token used by travis.

PR #139, opened 2019-12-12
new getting-started

freddidierRTE:
New getting started documentation . 

Script and configuration associated with this new getting started in the new git repository : https://github.com/opfab/operatorfabric-getting-started


HanaeSafiRTE:
Hi Frédéric 
Thanks for this new getting started. I've tested the 4 examples, all of them are working.
However I've a few remarks:
1) In the section **_2. Install and run server_** , I think it's better to be more specific about the 
2) In the section  **_3.2. Example 2 : Publish a new bundle_**  there is 2 typos in the following sentence: **The bundle is **provide** in the bundle directory of **example3**** . It's example2 rather than example3. And  "provided "rather than "provide"
3) In the section  **_3.2. Example 2 : Publish a new bundle_** there is another typo in the first sentence. Same as the remark 2 "provided "instead of "provide"
4) In the section **_3.2.2. Get a Token_** , I think that's it is better to precise that the file sendBundle.sh should be opened to set the value of "token"
5) And I've found to mi

PR #192, opened 2020-02-05
OC-579 - Simplify Docker image production process

AlexGuironnetRTE:
Now every release will create both latest and X.X.X.RELEASE images (the ci_latest keyword is no longer needed), while CRON jobs will create snapshot (with no version number) images.
The documentation has been updated accordingly.


vlo-rte:
Code review ok.

PR #196, opened 2020-02-10
Oc 549

JeroenGommans:
* [X] The commit message follows our guidelines
* [X] Tests for the changes have been added (for bug fixes / features)
* [X] Docs have been added / updated (for bug fixes / features)

This PR fixes OC-549. Now instead of an HTTP 200, a 404 is returned when a non-existing bundle is requested.


services/core/thirds/src/main/java/org/lfenergy/operatorfabric/thirds/controllers/ThirdsController.java
```diff
@@ -90,7 +90,14 @@ public ThirdsController(ThirdsService service) {
 
     @Override
     public Third getThird(HttpServletRequest request, HttpServletResponse response, @PathVariable String thirdName, String apiVersion) {
-        return service.fetch(thirdName, apiVersion);
+        Third third = service.fetch(thirdName, apiVersion);
+        if (third == null) {
+            throw new ApiErrorException(ApiError.builder()
+                    .status(HttpStatus.NOT_FOUND)
+                    .message(String.format("Third with name %s was not found", thirdName))
```
AlexGuironnetRTE: Everything looks good, and I do get the 404 status when requesting a third that doesn't exist, but not the message?

Here's what I get if I do a GET on http://localhost:2100/third/DoesntExist using PostMan:

```
{
    "timestamp": "2020-02-11T16:07:52.565+0000",
    "status": 404,
    "error": "Not Found",
    "message": "Not Found",
    "path": "/third/DoesntExist"
}
```
I can't figure out what that is... If I try another similar method and do a GET on : http://localhost:2103/users/doesntexist/settings

I do get the correct message:

```
{
    "status": "NOT_FOUND",
    "message": "User setting for user doesntexist not found"
}
```
I've compared the two but can't find the difference ?

JeroenGommans: Good find and interesting!
If I issue a curl it seems to work though:

`curl -v "http://localhost:2100/thirds/DoesNotExist" -H  "accept: application/json" -H "Authorization:Bearer $token"`

I get the following response (removed uninterested parts):

` < HTTP/1.1 404 Not Found`
<...>
`* Connection #0 to host localhost left intact
{"status":"NOT_FOUND","message":"Third with name DoesNotExist was not found"}`

I'll look into it and will repeat it using postman as well.

JeroenGommans: I think you made a mistake in the URL by omitting the last s in http://localhost:2100/thirds

AlexGuironnetRTE: Hi Jeroen, you're completely right, my mistake! Sorry!

PR #229, opened 2020-03-20
[OC-773] Making sound notifications manageable in settings

AlexGuironnetRTE:
I decided to have four separate settings rather than a list because it made the checkbox component more reusable.
Note: I had to change the names of the configuration properties due to settings constraints.


vlo-rte:
Code review OK.

PR #234, opened 2020-03-25
[OC-797] Correct bug in x axis display and refactoring

freddidierRTE:


AlexGuironnetRTE:
Tests of expectations described in issue (on Firefow & Chrome):
- Labels are displayed for all visible days
- On 29/03/2020, graduation shifts from 00 to 05 (instead of 04) -> not a bug, it's because we'll be switching to daylight saving time (maybe for the last time ;-) ).

This doesn't address the fact that following days have a 1h -> 5h graduation as long as the day on which daylight saving time starts is present in the timeline.

PR #266, opened 2020-04-20
[OC-737] Add an object entity in the user service

vlo-rte:
Please can you review this pull request ?
Thank you.


services/core/users/src/main/java/org/lfenergy/operatorfabric/users/controllers/EntitiesController.java
```diff
@@ -0,0 +1,215 @@
...
+        List<UserData> formerlyBelongs = userRepository.findByEntitiesContaining(id);
+        List<String> newUsersInEntity = new ArrayList<>(users);
+
+        //Make sure the intended updated users list only contains logins existing in the repository, throwing an error if this is not the case
+        retrieveUsers(users);
+
+        List<UserData> toUpdate = new ArrayList<>();
+        formerlyBelongs.stream()
+           .filter(u->!users.contains(u.getLogin()))
+           .forEach(u-> {
```
rlg-rte, line 171: It could may be possible to use `peek` method to putt all the actions perform upon the userData to update into one place and then collect it in a list in order to save it ? it would more compact and more functional.

vlo-rte: Code modified following your advice. Thank you Ronan.

PR #307, opened 2020-05-18
[OC-898]  First and last name not displayed in implicit mode

freddidierRTE:


AlexGuironnetRTE:
Tested with implicit flow, mode dev
Added first name and last name to user through user service
Added different first name and last name to user in Keycloak
Log in : First & last name from Keycloak are displayed
Log out, change name in Keycloak
Log back in : Updated name is displayed OK

Given scope of changes, I saw no need to re-test other modes.

PR #335, opened 2020-06-09
[OC-871] Smart Notification - Iteration 1

schehade-rtei:


freddidierRTE:
- Remove exception when externalRecipient is not present or empty
- Replace name action with response in UI method and translations
- Rename test.js in templateGateway.js
- Restrict access only to userCard endpoint in nginx

PR #338, opened 2020-06-10
[OC-742] When sending a card to an entity , use the group perimeter t…

vlo-rte:
Release note :
[OC-742](https://opfab.atlassian.net/browse/OC-742) When sending a card to an entity , use the group perimeter to send to the good users


vlo-rte:
There is a bug, I have to update the code.

PR #347, opened 2020-06-12
[535] : bin/run_all.sh start, make it run always from the root opfab …

bendaoudmba:
nothing for release notes


freddidierRTE:
Start works 
But status and stop does not work , it does not find the *.pid files

bendaoudmba:
fixed, go

PR #350, opened 2020-06-15
[OC-944] new endpoint to delete user acknowledgement from a card

vitorz:


freddidierRTE:
1) When we update a card , all ack should be deleted . Check if it's the case if not , implement it . 

Use uid field instead of id field to search the card to update in Mongo to be sure to ack the good card  . When a card is updated , his id is the same but his uid changes , as we want the acknowledgment to be relative to a version of card we need to use uid .(there could be case where the card has been udpated just before the ack , the front has not receveid yet the updated card and the ack is relative to the old one -> it shoud not in this case ack the new card , using uid prevent form that) 

Make a test with karate :
- post a card 
- get the card  
- ack a card with uid 
- check the card is acked
- update the card : POST a new card with the same publisher / processId
-  check there's no ack in the updated card .

2) Solve minor sonar issues : https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&pullRequest=350&resolved=false&types=CODE_SMELL

3) Squash the two commit  , the first commit contains just new space

vitorz:
> 1. When we update a card , all ack should be deleted . Check if it's the case if not , implement it .
> 
> Use uid field instead of id field to search the card to update in Mongo to be sure to ack the good card . When a card is updated , his id is the same but his uid changes , as we want the acknowledgment to be relative to a version of card we need to use uid .(there could be case where the card has been udpated just before the ack , the front has not receveid yet the updated card and the ack is relative to the old one -> it shoud not in this case ack the new card , using uid prevent form that)
> 
> Make a test with karate :
> 
> * post a card
> * get the card
> * ack a card with uid
> * check the card is acked
> * update the card : POST a new card with the same publisher / processId
> * check there's no ack in the updated card .
> 
> 1. Solve minor sonar issues : https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&pullRequest=350&resolved=false&types=CODE_SMELL
> 2. Squash the two commit  , the first commit contains just new space

fixed/added the things in the quoted comment.

freddidierRTE:
Still the id to replace with uid : 

Use uid field instead of id field to search the card to update in Mongo to be sure to ack the good card . When a card is updated , his id is the same but his uid changes , as we want the acknowledgment to be relative to a version of card we need to use uid .(there could be case where the card has been udpated just before the ack , the front has not receveid yet the updated card and the ack is relative to the old one -> it shoud not in this case ack the new card , using uid prevent form that)

vitorz:
> Still the id to replace with uid :
> 
> Use uid field instead of id field to search the card to update in Mongo to be sure to ack the good card . When a card is updated , his id is the same but his uid changes , as we want the acknowledgment to be relative to a version of card we need to use uid .(there could be case where the card has been udpated just before the ack , the front has not receveid yet the updated card and the ack is relative to the old one -> it shoud not in this case ack the new card , using uid prevent form that)

fixed/added the things in the quoted comment, really.

PR #352, opened 2020-06-15
[OC-957] : Field recipient in card shoud be optionnal

bendaoudmba:
for release  notes:

[OC-957] : Field recipient in card shoud be optionnal

[OC-957]: https://opfab.atlassian.net/browse/OC-957


freddidierRTE:
add a karate test sending a card with no recipient and entityRecipients to ["TSO1"]
in  file : src/test/api/karate/cards/cards.feature

PR #360, opened 2020-06-19
sets timeout for SSE to 1 year

rlg-rte:
Signed-off-by: LE-GALL Ronan Ext <43667786+rlg-rte@users.noreply.github.com>


rlg-rte:
doesn't pass the branch name chec.

PR #368, opened 2020-06-22
[OC-951] Process and state fields of a card must be mandatory

vlo-rte:
Release note :
[OC-951](https://opfab.atlassian.net/browse/OC-951) Process and state fields of a card must be mandatory


freddidierRTE:
Add a karate test with no process  and a test with no state

PR #421, opened 2020-07-20
[OC-980] Action & acknowledge button in bottom right

schehade-rtei:


freddidierRTE:
1) Vertical Scroll bar is always present event if not needed : 
it seems to work if you remove  `cardTemplate.style.overflowY = 'scroll';` from detail.component.ts

2) When responding to a card , the response(child)  is in the card but if we go to see an another card and we go back to the card we do not see anymore responses . 

3) Sonar issues :  https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&pullRequest=421&resolved=false

PR #429, opened 2020-07-28
[OC-936] corrects sonar alerts

rlg-rte:
Corrects code to be  sonar requirement compliant.


rlg-rte:
Another failure. The logging table is in a wrong state of code.

PR #430, opened 2020-07-28
[OC-936] corrects sonar alerts and translates logging table headers

rlg-rte:
Aligns code to be Sonar compliant and translates monitoring table headers.

### DIFF
--- a/ui/main/src/assets/i18n/fr.json
+++ b/ui/main/src/assets/i18n/fr.json
@@ -123,7 +123,13 @@
 
   },
   "monitoring": {
-    "filters": {"process": "Processus"}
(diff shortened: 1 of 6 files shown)


ui/main/src/assets/i18n/fr.json
```diff
@@ -123,7 +123,13 @@
 
   },
   "monitoring": {
-    "filters": {"process": "Processus"}
+    "filters": {"process": "Processus"},
+    "time": "Date",
+    "businessPeriod": "Business Période",
```
AlexGuironnetRTE, line 128:
```suggestion
    "businessPeriod": "Période métier",
```

PR #431, opened 2020-07-28
[Karate] explains how to use karate utilities

rlg-rte:
Clarifies how to uses `karate` framework and ready made scripts of the `${OF_HOME}/src/test/utils/karate` folder.


rlg-rte:
Renames the branch and the PR in order to pass the CI.

PR #468, opened 2020-09-11
[OC-1076] Add a filter to see only revelant process

vlo-rte:
Release note : 
[[OC-1076](https://opfab.atlassian.net/browse/OC-1076)] : Add a filter to see only revelant process


freddidierRTE:
Add a field calendar visibility in backend service (uiVisibility.calendar : :set to true if cards of this process are visible in the calendar screen)

Set the flags to true for bundle example  :
- services/core/businessconfig/src/main/docker/volume/businessconfig-storage/first
- services/core/businessconfig/src/main/docker/volume/businessconfig-storage/TEST

PR #504, opened 2020-10-05
[OC-1107] Lttd- only minutes and seconds are displayed in the countdown

taoufikbermaki:


freddidierRTE:
It does not work on my computer . I use the src/test/utils/karate/cards/post6CardsSeverity.feature and set the lttd to 1 hour and a minute (line 167 of the file) . See the two screen shot : the first when when there is more than one our left and the second when there is less than on hour left 

image

image

PR #505, opened 2020-10-05
[OC-1117] Remove template and css arrays in bundle defintion

quinarygio:


services/core/businessconfig/src/main/java/org/lfenergy/operatorfabric/businessconfig/services/ProcessesService.java
```diff
@@ -202,11 +203,17 @@ public Resource fetchResource(String processId, ResourceTypeEnum type, String ve
...
+        if (type.isLocalized() && locale == null)
+            throw new FileNotFoundException("Unable to determine resource for undefined locale");
+
         String finalName;
         if (type == ResourceTypeEnum.I18N) {
             finalName = locale;
         } else {
+            if (!SecurityUtils.checkFilename(name)) {
+                throw new FileNotFoundException("Filename contains forbidden characters");
+            }
```
freddidierRTE, line 216: You shall test locale and version parameters to not contains forbidden characters

freddidierRTE (changes requested):
- After some reflection and test : in our context we in fact only need to the presence of ../ or ..\
and so not even use regexp but method contains from String 
- Please squash your two commits 
- Add two karate test with illegal filenames . Create the test in src/test/api/karate/businessconfig , name it : security.feature
and add it to the script src/test/api/karate/launchAllBusinessconfig.sh

services/core/businessconfig/src/main/java/org/lfenergy/operatorfabric/businessconfig/services/ProcessesService.java
quinarygio: Ok to check the locale. I think that the version parameter is already validated against the available process versions and if it does not match an Exception is thrown.

PR #506, opened 2020-10-05
[OC 1143] Improve new card feature 

freddidierRTE:
Release note : 

Feature 
[OC 1143] Improve new card feature

### DIFF
--- a/src/docs/asciidoc/reference_doc/template_description.adoc
+++ b/src/docs/asciidoc/reference_doc/template_description.adoc
@@ -416,9 +416,9 @@ The output will be:
 <p>student3: 9</p>
 ....
 
-=== textToHtml
+=== keepSpacesAndEndOfLine
 
-Convert a string to HTML by replacing :
+Convert a string to a light HTML by replacing :
 
   -  each new line character with <br/>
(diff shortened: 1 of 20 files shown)


ui/main/src/app/modules/cards/services/handlebars.service.ts
```diff
@@ -340,6 +341,16 @@ export class HandlebarsService {
             return buffer;
         });
     }
+
+    private registerTextToHtml() {
```
rlg-rte, line 345: As this helper handle only end of line and spaces, call it something like `keepSpacesAndEndOfLine` may be less misleading. At the first glance, I thought this helper was able to manage strong, italic and stuff like that.

src/docs/asciidoc/reference_doc/template_description.adoc
```diff
@@ -416,6 +416,13 @@ The output will be:
 <p>student3: 9</p>
 ....
 
+=== textToHtml
```
rlg-rte, line 419: For this name, see my review below.

rlg-rte (approved):

PR #520, opened 2020-10-12
[OC-1126] Add field userRecipients and groupRecipients when posting a card

quinarygio:


services/core/cards-publication/src/main/java/org/lfenergy/operatorfabric/cards/publication/services/RecipientProcessor.java
```diff
@@ -38,8 +41,21 @@
...
+        if ((card.getUserRecipients() != null && !card.getUserRecipients().isEmpty())
+                        || (card.getGroupRecipients() != null && !card.getGroupRecipients().isEmpty())) {
+            ComputedRecipient.BuilderEncapsulator result = ComputedRecipient.encapsulatedBuilder();
+            ComputedRecipient.ComputedRecipientBuilder builder = result.builder();
+
+            builder.users(card.getUserRecipients() == null ? Collections.emptyList() : card.getUserRecipients());
+            builder.groups(card.getGroupRecipients() == null ? Collections.emptyList() : card.getGroupRecipients());
+            computedRecipient = builder.build();
+        } else {
+            log.warn("Using deprecated recipient field");
```
freddidierRTE, line 54: Change the warning text to : 

_"Use of recipient field is deprecated. Please use usersRecipients and groupsRecipients fields (See documentation)"_

freddidierRTE (changes requested):
- Keep the old version of recipient for src/docs/asciidoc/getting_started/index.adoc (as this file is used with getting started repository which does not use the last version of opfab)
- Keep some karate tests for deprecated version : i suggest to duplicate the two following karate test :  post2CardsGroupRouting.feature & postCardFor3Users.feature
  The two old test will use the recipient field and will be named  :
    - post2CardsGroupRoutingDeprecated.feature
    - postCardFor3UsersDeprecated.feature
 -> change processInstanceId in the card to not confuse with card with the new syntax

PR #524, opened 2020-10-13
[OC-1162] For the archives result, the tags and the processes chosen are not considered

schehade-rtei:
[OC-1162] For the archives result, the tags and the processes chosen are not considered

[OC-1162]: https://opfab.atlassian.net/browse/OC-1162


rlg-rte (commented):
Could you add this folder `ui/.vscode` into the `${OF_HOME}/.gitignore` file ?

PR #544, opened 2020-10-20
[OC-1148] Remove email address and date/time formatting parameters fr…

quinarygio:
…om the settings


ui/main/src/app/modules/settings/components/settings/settings.component.ts
```diff
@@ -37,13 +37,13 @@ export class SettingsComponent implements OnInit {
...
+  //email: boolean;
   description: boolean;
   language: boolean;
   timezone: boolean;
-  timeformat: boolean;
-  dateformat: boolean;
-  datetimeformat: boolean;
+  //timeformat: boolean;
+  //dateformat: boolean;
+  //datetimeformat: boolean;
```
freddidierRTE, line 46: Remove comments

config/dev/web-ui-test.json
```diff
@@ -122,9 +122,6 @@
     },
     "infos": {
       "description": true,
-      "dateformat":true,
-      "timeformat":true,
-      "datetimeformat":true,
       "disable": false
```
freddidierRTE, line 125: remove disable field
To do as well in web-ui.json and config/docker/web-ui.json

ui/main/src/app/modules/settings/components/settings/settings.component.ts
freddidierRTE: remove disableInfos variable as it is not used anymore

freddidierRTE (changes requested):
In ui/main/src/app/services/time.service.ts  

```
    private timeFormat;
    private dateFormat;
    private dateTimeFormat;  
```

--> refacto to not use variable anymore but  constants

PR #548, opened 2020-10-22
[OC-1190] Add an option in card to keep child card

vlo-rte:
Release note : 

In Tasks section : 

[[OC-1190](https://opfab.atlassian.net/browse/OC-1190)] Add an option in card to keep child card. You can find more information in the documentation : https://opfab.github.io/documentation/current/reference_doc/#_send_a_question_card


freddidierRTE:
Do not declare the new field in LightCardConsultationData as it will be not used in this object

vlo-rte:
> Do not declare the new field in LightCardConsultationData as it will be not used in this object

Done

PR #563, opened 2020-11-03
[OC-1214] Add option to hide severity sort & read sort in UI

quinarygio:


freddidierRTE:
If feed.card.hideReadSort is true , by default card are sort by read . If feed.card.hideReadSort is false , the cards are not sorted by read

In the actual code , it is always not sorted by read by default

PR #574, opened 2020-11-13
[OC-1233] Add prometheus monitoring endpoint

freddidierRTE:
Release note:

Features:
[OC-1233] Add prometheus monitoring endpoint (see documenation https://opfab.github.io/documentation/current/deployment/#_monitoring) 

[OC-1233]: https://opfab.atlassian.net/browse/OC-1233

### DIFF
--- a/config/monitoring/stopPrometheus.sh
+++ b/config/monitoring/stopPrometheus.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+docker stop prometheus
+docker rm  prometheus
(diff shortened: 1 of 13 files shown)


config/monitoring/stopPrometheus.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+docker rm  prometheus
```
rlg-rte, line 2: When used gives:

`Error response from daemon: You cannot remove a running container <containerId>. Stop the container before attempting removal or force remove`

where `<containerId>` is the Id of the running container.

should have before:

`docker stop prometheus`

rlg-rte (commented):
Allow a clean stop of prometeus docker image with the script would be more user friendly.

rlg-rte (approved):

PR #577, opened 2020-11-17
[trinity] : fix problem==>  Error: Uncaught (in promise): issuer must…

bendaoudmba:
… use https, or config value for property requireHttps must allow http est fixé  sur SCC,


freddidierRTE:
transfered in OC-1235

PR #589, opened 2020-11-20
[OC-1212] Fix bug on page reload when an external page is open in iframe

quinarygio:


ui/main/src/app/services/config.service.ts
```diff
@@ -60,16 +60,15 @@ export class ConfigService {
...
         } else {
-            resp = this.getMenuEntryURL(id, menuEntryId);
+            return this.getMenuEntryURL(this.customMenus, id, menuEntryId);
         }
-        return resp;
+
     }
 
-    private getMenuEntryURL(id: string, menuEntryId: string): Observable<string> {
+    private getMenuEntryURL(menus: Menu[], id: string, menuEntryId: string): Observable<string> {
```
freddidierRTE, line 71: Why adding this menus that is not used in the method ?

freddidierRTE (changes requested):

PR #614, opened 2020-12-03
[OC-1242] View repeating events in timeline

quinarygio:


freddidierRTE:
No circle/bubble shown for timeSpan.start and timeSpan.end when the recurrence is used

PR #616, opened 2020-12-03
OC-1219 : Use notification filters to filter the feed of the user

vlo-rte:
This PR concernes [OC-1219](https://opfab.atlassian.net/browse/OC-1219) : Use notification filters to filter the feed of the user

**!! Warning !!** : the unit test receiveCards() in cards-consultation is desactived because it fails everytime with this modification.

### DIFF
--- a/services/core/users/src/main/java/org/lfenergy/operatorfabric/users/controllers/UsersController.java
+++ b/services/core/users/src/main/java/org/lfenergy/operatorfabric/users/controllers/UsersController.java
@@ -125,6 +125,7 @@ public class UsersController implements UsersApi {
     public UserSettings patchUserSettings(HttpServletRequest request, HttpServletResponse response, String login, UserSettings userSettings) throws Exception {
         UserSettingsData settings = userSettingsRepository.findById(login)
                 .orElse(UserSettingsData.builder().login(login).build());
+        publisher.publishEvent(new UpdatedUserEvent(this, busServiceMatcher.getServiceId(), login));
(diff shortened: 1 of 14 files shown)


services/core/users/src/main/java/org/lfenergy/operatorfabric/users/controllers/UsersController.java
```diff
@@ -125,6 +125,7 @@ public UserSettings fetchUserSetting(HttpServletRequest request, HttpServletResp
     public UserSettings patchUserSettings(HttpServletRequest request, HttpServletResponse response, String login, UserSettings userSettings) throws Exception {
         UserSettingsData settings = userSettingsRepository.findById(login)
                 .orElse(UserSettingsData.builder().login(login).build());
+        publisher.publishEvent(new UpdatedUserEvent(this, busServiceMatcher.getServiceId(), login));
```
freddidierRTE, line 128: Do it only if patching processesStatesNotNotified

freddidierRTE (commented):
Correct sonar warning
Add karate tests using a call to cardSubscription endpoint with parameter notification set to false

services/core/users/src/main/java/org/lfenergy/operatorfabric/users/controllers/UsersController.java
vlo-rte: Sonar warnings solved. 
Complexity of method checkIfUserMustReceiveTheCard(...) reduced.
The karate tests will be done with another ticket.

PR #617, opened 2020-12-04
[OC-1136] Change checkbox labels style in settings page and feed type…

quinarygio:
… filter


freddidierRTE:
image

Align text/color with checkbox (using for example : margin-top css directive) 

Same for settings screen

PR #624, opened 2020-12-11
[OC-1272]  Show ack button only if action is disabled for the user

quinarygio:


freddidierRTE (changes requested):
Add control for  the acknowledgmentAllowed field in karate test /home/didierfre/projets/opfab/git/operatorfabric-core/src/test/api/karate/businessconfig/getABusinessconfig.feature

PR #630, opened 2020-12-16
[OC-1279] Do not show the form for user card When no process is avail…

quinarygio:
…able for user card.

If usercard field is not set in config.json, state should not be available for selection in usercard.
Changed severity choice design in usercard using radio buttons instead of select.

### DIFF
--- a/ui/main/src/app/modules/usercard/usercard.component.html
+++ b/ui/main/src/app/modules/usercard/usercard.component.html
@@ -23,10 +23,26 @@
                     [values]="stateOptions"></of-single-filter>
 
             </div>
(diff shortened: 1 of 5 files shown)


freddidierRTE (changes requested):
Not initally in the issue but could you change french translation from : 
        "ALARM": "Alarm",
        "ACTION": "Action",
        "COMPLIANT": "Compliant",
        "INFORMATION": "Information"

To 
        "ALARM": "Alarme",
        "ACTION": "Action",
        "COMPLIANT": "Conforme",
        "INFORMATION": "Information"

PR #680, opened 2021-01-13
[OC-1320] Add state name in card header for question card

vlo-rte:


ui/main/src/app/modules/cards/components/detail/detail.component.scss
```diff
@@ -155,6 +155,10 @@
   mask: url(assets/images/icons/help.svg);
 }
 
+.question-card-state-name{
+  color: var(--opfab-light-card-lttd-timeleft);
+}
+
```
freddidierRTE, line 161: class name to prefix with opfab-

freddidierRTE (changes requested):

ui/main/src/app/modules/cards/components/detail/detail.component.scss
vlo-rte: Done

PR #683, opened 2021-01-14
[OC-1325] Trim list of entity to answer if there are too many entitie…

quinarygio:
…s to show

### DIFF
--- a/ui/main/src/app/modules/cards/components/detail/detail.component.ts
+++ b/ui/main/src/app/modules/cards/components/detail/detail.component.ts
@@ -18,7 +18,8 @@ import {
     OnDestroy,
     OnInit,
     TemplateRef,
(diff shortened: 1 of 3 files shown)


ui/main/src/app/modules/cards/components/detail/detail.component.ts
```diff
@@ -94,10 +95,13 @@ const enum EntityMsgColor {
     ORANGE = '#ff6600'
 }
 
+const maxVisibleEntitiesToRespond = 2;
```
freddidierRTE, line 98: Set value to 3

freddidierRTE (commented):
image

- Reduce margin/padding on top 
- Remove vertical bar between answers 
- Align left the list in the popup

PR #688, opened 2021-01-15
[OC-1330] About screen in a modal

quinarygio:


freddidierRTE (commented):
The logo is not centered the same as the text 
Change the color of the text and title to `var(--opfab-text-color-stronger)` 
Position of the modal should be in the center of the screen

PR #690, opened 2021-01-18
[OC-1353] Do not permit to have preview of card at the same time as view card in the feed

freddidierRTE:


AlexGuironnetRTE (approved):
Comment is clear and behaves as expected (test with ng serve, on chrome).

PR #702, opened 2021-01-19
[OC-1322] Add link to hide filter in monitoring screen

AlexGuironnetRTE:
## Implementation choices

* No need to make default value configurable like hide/show timeline, by default filters are shown.
* Show/hide value is stored in local storage.
* The "summary" shown when the filters are hidden displays the values from the last sent query (which matches the results that are being displayed)
Meaning that if I send a query filtering on process "Test1", then change to "Test2" (without submitting the query), when I hide the filters the "summary" will display Test1. It might sometimes be confusing for the user but this way at least the filter summary is consistent with the results.
* I didn't do the white/bold style on the summary values (as described in Invision) because it wasn't the case for the existing text displayed when the timeline is hidden, I think we should either do it for both or not do it at all.

## Tests
* Tested with night and day mode
* Tested with long lists of processes, with no selected values
* Tested translation

## Notes
Created OC-1368 for issue with i18n assets

### DIFF
--- a/ui/main/src/app/modules/monitoring/components/monitoring-filters/monitoring-filters.component.ts
+++ b/ui/main/src/app/modules/monitoring/components/monitoring-filters/monitoring-filters.component.ts
@@ -7,7 +7,7 @@
  * This file is part of the OperatorFabric project.
  */
 
-import {Component, Input,OnDestroy,OnInit} from '@angular/core';
+import {Component, Input, OnDestroy, OnInit} from '@angular/core';
 import {AbstractControl, FormControl, FormGroup} from '@angular/forms';
 import {Store} from '@ngrx/store';
 import {AppState} from '@ofStore/index';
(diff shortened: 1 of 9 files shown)


ui/main/src/app/modules/monitoring/components/monitoring-filters/monitoring-filters.component.ts
```diff
@@ -192,6 +218,39 @@ export class MonitoringFiltersComponent implements OnInit, OnDestroy {
         this.sendQuery();
     }
 
+    showOrHideFilters() {
+        console.log("AGU ",this.selectedProcesses.value);
```
freddidierRTE, line 222: Remove console.log

freddidierRTE (changes requested):
Solve sonar issue : 
https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&open=AXcfNRbnkk-kPJ77iZKE&pullRequest=702&resolved=false&types=BUG

Bug : 
 1) Go to monitoring screen
 2) Hide the filters
 3) Go in the feed or an other screen 
3)  Go back to monitoring
--> Filter should be at initial state and so have default activation date which is not the case (if we unhide the filters , we see the good activation period)

PR #710, opened 2021-01-21
[OC-1372] Use common active date filters initialization for monitoring

AlexGuironnetRTE:


freddidierRTE:
remove MONITOR_DATE_FILTER in ui/main/src/app/services/filter.service.ts
remove use of MONITOR_DATE_FILTER in ui/main/src/app/modules/monitoring/components/monitoring-filters/monitoring-filters.component.ts

AlexGuironnetRTE:
> remove MONITOR_DATE_FILTER in ui/main/src/app/services/filter.service.ts
> remove use of MONITOR_DATE_FILTER in ui/main/src/app/modules/monitoring/components/monitoring-filters/monitoring-filters.component.ts

I thought this was still needed because of the separate behaviour in monitoring-filter component but I can't figure out was that was for. I removed it and did a few tests and everything looks ok.
- Filter initialization when arriving on monitoring directly, results OK
- Filter initialization when coming back to feed/archives OK

PR #713, opened 2021-01-22
[OC-1376] When clicking on an unread card, it should not move down in the feed

freddidierRTE:
Solution is to keep card as unread in the UI until we exit the detail of the card.

### DIFF
--- a/ui/main/src/app/modules/cards/components/detail/detail.component.ts
+++ b/ui/main/src/app/modules/cards/components/detail/detail.component.ts
@@ -139,6 +139,8 @@ export class DetailComponent implements OnChanges, OnInit, OnDestroy, AfterViewC
     public showActionButton = false;
     public showEditAndDeleteButton = false ;
 
+    private cardSetToReadButNotYetOnUI;
+
     modalRef: NgbModalRef;
 
     public displayDeleteResult = false;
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/cards/components/detail/detail.component.ts
```diff
@@ -247,9 +248,6 @@ export class DetailComponent implements OnChanges, OnInit, OnDestroy, AfterViewC
                         });
                 });
         }
-        this.loadTranslationForTitleAndStateName();
```
AlexGuironnetRTE, line 250: I understand why the call to markAsReadIfNecessary is removed, but why the ones to checkIfHasAlreadyResponded, markAsReadIfNecessary and setButtonsVisibility() as well ?

AlexGuironnetRTE (changes requested):
Test OK : click on unread card, card doesn't change position in feed.
Once we click on a second card or close the details, card takes appropriate position in feed.

Note : There is one collateral small change in behaviour with this version. 
In the current develop version, the card loses its "unread" style (bold etc.) as soon as it is clicked. With this code, the card keeps its "unread" style until the detail is closed.
I don't think it's a problem but it might be worth mentioning it somewhere so users don't think it's a bug ?

I just have one question about the modified code (see comment).

ui/main/src/app/modules/cards/components/detail/detail.component.ts
freddidierRTE: I ve seen that this methods are called in ngOnChange and each time ngOnInit is called ngOnChange is called . So it was called double .

PR #716, opened 2021-01-22
[OC-1373] Show header card for question card shoud be optionnal

vlo-rte:


ui/main/src/app/modules/cards/components/detail/detail.component.ts
```diff
@@ -306,6 +306,10 @@ export class DetailComponent implements OnChanges, OnInit, OnDestroy, AfterViewC
         return this.card.hasBeenAcknowledged ? AckI18nKeys.BUTTON_TEXT_UNACK : AckI18nKeys.BUTTON_TEXT_ACK;
     }
 
+    get showDetailCardHeader(): boolean {
+        return (this.cardState.showDetailCardHeader === null) || (this.cardState.showDetailCardHeader === true);
```
freddidierRTE, line 310: I think you can write : 
`
return !!this.cardState.showDetailCardHeader`

which will return true if it exists and has value true

vlo-rte: Yes but I thought that we wanted it to return true also if the value is absent, no ?

freddidierRTE: Yes you are right,   i merge

PR #720, opened 2021-01-25
[OC-1375] Ack & time style in feed

quinarygio:


ui/main/src/app/modules/share/light-card/light-card.component.html
```diff
@@ -16,32 +16,41 @@
     <div class="d-flex  p-0 border-0">
         <div class="badge-border  opfab-bg-sev-{{lightCard.severity.toString()|lowercase}}">
         </div>
-        <div class="p-1" [class.text-truncate]="!open">
+        <div class="p-1" [class.text-truncate]="!open" style="width: 100%">
```
freddidierRTE, line 19: remove [class.text-truncate]="!open"  :  when reducing the size of the window , the ack icon disappear because of this style .
And we do not really need this truncating of the text

freddidierRTE (changes requested):

PR #737, opened 2021-01-27
[OC-1403] Fix delete card popup style

quinarygio:


freddidierRTE:
image

It is ok in english but in french it does not fit .

PR #777, opened 2021-02-05
[OC-1396] Communicate to the template the screen size

quinarygio:
Maybe it should be added also in release notes ?

In Tasks section:
OC-1396 Communicate to the template the screen size calling setScreenSize() method


freddidierRTE:
When in french , it does not work with the process example .
-- > Need to add svg id="chart" in french template

PR #782, opened 2021-02-08
Oc 1283

sadema:
This pull request contains no breaking changes.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -82,7 +82,8 @@ cd operatorfabric-core
 . Launch our demo docker-compose file
 +
 ----
-./config/docker/docker-compose.sh
+cd ./config/docker
+./docker-compose.sh
 ----
 
 . After a little while, log into the application UI at *localhost:2002/ui/* using admin/test as credentials.
--- a/bin/run_all.sh
+++ b/bin/run_all.sh
@@ -118,7 +118,7 @@ startProject(){
         cd $OF_HOME
         java -Xss512k -XX:MaxRAM=512m $debugOptions \
              -jar $projectBuildPath/libs/$1-$version.jar \
-             $applicationOptions 2>&1 > $projectBuildPath/logs/$(date -I).log &
+             $applicationOptions 2>&1 > $projectBuildPath/logs/$(date \+"%y-%m-%d").log &
 #        set +x
         echo $! > $projectBuildPath/PIDFILE
 
@@ -166,7 +166,7 @@ stopCommand(){
   for ((i=0; i<${#dependentProjects[*]}; ));
   do
     stopProject ${dependentProjects[i]} ${dependentProjects[i+1]}
-    i=$((i+"$PRJ_STRC_FIELDS"))
+    i=$((i+$PRJ_STRC_FIELDS))
   done
 }
 
@@ -192,7 +192,7 @@ hardstopCommand(){
   for ((i=0; i<${#dependentProjects[*]}; ));
   do
     hardstopProject ${dependentProjects[i]} ${dependentProjects[i+1]}
-    i=$((i+"$PRJ_STRC_FIELDS"))
+    i=$((i+$PRJ_STRC_FIELDS))
   done
 }
 
--- a/config/docker/docker-compose.sh
+++ b/config/docker/docker-compose.sh
@@ -1,7 +1,21 @@
 #!/bin/bash
 
+echo "Usage: "
+echo "   1: ./docker-compose.sh"
+echo "   2: ./docker-compose.sh <pathToExternalEnvironmentFile>"
+
 echo USER_ID="$(id -u)" > .env
 echo USER_GID="$(id -g)" >> .env
 # create directory for bundle storage if not existing
 mkdir -p businessconfig-storage
+if [ "$#" -eq 0 ]; then
+  echo CONFIG_PATH=./ >> .env
+  echo SPRING_PROFILES_ACTIVE=docker >> .env
+else
+  EXTERNAL_CONFIGURATION_FILE=$1
+  CONFIG_PATH=$(dirname "$1")
+  cat ${EXTERNAL_CONFIGURATION_FILE} >> .env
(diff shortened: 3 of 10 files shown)


freddidierRTE:
1) When I launch the script config/docker/docker-compose.sh, it does not work 

It is because the variable :  SPRING_PROFILES_ACTIVE  SPRING_CONFIG_LOCATION are not set by the script but define in docker-compose.yml 

If i remove the variables form docker compose, it starts well with the default config.

I guess ,  you have to define default values in the docker-compose.sh 

2) Instead of modifying the readme, please put the documentation in the configuration documentation (src/docs/asciidoc/deployment/configuration/configuration.adoc)  ->  create a new chapter in  chapter "Business service configuration"

sadema:
The documentation is updated and the readme has been restored. A fix is made in the shell script.
Thanks for the feedback!

AlexGuironnetRTE:
Hello !

Thanks for the PR and sorry for the late reply. I tested running both the docker mode and the dev + run all mode and everything looks good! I do have a few questions though, sorry if some of them are obvious as I haven't followed the subject from the start and I'm a bit rusty on docker.

- From what I read from @JeroenGommans on the JIRA ticket the initial goal of this issue was to avoid duplicating configuration across files for the different "modes" that we use to run OpFab (namely the .yml files under config/dev and config/docker I assume), but I see that so far we're keeping both sets unchanged, is that normal?

- The SPRING_CONFIG_LOCATION is mentioned under "environment" in the config/docker docker-compose file, which I understand as "take the value from this variable from the current shell", but it's not set in the docker-compose.sh script, is it meant to be set manually?

- This SPRING_CONFIG_LOCATION is also set to './config/,./external-config/' in the Dockerfiles for each service, so it's this value that will be embedded in the containers when they are built, but it will be overridden by the value from the shell (if it exists) if we run them using the config/docker docker-compose file, correct? If so, shoudn't we make the same change to the config/dev docker-compose?

I'm still looking into it because reviewing this PR made me realize that it's not clear in my head (or in the doc either) where/when (build time or run time) the various configuration files we have are taken into account and what's the overriding hierarchy (something like [this](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config)).
For example, if your change is only meant for dev mode put it there, if it's a default value that needs to be shipped with the jars (or with the docker images) and can be overridden later put it there, or on the contrary if it's a property that we can't provide a default for and needs to be set by the person doing the deployment, etc.

Thanks in advance for your reply!

sadema:
Hi Alex, thanks for your reply!

> On 17 Feb 2021, at 15:25, AlexGuironnetRTE <notifications@github.com <mailto:notifications@github.com>> wrote:
> 
> 
> Hello !
> 
> Thanks for the PR and sorry for the late reply. I tested running both the docker mode and the dev + run all mode and everything looks good! I do have a few questions though, sorry if some of them are obvious as I haven't followed the subject from the start and I'm a bit rusty on docker.
> 
> From what I read from @JeroenGommans <https://github.com/JeroenGommans> on the JIRA ticket the initial goal of this issue was to avoid duplicating configuration across files for the different "modes" that we use to run OpFab (namely the .yml files under config/dev and config/docker I assume), but I see that so far we're keeping both sets unchanged, is that normal?
> 
> The whole goal here is that we want to avoid making changes in the configuration files when we are developing. In our development environment we have to enable Kafka and therefore need to provide a lot of additional settings. We do not want to touch the default configuration files because we can accidentally check them in. We do not want to enter the configuration every time we change branches either. With the changes in this pull request we can start all the services with the docker-compose.sh command and provide an external location with additional configuration settings.
> The SPRING_CONFIG_LOCATION is mentioned under "environment" in the config/docker docker-compose file, which I understand as "take the value from this variable from the current shell", but it's not set in the docker-compose.sh script, is it meant to be set manually? Sharp notice ;-) I can take this environment variable out because we do not use it at the moment in the compose file.
> 
> This SPRING_CONFIG_LOCATION is also set to './config/,./external-config/' in the Dockerfiles for each service, so it's this value that will be embedded in the containers when they are built, but it will be overridden by the value from the shell (if it exists) if we run them using the config/docker docker-compose file, correct? If so, shoudn't we make the same change to the config/dev docker-compose? I don’t think so because in the dev environment the Java processes are not started. Just keycloack, mongodb, etc. 
> 
> I'm still looking into it because reviewing this PR made me realize that it's not clear in my head (or in the doc either) where/when (build time or run time) the various configuration files we have are taken into account and what's the overriding hierarchy (something like this <https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config>). 
> For example, if your change is only meant for dev mode put it there, if it's a default value that needs to be shipped with the jars (or with the docker images) and can be overridden later put it there, or on the contrary if it's a property that we can't provide a default for and needs to be set by the person doing the deployment, etc.
> 
> I hope my reply on the first bullet clarified our use-case a bit ;-) Summarizing, the changes are primary ment for development. Bycatch is that we can mount the external location providing  additional settings when we deploy to Kubernetes.
> Thanks in advance for your reply!
> 
> No problem! Get in touch when I have to clarify things!

> Kind Regards
   Sjoerd Adema

AlexGuironnetRTE:
Hi Sjoerd, 

Thanks for the explanation, now I understand! It's quite a different use case from what I understood so now it makes sense ;-) !
In fact we do have the same need for our test/demo instances (@HanaeSafiRTE), we were managing using git stash but it's quite brittle indeed, so it's a very welcome change!
Yes you're right about the dev docker-compose, so if that's ok for you I'll just remove the SPRING_CONFIG_LOCATION mention from the docker-compose, squash and merge.

And I've created a different issue (OC-1493 - Review configuration files and document them better) for what I had in mind!

Thanks again!

Alex

sadema:
Hey Alex,

Great! Totally agree with your comments!!

Kind regards
Sjoerd

PR #784, opened 2021-02-10
[OC-1061] Upgrade Angular from v8.2 to v10

AlexGuironnetRTE:
## To run this branch locally

- Pull the branch
- Source the bin/load_environment_light.sh script to make sure you're using the correct versions of node
- Run npm install
If you run into any errors with ng build, test or serve, delete the node_modules folder entirely and run npm install again.
If the build stalls or is taking to long because of the large number of dependencies to re-package, you might need to use 
`node --max_old_space_size=4096 node_modules/@angular/cli/bin/ng build`

## Comments / issues to watch out for ##

Below is a small list of the things I saw in the update guides that I think we should watch out for in tests and take into account for future developments : 

- Issue with url paths referenced from css files (mostly icons)
Url paths to resources in css files are now resolved relative to the css file location rather than the assets file specified in angular.json
See these for details:
https://github.com/angular/angular-cli/issues/18013
https://stackoverflow.com/questions/62619062/unable-to-use-assets-in-url-in-scss-after-upgrading-to-angular-10/62619147#62619147
For now I've fixed the issue by changing the paths to take that into account (solution 1) but it's not very handy so l logged OC-1467 to find a better solution.

- Dynamic CSS classes
> Bound CSS styles and classes previously were applied with a "last change wins" strategy, but now follow a defined precedence. Learn more about Styling Precedence.
> https://angular.io/guide/attribute-binding#styling-precedence

- Everyone should have a decorator
> All classes that use Angular DI must have an Angular decorator like @Directive() or @Injectable (previously, undecorated classes were allowed in AOT mode only or if injection flags were used).

- Change in precedence rules for attributes
> Static attributes set directly in the HTML of a template will override any conflicting host attributes set by directives or components (previously, static host attributes set by directives / components would override static template attributes if conflicting).

## Release notes
Under tasks ? 
[OC-1061] Upgrade Angular from v8.2 to v10.2

## Next steps
- Fix CommonJs warning -> OC-1468
- Fix CSS for calendar -> OC-1470
- Upgrade to Angular v11... -> OC-1071

[OC-1061]: https://opfab.atlassian.net/browse/OC-1061


AlexGuironnetRTE:
== Test plan

=== Tests on serve by comparing to develop on 2002/ui

==== admin user



==== operator1 user

=== Tests on the built version using config/dev docker-compose
_To make sure there are no errors in consoles or issues that didn't happen with serve)._

PR #785, opened 2021-02-10
[OC-1347] Fix bug with keepChilCards default value

quinarygio:
Release notes :

In Tasks section :
[OC-1347] Fix bug with keepChilCards default value

[OC-1347]: https://opfab.atlassian.net/browse/OC-1347

### DIFF
--- a/services/core/cards-publication/src/main/java/org/lfenergy/operatorfabric/cards/publication/model/CardPublicationData.java
+++ b/services/core/cards-publication/src/main/java/org/lfenergy/operatorfabric/cards/publication/model/CardPublicationData.java
@@ -161,4 +161,13 @@ public class CardPublicationData implements Card {
             result.timeSpansSet(new HashSet<>(this.getTimeSpans()));
         return result.build();
     }
+
+
+    public Boolean getKeepChildCards() {
+        return keepChildCards;
+    }
+
+    public void setKeepChildCards(Boolean keepChildCards) {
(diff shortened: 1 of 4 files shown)


AlexGuironnetRTE (commented):
Everything works perfect, the only thing that is bugging me is that the fact that null is replaced with false is "hidden away" in the deleteChildCards method... 
And it's no big deal because it's unlikely to change, but it means that the default value is defined twice : once on CardPublicationData and once in this method...

I was hoping we could manage that at the model level with Lombok or the serializer/deserializer but it doesn't look like it..
https://stackoverflow.com/questions/43851398/lombok-set-default-value-when-an-attribute-is-explicity-set-to-null-by-builder

quinarygio:
> Everything works perfect, the only thing that is bugging me is that the fact that null is replaced with false is "hidden away" in the deleteChildCards method...
> And it's no big deal because it's unlikely to change, but it means that the default value is defined twice : once on CardPublicationData and once in this method...
> 
> I was hoping we could manage that at the model level with Lombok or the serializer/deserializer but it doesn't look like it..
> https://stackoverflow.com/questions/43851398/lombok-set-default-value-when-an-attribute-is-explicity-set-to-null-by-builder

Found a workaround by defining the setter method for keepChildCards in CardPublicationData

AlexGuironnetRTE (approved):
Great! Thanks for the change!

PR #793, opened 2021-02-11
[OC-1439] New archives screen filters

vlo-rte:
Release notes : 

In Tasks section : 

[OC-1439](https://opfab.atlassian.net/browse/OC-1439) : New archives screen filters

### DIFF
--- a/ui/main/src/app/modules/archives/archives.component.html
+++ b/ui/main/src/app/modules/archives/archives.component.html
@@ -11,21 +11,62 @@
     <form [formGroup]="archiveForm" #currentForm>
       <div class="opfab-archives">
         <div style="display: flex;">
-          <div style="margin-right:40px">
-            <div style="min-width:300px;max-width:400px;margin-top:28px;margin-bottom: 54px;">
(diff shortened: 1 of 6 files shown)


ui/main/src/app/modules/archives/archives.component.html
```diff
@@ -85,6 +127,9 @@
         <td translate [translateParams]="card.summary.parameters">
           {{ card.process + '.' + card.processVersion + '.' + card.summary.key}}
         </td>
+        <td translate>
+          {{ this.findServiceForProcess(card.process) }}
```
quinarygio, line 131: Check if displayServiceFilter() as for column header

quinarygio (commented):
When selecting only the service and not the processes the query does not filters on selected "service"

ui/main/src/app/modules/archives/archives.component.html
vlo-rte: Done.

vlo-rte:
> When selecting only the service and not the processes the query does not filters on selected "service"

I've made an update to filter also on services (when no process selected). Thank you for your remark.

quinarygio (approved):

PR #799, opened 2021-02-16
[OC-1431] Display names rather than IDs for entities/groups in the users table

AlexGuironnetRTE:
I ended up refactoring the components a bit for this ticket, but with perimeters round the corner (OC-1089) I wanted to make things as easy to extend as possible.

Release notes:
Under features, OC-1431 - Display names rather than IDs for entities/groups in the users table (Administration)


quinarygio (commented):
After deleting a Group or an Entity they are still listed in User modal dropdowns, I think they should be removed also  from the cached values.

AlexGuironnetRTE:
> After deleting a Group or an Entity they are still listed in User modal dropdowns, I think they should be removed also from the cached values.

That's right thanks! Would you rather I pushed the fix on a separate commit or did an amend and force-push?

quinarygio:
> > After deleting a Group or an Entity they are still listed in User modal dropdowns, I think they should be removed also from the cached values.
> 
> That's right thanks! Would you rather I pushed the fix on a separate commit or did an amend and force-push?

Maybe better to squash and force push

AlexGuironnetRTE:
Done!

quinarygio (approved):

PR #800, opened 2021-02-17
[OC-1489] Fix bug on delayed display of countdown counter

quinarygio:
The countdown counter disappears for 1 second when changing timeline period choosing a broader period that triggers the reload of cards list.

Release notes:

In Bugs section:
[OC-1489] Fix bug on delayed display of countdown counter 

[OC-1489]: https://opfab.atlassian.net/browse/OC-1489

### DIFF
--- a/ui/main/src/app/modules/share/countdown/countdown.component.ts
+++ b/ui/main/src/app/modules/share/countdown/countdown.component.ts
@@ -103,13 +103,18 @@ export class CountDownComponent implements OnInit, DoCheck, OnChanges, OnDestroy
         } else if (this.getSecondsBeforeLttd() <= 0) {
             this.stopCountDown();
         } else {
+            this.startCountDownConfigWhenNecessary();
             this.interval = setInterval(() => {
-                if (this.isTimeToStartCountDown()) {
-                    this.startCountDownConfig();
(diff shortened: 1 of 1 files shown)


AlexGuironnetRTE (commented):
Hello,

I tested the fix and it works perfectly, but I didn't understand why the startCountDownConfigWhenNecessary method needs to be called twice (at the top of the block then after a `MILLISECONDS_SECOND` delay? Why was the delay needed in the first place? Maybe @vlo-rte knows?

quinarygio:
> Hello,
> 
> I tested the fix and it works perfectly, but I didn't understand why the startCountDownConfigWhenNecessary method needs to be called twice (at the top of the block then after a `MILLISECONDS_SECOND` delay? Why was the delay needed in the first place? Maybe @vlo-rte knows?

The counter has to be displayed only "secondsBeforeLttdForClockDisplay" seconds before lttd. So the delay is needed to check periodically if it's time to display the counter. The problem with setInterval() method is that the first execution is after the specified interval (1 second), so the counter was displayed with 1 second delay.

AlexGuironnetRTE:
Oooh ok I had misunderstood the interval thing completely, thanks!
________________________________

PR #802, opened 2021-02-18
[OC-964] Deal with user in more than one entity

quinarygio:
Release notes:

In Features section:
[OC-964] Deal with user in more than one entity

[OC-964]: https://opfab.atlassian.net/browse/OC-964


AlexGuironnetRTE (approved):
Good thinking about the countdown!

PR #808, opened 2021-02-25
[OC-1446] In client lib avoid serialization of null fields

quinarygio:


src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/cards/CardData.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+@Getter
+@NoArgsConstructor
+@AllArgsConstructor
+public class CardData {
+    Card card;
+    Card[] childCards;
+}
```
AlexGuironnetRTE, line 23: I'm not sure you should create a CardData class like that as there would be nothing enforcing the fact that it should stay aligned with the one under [cards-publication](https://github.com/opfab/operatorfabric-core/tree/develop/services/core/cards-publication/src/main/java/org/lfenergy/operatorfabric/cards/publication/model), which is precisely what the whole swagger-generated interfaces (for all the pains that it causes...) was meant to do.
I would have expected it to follow the usual pattern of swagger def -> generates concrete class under client and matching interface under services... It's my bad because I merged Samy's PR that introduced it... @freddidierRTE what do you think?

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/config/JacksonConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import org.lfenergy.operatorfabric.springtools.json.InstantModule;
+
+/**
+ * Json configuration
+ *
+ *
+ */
+@Configuration
+@Slf4j
+public class JacksonConfig {
```
AlexGuironnetRTE, line 29: Again, that's not really in the scope of this PR but the fact that you created this test app shows that there might be an issue with the existing lib: id we need this so the serialization to create the body of the card is correct from the API point of view, shouldn't it be somehow bundled with the client library (or failing that, mentioned in big red letters in the docs)?
Because my understanding of the client lib is that we could tell people who want to code stuff that sends card to OpFab "here, take that lib, and if you use the card builder in it we guarantee that the generated cards will be in a format that's acceptable by the API"...
Here if they don't add this config to their app they'll get their cards rejected at least because they'll have their dates as timestamps rather than milliseconds from epoch, right?

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/common/CreationReport.java
```diff
@@ -0,0 +1,26 @@
...
+
+
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+
+@Getter
+@AllArgsConstructor
+@NoArgsConstructor
+@ToString
+public class CreationReport {
```
AlexGuironnetRTE, line 23: There's already a CardCreationReport representing the same object that's generated in the client library (albeit not with the same Lombok annotations), shouldn't we use it instead?

AlexGuironnetRTE (changes requested):
Builds, generated classes and tests ok, logs show that even though the null are presents in the Java objects on both sides they're correcly removed from the request body.
Speaking of which, shouldn't we create a ticket so that null fields aren't serialized on the API side either? Meaning if I do a GET on a specific card I don't want to get nulls.

I also think it would be useful to have a pointer to this app in the dev documentation as it can be a very useful example for people looking to send cards to OpFab in their application (maybe in the development environment documentation before the Kafka implementation ?).

AlexGuironnetRTE:
Also I've created [OC-1513 : Have test application depend on published jars rather than project for the client lib](https://opfab.atlassian.net/browse/OC-1513) so we don't forget to make the switch once the client lib jars are on maven Central

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/common/CreationReport.java
quinarygio: Done

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/config/JacksonConfig.java
quinarygio: Not bundled with client library to avoid adding dependency to spring framework

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/cards/CardData.java
AlexGuironnetRTE: Created https://opfab.atlassian.net/browse/OC-1516 to address this.

src/test/clientApp/src/main/java/org/lfenergy/operatorfabric/client/config/JacksonConfig.java
AlexGuironnetRTE: Created https://opfab.atlassian.net/browse/OC-1517 to address it.

PR #810, opened 2021-02-26
OC-1503 Add an opfab style for border box with label in template

freddidierRTE:
Release note :

Features : 

OC-1503 Add an opfab style for border box with label in template


AlexGuironnetRTE (approved):
The two modified templates are ok, and I saw no side effects on other inputs.

PR #822, opened 2021-03-04
[OC-1521] Upgrade to MongoDB 4.4.4

freddidierRTE:
Remove old mongo docker container before starting (docker rm container_id ) 

Release note :

Task : 
OC-1521 Upgrade to MongoDB 4.4.4

### DIFF
--- a/config/dev/docker-compose.yml
+++ b/config/dev/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       - "15672:15672"
       - "15674:15674"
   mongodb:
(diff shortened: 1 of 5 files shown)


AlexGuironnetRTE:
Looking for references to 'xenial' in the project I found one in .travis.yml but it's related to the Ubuntu Xenial distribution (see https://docs.travis-ci.com/user/reference/xenial/), which relies on Mongo 4.0 but as we're using Mongo in docker it's not a problem.

PR #826, opened 2021-03-05
[OC-1532] BUG - usercard - if process has no state available for send…

freddidierRTE:
Release note : 

BUG

[OC-1532] Usercard - if process has no state available for sending card , do not show it

[OC-1532]: https://opfab.atlassian.net/browse/OC-1532

### DIFF
--- a/ui/main/src/app/modules/usercard/usercard.component.ts
+++ b/ui/main/src/app/modules/usercard/usercard.component.ts
@@ -203,8 +203,10 @@ export class UserCardComponent implements OnDestroy, OnInit {
                         const _i18nPrefix = process.id + '.' + process.version + '.';
                         const label = process.name ? (_i18nPrefix + process.name) : process.id;
                         const processToShow = { value: process.id, label: label };
-                        this.processOptions.push(processToShow);
+                        
                         this.loadStatesForProcess(process);
+                        // Add process option only if there is at least one state
+                        if (this.statesPerProcesses.get(process.id).length > 0) this.processOptions.push(processToShow);
                     }
                 });
     }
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/usercard/usercard.component.ts
```diff
@@ -250,25 +251,32 @@ export class UserCardComponent implements OnDestroy, OnInit {
...
-                        }
-                    } else console.log('WARNING : state', perimeter.state , 'is present in perimeter for process'
-                                    , process.id , 'but not in process definition');
-
+                    const state = this.getStateFromProcessDefinition(process,perimeter.state)
+                    if (!!state) statesList.push(state);
                 }
             });
-        this.statesPerProcesses.set(process.id, statesList);
+        if (statesList.length>0) this.statesPerProcesses.set(process.id, statesList);
```
AlexGuironnetRTE, line 264: With this condition, if a process had states available for usercards, and then after an update it doesn't anymore, won't the statePerProcesses continue to contain the previous list (of states that are no longer avaiblable) as the temporary variable "stateList" will have a length of 0?

ui/main/src/app/modules/usercard/usercard.component.ts
```diff
@@ -203,8 +203,9 @@ export class UserCardComponent implements OnDestroy, OnInit {
                         const _i18nPrefix = process.id + '.' + process.version + '.';
                         const label = process.name ? (_i18nPrefix + process.name) : process.id;
                         const processToShow = { value: process.id, label: label };
-                        this.processOptions.push(processToShow);
-                        this.loadStatesForProcess(process);
+                        
+                        // Add process option only if there is at least one state
+                        if (this.loadStatesForProcess(process)) this.processOptions.push(processToShow);
```
AlexGuironnetRTE, line 208: What is the advantage of returning a boolean rather than testing this.statesPerProcesses.get(process.id) for existence and length directly?

AlexGuironnetRTE (commented):
Test ok (using "first" bundle, behaviour as expected) but I have two questions (see comments).

AlexGuironnetRTE (approved):

PR #836, opened 2021-03-10
 [OC-1553] sort by read even if config is set to not sort by read

freddidierRTE:


AlexGuironnetRTE (approved):
Tests OK in 4200
xit for unit test ok, was failing randomly on previous commit

PR #839, opened 2021-03-11
[OC-1531] Change rule of visiblity for cards with no endDate

quinarygio:
In Release notes:

In Tasks section:
[OC-1531] Change rule of visiblity for cards with no endDate to be visible when startDate is between start and end of chosen period.

[OC-1531]: https://opfab.atlassian.net/browse/OC-1531


freddidierRTE:
- Bug in Archives (or logging screen) : 
Card with no endDate --> Searching only with activeFrom does not work as wanted the card are always visible no matter the  value 

- For the examples to be coherent in src/test/utils/karate/cards : put an endDate set to the value of the older timespan . Only src/test/utils/karate/cards/postAnAlarmChartCard.feature will have no endDate

PR #844, opened 2021-03-12
[OC-1540] Bug fix in JAVA_OPTIONS in docker-entrypoint script

geppyz:
Moved JAVA_OPTIONS environment parameter in docker-entrypoint shell script. In the old position it was placed behind the -jar parameter. This caused java to skip the JAVA_OPTIONS and ignore the optional extra settings.

Signed-off-by: Gerben Danen <gerben.danen@alliander.com>


geppyz:
This caused problems with our Kafka enviroment, because certain configuration that was set with -D parameters was not read before this fix.

PR #849, opened 2021-03-16
[OC-1567] Activate nginx gzip compression

freddidierRTE:
In release note : 

[OC-1567] Activate nginx gzip compression

[OC-1567]: https://opfab.atlassian.net/browse/OC-1567


AlexGuironnetRTE:
Tested with the config/dev docker, the compression is really impressive!

before
after

PR #854, opened 2021-03-16
[OC-1509] Recurrent error message when posting user cards

vlo-rte:
[OC-1509](https://opfab.atlassian.net/browse/OC-1509) : Recurrent error message when posting user cards

### DIFF
--- a/ui/main/src/app/modules/usercard/usercard.component.ts
+++ b/ui/main/src/app/modules/usercard/usercard.component.ts
@@ -178,7 +178,8 @@ export class UserCardComponent implements OnDestroy, OnInit {
                         this.messageForm.get('process').disable();
                         this.messageForm.get('state').setValue(this.cardToEdit.card.state);
                         this.messageForm.get('startDate').setValue(getDateTimeNgbFromMoment(moment(this.cardToEdit.card.startDate)));
-                        this.messageForm.get('endDate').setValue(getDateTimeNgbFromMoment(moment(this.cardToEdit.card.endDate)));
+                        if (this.cardToEdit.card.endDate || this.endDateVisible)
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/usercard/usercard.component.ts
```diff
@@ -178,7 +178,8 @@ export class UserCardComponent implements OnDestroy, OnInit {
                         this.messageForm.get('process').disable();
                         this.messageForm.get('state').setValue(this.cardToEdit.card.state);
                         this.messageForm.get('startDate').setValue(getDateTimeNgbFromMoment(moment(this.cardToEdit.card.startDate)));
-                        this.messageForm.get('endDate').setValue(getDateTimeNgbFromMoment(moment(this.cardToEdit.card.endDate)));
+                        if (this.cardToEdit.card.endDate || this.endDateVisible)
+                            this.messageForm.get('endDate').setValue(getDateTimeNgbFromMoment(moment(this.cardToEdit.card.endDate)));
```
AlexGuironnetRTE, line 182: I'm not sure I understand the point of "|| this.endDateVisible" because if this.cardToEdit.card.endDate is undefined or null and this.endDateVisible is true, it will cause the value to be set with getDateTimeNgbFromMoment(moment(null)). What does moment(null) return ?

AlexGuironnetRTE (commented):
Tested on 4200, no error message anymore and the card is correctly edited.

I do have one question on the code though (see comment). Thanks!

ui/main/src/app/modules/usercard/usercard.component.ts
vlo-rte: It will return current datetime.

AlexGuironnetRTE: Clever ;-)

AlexGuironnetRTE (approved):

PR #857, opened 2021-03-17
[OC-1559] EntitiesRequiredToRespond constraint doesn't handle parent …

vlo-rte:
Release notes : 

In bugs section : 

[OC-1559](https://opfab.atlassian.net/browse/OC-1559) : EntitiesRequiredToRespond constraint doesn't handle parent entities

### DIFF
--- a/services/core/cards-publication/src/main/modeling/swagger.yaml
+++ b/services/core/cards-publication/src/main/modeling/swagger.yaml
@@ -350,8 +350,9 @@ definitions:
           type: string
         example: ["Dispatcher","Planner"]
         description: >-
-          List of entities that have to respond (should be a subset of entitiesAllowedToRespond). If present and not
-          empty, this list will be used for display in the card detail header instead of entitiesAllowedToRespond.
+          List of entities that have to respond. It will be taken into account in addition to entitiesAllowedToRespond
+          to determine if a user can send a response. If present and not empty, this list will be used for display in the
+          card detail header instead of entitiesAllowedToRespond.
       entityRecipients:
         description: List of entity recipients
         type: array
@@ -505,8 +506,9 @@ definitions:
           type: string
         example: ["Dispatcher","Planner"]
         description: >-
-          List of entities that have to respond (should be a subset of entitiesAllowedToRespond). If present and not
-          empty, this list will be used for display in the card detail header instead of entitiesAllowedToRespond.
+          List of entities that have to respond. It will be taken into account in addition to entitiesAllowedToRespond
+          to determine if a user can send a response. If present and not empty, this list will be used for display in the
+          card detail header instead of entitiesAllowedToRespond.
       title:
         description: Card i18n title
         $ref: '#/definitions/I18n'
--- a/services/core/cards-publication/src/test/java/org/lfenergy/operatorfabric/cards/publication/services/CardProcessServiceShould.java
+++ b/services/core/cards-publication/src/test/java/org/lfenergy/operatorfabric/cards/publication/services/CardProcessServiceShould.java
@@ -533,15 +533,11 @@ class CardProcessServiceShould {
         LocalTime nine = LocalTime.of(9, 0);
         LocalTime fifteen = LocalTime.of(17, 0);
 
-        // entitiesRequiredToRespond is excluded from the randomization (field will be null) because otherwise the
-        // the card would not abide by the constraint that entitiesRequiredToRespond should be a subset of
-        // entitiesAllowedToRespond.This will be tested in dedicated tests.
         EasyRandomParameters parameters = new EasyRandomParameters().seed(5467L).objectPoolSize(100)
                 .randomizationDepth(3).charset(forName("UTF-8")).timeRange(nine, fifteen)
                 .dateRange(today, tomorrow).stringLengthRange(5, 50).collectionSizeRange(1, 10)
                 .excludeField(named("data"))
                 .excludeField(named("parameters"))
-                .excludeField(named("entitiesRequiredToRespond"))
                 .excludeField(named("shardKey"))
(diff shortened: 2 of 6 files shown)


services/core/cards-publication/src/test/java/org/lfenergy/operatorfabric/cards/publication/services/CardProcessServiceShould.java
```diff
@@ -1078,7 +1078,7 @@ void processKeepChildCardsNull() {
     }
 
     @Test
-    void processCardWithEntitiesRequiredToRespondCorrectlySet() {
+    void processCardWithEntitiesRequiredToRespondSubsetOfEntitiesAllowedToRespond() {
```
AlexGuironnetRTE, line 1081: I don't think we should keep both tests as there is now no constraint on `entitiesRequiredToRespond`, people reading the code later might wonder why we're testing two cases (subset and not a subset). If I'm not mistaken there are no specific tests for other properties of this type, so I think that's OK if there is no test for this one.

src/test/api/karate/cards/postCardsEntitiesToRespond.feature
```diff
@@ -59,7 +59,6 @@ Feature: Post cards with entitiesAllowedToRespond and/or entitiesRequiredToRespo
     Then status 201
     And match response.count == 1
 
-  Scenario: Push cards non-compliant values of entitiesAllowedToRespond and entitiesRequiredToRespond
```
AlexGuironnetRTE, line 62: We should also remove the "with compliant values" in the remaining scenario. 

I was also going to say that we don't need the karate tests anymore now that there's no constraint (see comment above), but then they're handy to test the behaviour of the header in card details in the UI, so maybe we can keep them. What do you think?

ui/main/src/app/modules/cards/components/detail/detail.component.ts
```diff
@@ -457,22 +457,29 @@ export class DetailComponent implements OnChanges, OnInit, OnDestroy, AfterViewC
...
             // This will be overwritten by the block below if entitiesRequiredToRespond is set and not empty/null
             // This is to avoid repeating the creation of the allowed list
             this._listEntitiesToRespondForHeader = this.createEntityHeaderFromList(allowed);
 
-            this._userEntitiesAllowedToRespond = allowed.filter(x =>this.user.entities.includes(x));
+            this._userEntitiesAllowedToRespond = allowed.filter(x => this.user.entities.includes(x));
             console.log(new Date().toISOString(), ' Detail card - users entities allowed to respond = ', this._userEntitiesAllowedToRespond);
-
+            if (this._userEntitiesAllowedToRespond.length > 1)
+                console.log(new Date().toISOString(), 'Warning : user can respond on behalf of more than one entities, so response is disabled');
```
AlexGuironnetRTE, line 479:
```suggestion
                console.log(new Date().toISOString(), 'Warning : user can respond on behalf of more than one entity, so response is disabled');
```

services/core/cards-publication/src/main/modeling/swagger.yaml
```diff
@@ -350,8 +350,8 @@ definitions:
           type: string
         example: ["Dispatcher","Planner"]
         description: >-
-          List of entities that have to respond (should be a subset of entitiesAllowedToRespond). If present and not
-          empty, this list will be used for display in the card detail header instead of entitiesAllowedToRespond.
+          List of entities that have to respond. If present and not empty, this list will be used for display in the
```
AlexGuironnetRTE, line 353:
```suggestion
          List of entities that have to respond. It will be taken into account in addition to entitiesAllowedToRespond to determine if a user can send a response. If present and not empty, this list will be used for display in the
```

services/core/cards-publication/src/main/modeling/swagger.yaml
```diff
@@ -505,8 +505,8 @@ definitions:
...
```
AlexGuironnetRTE, line 508:
```suggestion
          List of entities that have to respond. It will be taken into account in addition to entitiesAllowedToRespond to determine if a user can send a response. If present and not empty, this list will be used for display in the
```

AlexGuironnetRTE (changes requested):

ui/main/src/app/modules/cards/components/detail/detail.component.ts
vlo-rte: Done.

services/core/cards-publication/src/main/modeling/swagger.yaml
vlo-rte: Done.

services/core/cards-publication/src/main/modeling/swagger.yaml
vlo-rte: Done.

services/core/cards-publication/src/test/java/org/lfenergy/operatorfabric/cards/publication/services/CardProcessServiceShould.java
vlo-rte: Done.

src/test/api/karate/cards/postCardsEntitiesToRespond.feature
vlo-rte: I propose to not remove these karate tests. They are already written and they help to avoid regression.

AlexGuironnetRTE: True, they're already written.

AlexGuironnetRTE (approved):
Thanks for the changes, all tests OK (checked header and presence of response button in question card example, using combinations of simple and parent entities).

PR #865, opened 2021-03-19
[OC-1505] Add parent entities to edition modal in Admin

quinarygio:
Release notes

In Tasks section:
[OC-1505] Add parent entities to edition modal in Admin

[OC-1505]: https://opfab.atlassian.net/browse/OC-1505


AlexGuironnetRTE (approved):
Tested on 4200.
Creating entities with parent entities OK
Editing entities OK
Editing name of parent entity is taken into account in children entities immediately.
Deleting parent entity doesn't trigger removal in children entities, but I think it's an issue with the backend that doesn't do the deletion in children entities, so I created OC-1581 to handle that.

PR #872, opened 2021-03-22
[OC-1581] Deleting parent entity should remove the reference in child entities

quinarygio:
Release notes

In Bugs section:

[OC-1581] Deleting an entity that is listed as parent entity in some children entities should cause it to be removed from said children

[OC-1581]: https://opfab.atlassian.net/browse/OC-1581


AlexGuironnetRTE (approved):
Build and test (dev mode) OK!

PR #873, opened 2021-03-23
[OC-1582] Fix bug when exporting to excel in logging screen

quinarygio:
Release notes 

In Bugs section:
[OC-1582] Fix bug when exporting to excel in logging screen

[OC-1582]: https://opfab.atlassian.net/browse/OC-1582


vlo-rte (changes requested):
As discussed in the daily, please can you set empty field if the data is not available ? (In order to be coherent with the logging screen). Thank you.

PR #886, opened 2021-03-24
[OC-1585] Authorization mode 'NONE' fix with CardService and Perimeters.

geppyz:
Added isAuthModeNone check to getSecurityHeader function. This fixes an issue while using authorization mode 'NONE' in combination with the checking of perimeters in the Cards Publication backend. The CardService has two functions that call the getSecurityHeader function. This causes the Authorization header to be overwritten, thereby causing the Cards Publication backend to not allow the user to publish a card. Bonus fix: added a space to a logger in the CardService to format the text the same as the other loggers.

Signed-off-by: Gerben Danen <gerben.danen@alliander.com>

### DIFF
--- a/ui/main/src/app/services/card.service.ts
+++ b/ui/main/src/app/services/card.service.ts
@@ -108,10 +108,14 @@ export class CardService {
 
     private getCardSubscription(): Observable<CardOperation> {
         // security header needed here as SSE request are not intercepted by our header interceptor
+        let securityHeader;
+        if (!this.authService.isAuthModeNone()) {
+            securityHeader = this.authService.getSecurityHeader();
+        }
         const eventSource = new EventSourcePolyfill(
             `${this.cardOperationsUrl}&notification=true`
             , {
-                headers: this.authService.getSecurityHeader(),
+                headers: securityHeader,
                 // if necessary , we cans set here  heartbeatTimeout: xxx (in ms)
             });
         return Observable.create(observer => {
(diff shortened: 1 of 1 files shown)


freddidierRTE:
I think it is better to not modify the header in postCard(card: CardForPublishing) in  ui/main/src/app/services/card.service.ts .
In fact it is not necessary to do that as there is the interceptor , it s a mistake in the code . 

So I made some test with : 

```
  postCard(card: CardForPublishing): any {
        return this.httpClient.post<CardForPublishing>(`${this.cardsPubUrl}/userCard`, card);
    }

``` 

I let you test if it solve the issue with the none mode

geppyz:
> postCard(card: CardForPublishing): any

I am doing some testing of this today. I'll check the results and change this PR.

geppyz:
I've added some changes to the cardservice. The security header was also beeing set at the creation of the observer used by the getSubscription function.

ui/main/src/app/services/card.service.ts
```diff
@@ -108,10 +108,14 @@ export class CardService {
 
     private getCardSubscription(): Observable<CardOperation> {
         // security header needed here as SSE request are not intercepted by our header interceptor
+        let securityHeader;
+        if (!this.authService.isAuthModeNone) {
```
freddidierRTE, line 112: in fact it is : ` if (!this.authService.isAuthModeNone())`  otherwise it does not work as it tests the existence of the function and not the return value

freddidierRTE (changes requested):

ui/main/src/app/services/card.service.ts
geppyz: Yes, fixed it in the new commit. Thanks for spotting that!

PR #912, opened 2021-03-31
[OC-1606] Add method isUserMemberOfAnEntityRequiredToRespond for template

quinarygio:
Release notes:

In Tasks:

[OC-1606] Add method isUserMemberOfAnEntityRequiredToRespond() for card template

[OC-1606]: https://opfab.atlassian.net/browse/OC-1606


freddidierRTE:
The value for userMemberOfAnEntityRequiredToRespond shall not be set based on card header as the card header reflect card.entityAllowedToRespond when card.entityRequiredToRespond is empty or null . 
So you must use the entityRequiredToRespond card field to set userMemberOfAnEntityRequiredToRespond

PR #931, opened 2021-04-09
[OC-1607] Remove reactive code from card publication

freddidierRTE:
Reactive code is not needed  in card publication service (overkill)
and it causes at least one bug: when external application take too
much time to respond it hangs some of other requests to card publication.

Release note : 
Bugs  : 
[OC-1607] Remove reactive code from card publication and solve a bug (hang of HTTP requests) 
+
IMPORTANT: This change some endpoint for sending card see https://opfab.github.io/documentation/current/docs/single_page_doc.html#_migration_guide_from_release_2_3_0_to_release_2_4_0
+

[OC-1607]: https://opfab.atlassian.net/browse/OC-1607


services/core/cards-publication/src/main/modeling/swagger.yaml
```diff
@@ -700,19 +677,37 @@ paths:
   /cards:
     post:
       tags:
-        - cards
+        - card
         - creation
       summary: publish cards
-      description: Publish one or more card to OperatorFabric. Be careful, to know if the cards were published well, then it is necessary to consult the report returned in the body of the request (do not rely only on the response code).
-      operationId: publishCards
+      description: Publish one card to OperatorFabric - Deprecated endpoint (use /cards/card instead)
```
AlexGuironnetRTE, line 683: Why is it necessary to change /cards to /cards/card? Couldn't we keep /cards and advertise the fact that now it only accepts one card? Projects that used to send several cards will have to update their code regardless of the name of the endpoint.
In addition, what bothers me is that if posting cards is managed via /cards/card, it won't be coherent with the corresponding DELETE and GET operations that are still attached to the /cards endpoint, right?

AlexGuironnetRTE (changes requested):
Karate tests OK (on config/docker)
Manual tests OK as well

I'm not sure about the renaming of the endpoint though (see comment).

services/core/cards-publication/src/main/modeling/swagger.yaml
freddidierRTE: you are right , i was wrong about the naming convention , thinking that as we post a card it shall be singular. After having a look on the net , the good way is to POST /cards and not /cards/Card as you said . 
I will correct it

AlexGuironnetRTE (approved):
As discussed, I created OC-1646 to change the response of the card creation endpoint: now that we can only post one card at a time there is no need for the card creation report with the count of successfully handled cards.

PR #934, opened 2021-04-11
[OC-1604] Acknowledge all cards : manage states with acknowledgmentAllowed set to "OnlyWhenResponseDisabledForUser"

vlo-rte:
In the acknowledge.service.ts file, I only left functions directly related to the ack (even when they are not strictly speaking requests).
And I created a new file, action.service.ts, which contains functions allowing to know if the user has the right to answer or not.
And I propose to make another ticket to do a refacto consisting of using this new action service, in the Detail component (OC-1659).

Release notes : 

In Tasks section : 

[OC-1604](https://opfab.atlassian.net/browse/OC-1604) : Acknowledge all cards : manage states with acknowledgmentAllowed set to "OnlyWhenResponseDisabledForUser"

### DIFF
--- a/ui/main/src/app/modules/cards/components/detail/detail.component.ts
+++ b/ui/main/src/app/modules/cards/components/detail/detail.component.ts
@@ -50,6 +50,7 @@ import {TimeService} from '@ofServices/time.service';
 import {AlertMessage} from '@ofStore/actions/alert.actions';
 import {MessageLevel} from '@ofModel/message.model';
 import {RightsEnum} from '@ofModel/perimeter.model';
+import {AcknowledgeService} from '@ofServices/acknowledge.service';
 
 
 declare const templateGateway: any;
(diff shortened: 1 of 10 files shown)


AlexGuironnetRTE (commented):
Code structure seems clear to me, tests clearly show the expected behaviour.

I tested using config/docker(docker-compose-cypress.sh) and ng serve, and using the test configuration and cards from resources, everything looks good but there is one case I would like to confirm: 

I changed the configuration of defaultProcess (src/test/resources/bundles/defaultProcess_V1/config.json) so that the property acknowledgmentAllowed was set to OnlyWhenResponseDisabledForUser for the state questionState. 
Since the issue states the following :
>  "A card with a state having acknowledgmentAllowed set to OnlyWhenResponseDisabledForUser (in config.json) must be acknowledged only if user cannot respond. 

I was expecting the "Acknowledge all" link to acknowledge this card if clicked after the countdown has expired (because then the user can't respond anymore as the respond button is disabled), but it's not the case => is it what is intended ?
On the other end, it is coherent with the behaviour of the acknowledge button, which doesn't appear in this case.

There was one error message in the console but it doesn't seem related to this PR so I logged OC-1660.

AlexGuironnetRTE:
See OC-1604_B for final implementation

PR #935, opened 2021-04-12
[OC-1622] Set a value for state.description fields in all card examples

quinarygio:
A propose to add nothing in release notes


freddidierRTE:
To add in : 
src/test/resources/bundles/gridCooperation/config.json  (use same as value for name) 
src/test/resources/bundles/userCardExamples/config.json   (use same as value for name) 
src/test/resources/bundles/userCardExamples2/config.json   (use same as value for name) 
src/test/resources/bundles/userCardExamples3/config.json   (use same as value for name)

PR #942, opened 2021-04-13
[OC-1641] When card is updated and selected in the feed, it shall be seen as unread

freddidierRTE:
Release note: 

Feature : 
[OC-1641] When card is updated and selected in the feed, it shall be seen as unread

[OC-1641]: https://opfab.atlassian.net/browse/OC-1641


AlexGuironnetRTE:
As discussed, updated description of fixed problem:

A card appears in the feed, the user clicks it. So far, it remains unread (bold) and should become “read” only after either of the following actions:
- closing the card details
- exiting the feed screen
- switching to another card

But currently, if the card is updated (with the same id), it becomes read.
Expected behaviour : it should stay “unread”

AlexGuironnetRTE (commented):
Tests OK (docker + ng serve)

PR #951, opened 2021-04-15
[OC-1634] When changing perimeter attach to a group it is not propagate to the card-publication service

quinarygio:
Release notes

In Bugs section:
[OC-1634] When changing perimeter attach to a group it is not propagate to the card-publication service

[OC-1634]: https://opfab.atlassian.net/browse/OC-1634


AlexGuironnetRTE (approved):
Changes ok
Tests with dev docker + run all OK

PR #953, opened 2021-04-15
[OC-1657] Card deletion result should be displayed with an alert message

quinarygio:
Release notes

In Tasks section:

[OC-1657] Card deletion result should be displayed with an alert messageage instead of popover

[OC-1657]: https://opfab.atlassian.net/browse/OC-1657

### DIFF
--- a/ui/main/src/app/modules/cards/components/detail/detail.component.html
+++ b/ui/main/src/app/modules/cards/components/detail/detail.component.html
@@ -135,43 +135,11 @@
                 (click)="confirmDeleteCard()" translate>button.delete</button>
             </div>
           </ng-template>
-
-          <ng-template #cardDeletedWithNoErrorPopup let-modal>
-            <div class="modal-header text-center">
-              <span class="modal-title w-100" translate> userCard.deleteCard.cardDeletedWithNoError </span>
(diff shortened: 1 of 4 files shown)


ui/main/src/app/modules/cards/components/detail/detail.component.html
```diff
@@ -135,39 +135,6 @@
...
-              </div>
-            </div>
-            <div class="modal-body text-center">
-              <p translate> userCard.deleteCard.error.impossibleToDeleteCard </p>
-
-              <button id="opfab-card-details-delete-error-btn-close" type="button" class="opfab-btn"
-                (click)="modal.dismiss()" translate>button.ok</button>
-            </div>
-          </ng-template>
-        </div>
```
vlo-rte, line 168: I think this "< / div>" shouldn't be deleted (line 168). If you are ok, can you modify please ?
Thank you.

vlo-rte (changes requested):

ui/main/src/app/modules/cards/components/detail/detail.component.html
quinarygio: Done

PR #955, opened 2021-04-19
[OC-1654] After an update of a question card, the entities that have already answered shall be orange in card header

quinarygio:
Release notes

In Tasks section:
[OC-1654] After an update of a question card, the entities that have already answered shall be orange in card header
   Note: the entities will be marked in green only  when they answer to the updated version of card

[OC-1654]: https://opfab.atlassian.net/browse/OC-1654


AlexGuironnetRTE (approved):
Tests OK

PR #962, opened 2021-04-21
[OC-1639] Use user perimeter to populate services/process/state list in archives/logging/monitoring/feedconfiguration

vlo-rte:
Release notes : 

In Features section : 

[OC-1639](https://opfab.atlassian.net/browse/OC-1639) : Use user perimeter to populate services/process/state list in archives/logging/monitoring/feedconfiguration

Tests performed : 

For notification configuration screen : 
- missing variable : no change for the screen : OK
- variable set to false : no change for the screen : OK
- variable set to true : 
         - only the states are present for which a Receive right is granted : OK
         - if I remove the userCardExamples3 perimeter from the Dispatcher group (to which operator1 belongs) then the process is not displayed at all : OK (if no more state present for a process then do not display the process)
         - if I delete the gridCooperation and question perimeters from the Dispatcher group then the process group is not displayed : OK (if no more process present for a process group then do not display the process group)

Same tests for archives/logging/monitoring screen. With and without process groups.

### DIFF
--- a/ui/main/src/app/modules/feedconfiguration/feedconfiguration.component.html
+++ b/ui/main/src/app/modules/feedconfiguration/feedconfiguration.component.html
@@ -17,10 +17,10 @@
             <h5 class="font-weight-bold">{{processGroup.groupLabel}}</h5>
             <div class="row">
                 <div class="col-md" *ngFor="let processIdAndLabel of processGroup.processes; ">
-                    <div *ngIf="processesStatesLabels.get(processIdAndLabel.processId)">
-                        <p class="font-weight-bold">{{processesStatesLabels.get(processIdAndLabel.processId).processLabel}}</p>
+                    <div *ngIf="processesStatesLabels.get(processIdAndLabel.processId) as processData">
+                        <p class="font-weight-bold">{{processData.processLabel}}</p>
                         <div style="margin-left: 20px;" >
-                            <div class="row" formArrayName="processesStates" *ngFor="let state of processesStatesLabels.get(processIdAndLabel.processId).states; ">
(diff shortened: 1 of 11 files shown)


ui/main/src/app/modules/feedconfiguration/feedconfiguration.component.html
```diff
@@ -17,10 +17,10 @@
             <h5 class="font-weight-bold">{{processGroup.groupLabel}}</h5>
             <div class="row">
                 <div class="col-md" *ngFor="let processIdAndLabel of processGroup.processes; ">
-                    <div *ngIf="processesStatesLabels.get(processIdAndLabel.processId)">
-                        <p class="font-weight-bold">{{processesStatesLabels.get(processIdAndLabel.processId).processLabel}}</p>
+                    <div *ngIf="processesStatesLabels.get(processIdAndLabel.processId) as processData">
+                        <p class="font-weight-bold">{{processData.processLabel}}</p>
                         <div style="margin-left: 20px;" >
-                            <div class="row" formArrayName="processesStates" *ngFor="let state of processesStatesLabels.get(processIdAndLabel.processId).states; ">
```
AlexGuironnetRTE, line 23: That's much nicer to read, I didn't know that thanks!

AlexGuironnetRTE (approved):
Everything looks good, tests are ok!
Just a minor clarification in the doc and comment to avoid any misunderstanding.

Sonar is not happy about a few duplicated lines but from the look of it, it would be better to handle this as part of a more global refactoring of archives/logging/monitoring.

PR #976, opened 2021-04-26
[OC-1674] Editing a usercard : multi-select component for state field is not set with the right value

quinarygio:
Release notes:

In Bugs section:
[OC-1674]  Editing a usercard : multi-select component for state field is not set with the right value

[OC-1674]: https://opfab.atlassian.net/browse/OC-1674


AlexGuironnetRTE (approved):
Tests OK (on example from ticket and others)

PR #984, opened 2021-04-28
[OC-255] Configuring Gradle build to publish to staging repository

AlexGuironnetRTE:
I was going to say this should go in the tasks section but maybe we should advertise the fact that the client jars are now available in Maven Central as it's going to impact client applications?

What about this, in the features section:

 OC-255, OC-1513: Making the client library available in Maven Central
+
Starting from this release, you can now pull the client library as a regular dependency rather than building the jar locally. The clientApp example is now relying on these published jars.


AlexGuironnetRTE:
I decided to address OC-1513 (Have clientApp depend on published jars) as well on the same branch as it was a way to fully test that the publication had worked.

Regarding the issue with relying on publication to Maven local for development environments where the signing information (GPG key and passphrase) is not available (as discussed on Slack), I decided to go with a standard gradle dependency as a fallback (option 1 below), but don't hesitate to comment if you feel that another option would be preferable.

Proposed solutions:
1) If version is SNAPSHOT and the GPG information is not present in the gradle.properties file, resort to the old way of doing things (gradle compile dependency). Downside: not sure if it is possible, and it means that depending on the setup the clientApp.SNAPSHOT.jar contains either a signed or unsigned jar.
2) If version is SNAPSHOT, use the latest released version on Central. Pros: No compilation needed, and you're testing against the latest "publicly available" versions, which would be good for people outside the project, but not so great for our testing
3) Change things so that publication to maven local doesn't need signing. Pros : it's only local so there's no security issue that would require signing, Cons : If you use local publication as a dry-run you won't detect issues with signing anymore

AlexGuironnetRTE:
Merging as discussed during our daily meeting.

PR #1000, opened 2021-04-30
[OC-1648] Add a control of IP address for user access

quinarygio:
Release notes

In Features section:
[OC-1648] Add a control of IP address for user access. For each user it is possible to configure a list of authorized source ip addresses by setting the `authorizedIPAddresses` field in User object.

[OC-1648]: https://opfab.atlassian.net/browse/OC-1648

### DIFF
--- a/config/dev/common-dev.yml
+++ b/config/dev/common-dev.yml
@@ -1,3 +1,5 @@
+server:
+  forward-headers-strategy: framework
 spring:
   rabbitmq:
     host: localhost
--- a/config/dev/nginx.conf.template
+++ b/config/dev/nginx.conf.template
@@ -1,5 +1,6 @@
 # docker-compose DNS used to resolved keycloak services
 resolver 127.0.0.11 ipv6=off;
+
 server {
   listen 80;
   server_name localhost;
@@ -99,6 +100,7 @@ server {
     }
     proxy_set_header Host $http_host;
     proxy_pass http://${MY_DOCKER_HOST}:2100;
+    proxy_set_header X-Forwarded-For $remote_addr;
   }
 
   location /users {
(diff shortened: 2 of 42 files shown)


freddidierRTE (changes requested):
- For each service when connection is refused , add a log.info with IP and login of the user trying to connect 
- CurrentUserWithPerimetersInternal endpoint shall not be accessible thought nginx (use perhaps endpoint internal/CurrentUserWithPerimeter and filter it in nginx conf, if we need to hide other endpoint we will use this private path ) 
- Add the following Karate test scenario in a new feature file :
   * Update an existing user with a list of 2 authorized IP 
   * Call once an endpoint of each service  directly (i.e not through nginx) with an X-Forwarded-For not in the list of authorized IP and check connection is refused (be careful , the information has to propagate from user service to the other , i do not know if it will be fast enough to be done between the karate tests ... , starts with calling user service perhaps ? ) 
   * Call once the same endpoint of each service directly with an X-Forwarded-For in the list of authorized IP and check the connection is accepted 
   * Call once the same endpoint of each service via nginx  with an X-Forwarded-For in the list of authorized IP and check the connection is refused 
   * Remove the two IP adresses 
   *  Call once the same endpoint of each service directly (i.e not through nginx) with an X-Forwarded-For not in the list of authorized IP and check connection is accepted 
   * Call once the same endpoint of each service through nginx with no X-Forwarded-For and check connection is accepted 

Tip : to limit the number of lines in karate file you can use the keyword "Examples" as in the following example

```
# Setup
        Scenario Outline: Add new entities

            Given url opfabUrl + 'users/entities'
            And header Authorization = 'Bearer ' + adminToken
            And request <entity>
            When method post
            Then status <expected>

            Examples:
                |entity          |expected|
                |rootEntity      |201     |
                |childEntity     |201     |
                |grandChildEntity|201     |
                |singleEntity    |201     |

```

freddidierRTE:
Two last points : 
- set the configuration in the common-docker.yml instead of setting it in each service configuration file 
- avoid duplication in code (cf sonar analysis) , use for example a common lib

PR #1024, opened 2021-05-07
[OC-1653] See on feed when there is a response form the entity of the user 

freddidierRTE:
Add a blue arrow on feed card when there is a response from one of the entity of the current user 
Add a new cypress test 
Rename field lastCard in lastCardLoaded

Release note : 

Feature : 
* OC-1653 See on feed when there is a response form the entity of the user


quinarygio:
Test ok.

Fix Copyright header in :
ui/main/src/tests/helpers.ts
ui/main/src/app/services/reminder/reminder.service.ts

PR #1070, opened 2021-05-18
[OC-1723] UserCard: check that Lttd is not before start date and not after endDate

quinarygio:
Release notes

In Tasks section:
[OC-1723] UserCard: check that Lttd is not before start date and not after endDate


ui/main/src/assets/i18n/fr.json
```diff
@@ -222,7 +222,10 @@
       "notAvailable": "Fonctionnalité non disponible",
       "templateError" : "Erreur technique : template non conforme",
       "impossibleToSendCard" : "Impossible d'envoyer la carte pour des raisons techniques",
-      "endDateBeforeStartDate": "La date de fin doit être postérieure à la date de début"
+      "endDateBeforeStartDate": "La date de fin doit être postérieure à la date de début",
+      "lttdBeforeStartDate": "La date limite de réponse doit être postérieure à la date de début",
+      "lttdAfterEndDate": "La date limite de réponse doit être antérieur à la date de fin"
```
freddidierRTE, line 227: La date limite de réponse doit être antérieur**e**

freddidierRTE:
Just a translation correction do to (see comment in fr.json)

ui/main/src/assets/i18n/fr.json
quinarygio: Done

PR #1074, opened 2021-05-18
[OC-1728]  Acknowledging all cards shouldn't change their read status

AlexGuironnetRTE:
I reverted the fix I made in OC-1715 so updates triggered by acknowledgement (either single card using the button or all card with the ack all feature) no longer set the card to hasBeenRead = true.
In the case of the single card ack, the ack button triggers a closing of the details which in turns triggers an update of the card with hasBeenRead = true, and which is also propagated to the back.
In the case of the ack all features however, no changes are made to the hasBeenRead property of the card, so the user can tell which card he/she has actually read.

Tests (on 4200 chrome, with cypress-flavoured docker):
- Incoming cards (with send6cards) trigger sounds
- Ack all or single card ack don't trigger sounds
- Ack all sets all cards to acknowledged but read cards stay read and unread stay unread
- Reminders do trigger sounds (even if the original card had been read and/or acknowledged)
- No sounds are triggered for cards that are not visible in the feed (for example because of filter on severity or ack). This stands for all types of cards : new cards, new responses, reminders, etc. 

In release notes, under bug section: 

* OC-1728 : Acknowledging all cards shouldn't change their read status

### DIFF
--- a/ui/main/src/app/services/sound-notification.service.spec.ts
+++ b/ui/main/src/app/services/sound-notification.service.spec.ts
@@ -8,7 +8,7 @@
  */
 
 
-import { TestBed } from '@angular/core/testing';
+import {TestBed} from '@angular/core/testing';
 
 import { SoundNotificationService } from './sound-notification.service';
 import {StoreModule} from "@ngrx/store";
@@ -234,7 +234,7 @@ describe('SoundNotificationService', () => {
 
   /* For these tests, settings configuration property to true for all severities */
 
-  it('should not play sound if updated card is not visible', () => {
+  it('should not play sound if updated card is not visible, even if it is reminder', () => {
 
     configAllSeveritiesOn();
 
@@ -246,7 +246,7 @@ describe('SoundNotificationService', () => {
 
   });
 
-  it('should play sound if updated card is visible, even if it is not freshly published (', () => {
+  it('should play sound if updated card is reminder and is visible, even if it is not freshly published (', () => {
 
     configAllSeveritiesOn()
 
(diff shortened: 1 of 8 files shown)


ui/main/src/app/services/sound-notification.service.spec.ts
```diff
@@ -109,55 +112,162 @@ describe('SoundNotificationService', () => {
...
+
+    const today = new Date().getTime();
+    const card = getOneRandomLightCard({ publishDate: today});
+
+    service.handleUpdatedCard(card,UpdateTrigger.REMINDER,[]);
     expect(playSound).not.toHaveBeenCalled();
 
   });
 
+  it('should play sound if updated card is visible, even if it is not freshly published (', () => {
```
freddidierRTE, line 249: ... if udpate card **is reminder and** is visible , even if ....

ui/main/src/app/services/sound-notification.service.spec.ts
```diff
@@ -109,55 +112,162 @@ describe('SoundNotificationService', () => {
...
+    const card = getOneRandomLightCard({ publishDate: today});
+
+    service.handleUpdatedCard(card,UpdateTrigger.REMINDER,[card.id]);
+    expect(playSound).not.toHaveBeenCalled();
+
+  });
+
+  /* For these tests, settings configuration property to true for all severities */
+
+  it('should not play sound if updated card is not visible', () => {
```
freddidierRTE, line 237: ... if udpate card **is reminder and** is not visible

freddidierRTE:
Sonar warning 

https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&pullRequest=1074&resolved=false&types=CODE_SMELL

ui/main/src/app/services/sound-notification.service.spec.ts
AlexGuironnetRTE: You are right, fixed it.

ui/main/src/app/services/sound-notification.service.spec.ts
AlexGuironnetRTE: You are right, fixed it.

PR #1122, opened 2021-05-21
[OC-1736] BUG when child card received before parent card

freddidierRTE:
BUG : When loading application, it can happen that a child card is received
from back before his parent card. In this case when the child card is
a response from an entity of the current user, the blue response arrow
is not seen in feed as it should.

Release note : 

Bugs : 
OC-1736 Correct bug when child card is receive before parent card

### DIFF
--- a/ui/main/src/app/store/effects/card-operation.effects.ts
+++ b/ui/main/src/app/store/effects/card-operation.effects.ts
@@ -18,6 +18,7 @@ import {
     LoadLightCard,
     LoadLightChildCard,
     LoadLightParentCard,
+    NoopAction,
     UpdateALightCard
 } from '@ofActions/light-card.actions';
 import {Store} from '@ngrx/store';
@@ -34,6 +35,7 @@ import {UserService} from '@ofServices/user.service';
 @Injectable()
 export class CardOperationEffects {
 
+    private  orphanedChildCardsFromCurrentEntity : Set <string>  = new Set();
 
     constructor(private store: Store<AppState>,
                 private actions$: Actions,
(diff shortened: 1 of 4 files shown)


ui/main/src/app/store/effects/card-operation.effects.ts
```diff
@@ -49,26 +51,47 @@ export class CardOperationEffects {
...
             map(([lightCard, state]) => {
-                    if (!!lightCard.parentCardId) {
-                        return new LoadLightChildCard({lightCard: lightCard,isFromCurrentUserEntity : this.doesChildCardIsFromCurrentUserEntity(lightCard)});
-                    }
-                    else {
-                        let card: LightCard  = {...lightCard} ;
-                        card.hasChildCardFromCurrentUserEntity = this.doesNewCardVersionHasChildFromCurrentUserEntity(state.entities[card.id],card);
-                        return new LoadLightParentCard({lightCard: card});
+                if (!!lightCard.parentCardId) {
+                    const isFromCurrentUser = this.doesChildCardIsFromCurrentUserEntity(lightCard);
```
AlexGuironnetRTE, line 55:
```suggestion
                    const isFromCurrentUser = this.isChildCardFromCurrentUserEntity(lightCard);
```

ui/main/src/app/store/effects/card-operation.effects.ts
```diff
@@ -49,26 +51,47 @@ export class CardOperationEffects {
...
+                        }
+    
                     }
+                    return new LoadLightChildCard({lightCard: lightCard, isFromCurrentUserEntity: isFromCurrentUser});
                 }
-            )
+
+                else {
+                    let card: LightCard = {...lightCard};
+                    card.hasChildCardFromCurrentUserEntity = this.doesCardHasChildFromCurrentUserEntity(state.entities[card.id], card);
```
AlexGuironnetRTE, line 73: I think creating a variable for state.entities[card.id] (with the appropriate name, maybe currentLightCardState or something to that effect) rather than using it directly as a paramter would help code readability.
The "entities" naming (coming from ngrx entities) can be confusing in a method referring to "currentUserEntity".

ui/main/src/app/store/effects/card-operation.effects.ts
```diff
@@ -49,26 +51,47 @@ export class CardOperationEffects {
...
+                             // and that this child if from current user entity
+                             // this situation can arise as we do not have the card ordered when received from
+                             // the back via subscription (i.e a child can be received  before his parent)
+                            this.orphanedChildCardsFromCurrentEntity.add(lightCard.parentCardId);
+                            // do not need to load this child card
+                            return new NoopAction();
+                        }
+    
                     }
+                    return new LoadLightChildCard({lightCard: lightCard, isFromCurrentUserEntity: isFromCurrentUser});
```
AlexGuironnetRTE, line 68: I didn't manage to understand why the orphaned child card situation should only be handled in the case it is from the current user?

ui/main/src/app/store/effects/card-operation.effects.ts
```diff
@@ -49,26 +51,47 @@ export class CardOperationEffects {
...
+                    const isFromCurrentUser = this.doesChildCardIsFromCurrentUserEntity(lightCard);
+                    if (isFromCurrentUser) {
+                        if (!state.entities[lightCard.parentCardId]) {
+                             // if parent does not exist yet keep in memory the information that the card with id=parentCardId has a child
+                             // and that this child if from current user entity
+                             // this situation can arise as we do not have the card ordered when received from
+                             // the back via subscription (i.e a child can be received  before his parent)
+                            this.orphanedChildCardsFromCurrentEntity.add(lightCard.parentCardId);
+                            // do not need to load this child card
+                            return new NoopAction();
```
AlexGuironnetRTE, line 64: I don't think you need to define a dummy action, I changed the code to just "return;" and tried running the unit tests, and they passed. No complaint from ng serve either.

AlexGuironnetRTE (changes requested):
I haven't been able to reproduce the issue to test but I have a few remarks on the code.

ui/main/src/app/store/effects/card-operation.effects.ts
freddidierRTE: it complains in the console log

ui/main/src/app/store/effects/card-operation.effects.ts
freddidierRTE: The only use of child card in the state is when you have a detail card open and a child for this card arrives (the detail card get the child form the state via the lastCardLoaded value) . 

In the previous code , the parent card is not loaded so it is not open in the UI , so we do not need to had the child in the store

AlexGuironnetRTE (approved):

PR #1132, opened 2021-05-26
[OC-1672] Reload the UI configuration when bundles has changed

quinarygio:
Release notes

In Features section:
[OC-1672] Reload the UI configuration when bundles has changed

[OC-1672]: https://opfab.atlassian.net/browse/OC-1672

### DIFF
--- a/ui/main/src/app/store/effects/processes.effects.ts
+++ b/ui/main/src/app/store/effects/processes.effects.ts
@@ -1,0 +1,41 @@
+/* Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {Injectable} from '@angular/core';
+import {Actions, createEffect, ofType} from '@ngrx/effects';
+import {Observable, timer} from 'rxjs';
+import {ProcessesActionTypes} from '@ofActions/processes.actions';
+import {catchError, debounce, map} from 'rxjs/operators';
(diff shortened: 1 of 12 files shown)


ui/main/src/app/store/effects/processes.effects.ts
```diff
@@ -0,0 +1,42 @@
...
+            debounce(() => timer(10000)),
+            map(() => {
+                merge(
+                    this.service.loadAllProcesses(),
+                    this.service.loadProcessGroups(),
+                    this.service.areTranslationsLoaded()
+                ).subscribe(() => {});
+                    return of();
+                }
+            ),
```
AlexGuironnetRTE, line 35:
```suggestion
                        map(() => {
                    this.service.loadAllProcesses().subscribe();
                    this.service.loadProcessGroups().subscribe();
                }
            ),
```
Since we don't want to do any action once the loading is complete, I'm not sure the `merge` is necessary here, but maybe I'm missing the point?
Also, it looks like we can do without the subscription to `areTranslationsLoaded` as the loading of translations is triggered as part of `loadAllProcesses`, and we don't care about it being finished or not since we're not doing anything afterwards.
I've tried the above code and I looked like it worked on "passing" cases, but I don't know about error management as I haven't been able to provoke an error.

AlexGuironnetRTE (changes requested):
I've made a few manual tests and everything looks good except that the templates are not reloaded, is that intentional?

Other than that, I just had one remark on the effect code (see comment).

Also, there are some [Sonar code smells](https://sonarcloud.io/project/issues?id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&open=AXmnqJdPDcaWWieWH6j3&pullRequest=1132&resolved=false&types=CODE_SMELL) (but don't worry about the one for rule "is already declared in the upper scope", it's the same issue as for other actions).

ui/main/src/app/store/effects/processes.effects.ts
quinarygio: Done

AlexGuironnetRTE (approved):

PR #1140, opened 2021-05-28
[OC-1737] Do not block UI when loading a lot of cards

freddidierRTE:
to send a bunch of cards  use `/src/test/resources/cardsForStressTests/send6xN_TestCards.sh N`  with N being a number , the number of cards send will be 6 x N 

Release note : 

In task : 

[OC-1737] Do not block UI when loading a lot of cards

[OC-1737]: https://opfab.atlassian.net/browse/OC-1737


ui/main/src/app/modules/feed/feed.component.ts
```diff
@@ -26,26 +27,32 @@ export class FeedComponent implements OnInit {
...
             dow: 6, // First day of week is Saturday
             doy: 12 // First week of year must contain 1 January (7 + 6 - 1)
         }});
+
+        
+        this.lightCards$ = this.lightCardsService.getFilteredAndSortedLightCards().pipe(
+            delay(0), // Solve error: 'Expression has changed after it was checked' --> See https://blog.angular-university.io/angular-debugging/
+            map(cards => {
+                this.totalNumberOfLightsCards = cards.length;
+                return cards.slice(0, 100)
```
AlexGuironnetRTE, line 50: It would be better to make that "100" the same variable as the threshold for display in the template below.

ui/main/src/app/services/card.service.ts
```diff
@@ -74,7 +74,16 @@ export class CardService {
 
 
     public initCardSubscription() {
-        this.getCardSubscription()
+
+        // The use of concatMap + timer is to avoid having the browser stuck when 
+        // a lot of card is arriving. (It permit the browser to execute
```
AlexGuironnetRTE, line 79:
```suggestion
        // a lot of card is arriving. (It allows the browser to execute 
```

AlexGuironnetRTE: Hehe good thing you put a comment... otherwise I can see us puzzling over this one for a while in a few months ;-)

ui/main/src/app/services/lightcards.service.ts
```diff
@@ -0,0 +1,185 @@
...
+            })
+        ).subscribe((lightCards) => this.filteredLightCards.next(lightCards));
+    }
+
+ // --------------------
+ // When an flow of card is coming, for performance reasons , we do not want to update the card list 
+ // every time  a card is arriving so we wait for the end of the flow of cards.
+ // But if it takes too long, we want to show something so every seconds we make a rendering 
+ // even if the flow is still continuing.
+ // Do do that we combine a debounce waiting for the end of the flow and an interval to get the card list every seconds
```
AlexGuironnetRTE, line 83:
```suggestion
 // But if it takes too long, we want to show something so every second we make a rendering 
 // even if the flow is still continuing.
 // To do that we combine a debounce waiting for the end of the flow and an interval to get the card list every second 
```

ui/main/src/app/services/lightcards.service.ts
```diff
@@ -0,0 +1,185 @@
...
+        );
+    }
+
+    private getLightCardDebounce(): Observable<any> {
+        return this.store.pipe(
+            select(feedSelectors.selectFeed),
+            debounceTime(200),
+            tap(() => this.lastDebounce = (new Date()).valueOf())
+        );
+    }
```
AlexGuironnetRTE, line 103: Wouldn't the [throttleTime](https://rxjs.dev/api/operators/throttleTime) operator with 200 as parameter do the trick?
It would let lightcard updates pass as long as there are less than 1 every 200ms, then take one every 200ms if there are more.
In the end, you emit at most at a rate of 1 every 200ms.

AlexGuironnetRTE: But after all it works so... if it ain't broke don't fix it ;-)

AlexGuironnetRTE (changes requested):
My manual tests were ok (including sounds), I just have a few remarks (see comments).

ui/main/src/assets/i18n/fr.json
```diff
@@ -102,7 +102,8 @@
     "lttdFinished": "Terminée",
     "received": "Reçu le",
     "at" : "à",
-    "from": "de"
+    "from": "de",
+    "moreCards": "cartes en plus"
```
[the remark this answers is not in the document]
freddidierRTE: ok cartes supplémentaires

src/test/resources/cardsForStressTests/send6xN_TestCards.sh
```diff
@@ -0,0 +1,32 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of the OperatorFabric project.
+
+# This starts by moving to the directory where the script is located so the paths below still work even if the script
+# is called from another folder
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
```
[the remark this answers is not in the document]
freddidierRTE: OK

ui/main/src/app/modules/feed/components/card-list/card-list.component.html
```diff
@@ -22,6 +22,9 @@
         <of-light-card [lightCard]="currentLightCard" [open]="currentLightCard.id === (selection|async)"></of-light-card>
       </div>
     </div>
+    <div *ngIf="totalNumberOfLightsCards>100" style="text-align:right;font-weight: bold;margin-top:0px;margin-right:10px;">
```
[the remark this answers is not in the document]
freddidierRTE: ok , i will change the code to avoid the "magic number"

ui/main/src/app/modules/feed/feed.component.ts
freddidierRTE: ok

ui/main/src/app/store/states/feed.state.ts
```diff
@@ -9,20 +9,11 @@
...
- *  * EntityState of LightCard
- *  * selectedCardId: the currently selected card id
- *  * lastCardLoaded the last card added / updated to the feed
- *  * message: last message during state processing
- *  * filters: a collection of filter to apply to the rendered feed
- *  * sortBySeverity: Indicates whether the cards in the feed should be sorted by severity before being sorted by
- *    publishDate (desc = latest first)
- */
+
 export interface CardFeedState extends EntityState<LightCard> {
```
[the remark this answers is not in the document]
freddidierRTE: no just a remainder of the old code, i will remove

ui/main/src/app/services/card.service.ts
freddidierRTE: ok

ui/main/src/app/services/lightcards.service.ts
freddidierRTE: ok

ui/main/src/app/services/lightcards.service.ts
freddidierRTE: it will not have the behavior i want : 

- wait for the end of flow (debounce) 
- if flow is  too long, update at least every 1000ms  (sample(interval))
- if a few card arrives , do not wait 1000ms , process  after 200ms (debounce) 

using throttleTime(200) will process the flux every 200s instead of every 1000ms if there is a lot of card 
using throttleTime(1000) will wait 1000ms even if we have one card to process

ui/main/src/app/store/states/feed.state.ts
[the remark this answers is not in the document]
freddidierRTE: in fact , it is necessary otherwise we need to implement custom methods to add/update/delete  lightcard in the list of lightcards.  I do not think we will gain that much performance implementing it custom.

PR #1150, opened 2021-06-01
[OC-1740] Use entity description attribute for recipient list in user card as an option

quinarygio:
Release notes

In Features section:
[OC-1740] Use entity description attribute for recipient list in user card as an option

[OC-1740]: https://opfab.atlassian.net/browse/OC-1740


freddidierRTE:
Add the field in the documentation : https://opfab.github.io/documentation/current/deployment/index.html#ui_properties

PR #1170, opened 2021-06-08
[OC-1739] Simplify Gradle dependency management

AlexGuironnetRTE:
The aim of this issue was to simplify the management of our dependencies' versions to rely on Gradle features rather than "homemade" solutions, in the hope that it will be less likely to break with future Gradle updates and [so the Renovate bot is able to pick up and manage these dependencies as well](https://github.com/renovatebot/renovate/discussions/10136).

Changes:
- Most dependencies where already declared in the root build script, I moved the few that were scattered in other scripts there as well.
- Instead of using the versions.properties file and manually loading it into a variable, we define the versions in the matching gradle.properties file as explained [here](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties).

In the release notes, under tasks:
OC-1739: Simplify Gradle dependency management


AlexGuironnetRTE:
While looking into it I also saw several things that I think make our build needlessly complex (and possibly longer) so I created these tickets with a low priority:

- OC-1771
- OC-1773
- OC-1774

PR #1191, opened 2021-06-09
[OC-1671] UI : Telling the user that his config has changed. Remove spring cloud bus dependency

quinarygio:
Release notes:

In Features section:
[OC-1671] UI : Telling the user that his config has changed. 
Note: on the backend the notifications of users updates are distributed using rabbitMq exchanges instead of spring cloud bus. 

[OC-1671]: https://opfab.atlassian.net/browse/OC-1671


services/core/cards-consultation/src/test/java/org/opfab/cards/consultation/application/configuration/ThreadPoolTaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Configuration
+public class ThreadPoolTaskSchedulerConfiguration {
+    
+    @Bean
+    public ThreadPoolTaskScheduler taskScheduler(){
+        return new ThreadPoolTaskScheduler();
+    }
+}
```
AlexGuironnetRTE, line 22: Instead of creating a duplicate configuration class for the tests I think you can just import the "real" configuration in `services/core/cards-consultation/src/test/java/org/opfab/cards/consultation/application/IntegrationTestApplication.java` using the @Import annotation.
This way there is no risk of the test configuration not being the same as the real one.

ui/main/src/app/services/card.service.ts
```diff
@@ -34,6 +34,7 @@ import {
 } from '@ofActions/light-card.actions';
```
AlexGuironnetRTE, line 34: Sonar is complaining that `import * as moment from 'moment';` is unused (even though I don't know why it considers it as a new code smell when you didn't touch it).

ui/main/src/app/store/actions/user.actions.ts
```diff
@@ -19,7 +19,8 @@ export enum UserActionsTypes {
     CreateUserApplicationOnSuccess =    '[User] Create the User in the application on success',
     CreateUserApplicationOnFailure =    '[User] Create the User in the application on failure',
     QueryAllEntities =                  '[User] Ask to fetch all entities',
-    LoadAllEntities =                   '[User] Load all entities'
+    LoadAllEntities =                   '[User] Load all entities',
+    UserConfigChange =                  '[User] USer config changed'
```
AlexGuironnetRTE, line 23:
```suggestion
    UserConfigChange =                  '[User] User config changed'
```

AlexGuironnetRTE (changes requested):
I would suggest a few minor changes (see comments).
Other than that, I've been able to test that the signalling works as intended, in so far that when I played with groups, entities or perimeters I saw the `CardService - USER_CONFIG_CHANGE received` log only when the user I was logged in as was affected. But I couldn't find any example to show that it had been taken into account in the UI.

For example:
- Starting a brand new instance, I log in with operator1. The user card feature is not available since it doesn't have any perimeters or bundle yet. I run the loadConf script, then I do see the update log in the console, but the user card feature remains unavailable (even after 10s)
- I have to reload the page manually for the feature to become available. Then, if I remove the perimeters from this user, again I do see the message in the logs but it doesn't change anything, I can still create a card on a perimeter that I don't have access to anymore, so I get the "cannot send card" message when I try to send it. 
But maybe that's not part of this ticket?

Also, I couldn't find what makes the user only receive the updates that are meant for him: is it in rabbitMQ or in the code?

Thanks!

services/core/cards-consultation/src/test/java/org/opfab/cards/consultation/application/configuration/ThreadPoolTaskSchedulerConfiguration.java
quinarygio: Done

quinarygio:
> I would suggest a few minor changes (see comments).
> Other than that, I've been able to test that the signalling works as intended, in so far that when I played with groups, entities or perimeters I saw the `CardService - USER_CONFIG_CHANGE received` log only when the user I was logged in as was affected. But I couldn't find any example to show that it had been taken into account in the UI.
> 
> For example:
> 
> * Starting a brand new instance, I log in with operator1. The user card feature is not available since it doesn't have any perimeters or bundle yet. I run the loadConf script, then I do see the update log in the console, but the user card feature remains unavailable (even after 10s)
> * I have to reload the page manually for the feature to become available. Then, if I remove the perimeters from this user, again I do see the message in the logs but it doesn't change anything, I can still create a card on a perimeter that I don't have access to anymore, so I get the "cannot send card" message when I try to send it.
>   But maybe that's not part of this ticket?

You are right it should work, and it was working before a made a small refactoring...

> 
> Also, I couldn't find what makes the user only receive the updates that are meant for him: is it in rabbitMQ or in the code?
> 
> Thanks!

The check is in CardSubscription : 
```

            if (this.userLogin.equals(modifiedUserLogin)) 
                publishDataIntoSubscription("USER_CONFIG_CHANGE");
```

ui/main/src/app/services/card.service.ts
quinarygio: This warning appear when sonar make the analyse on the merge in develop so it was not due to this PR but to another PR that was merged before.

AlexGuironnetRTE (approved):
Tests all OK.

PR #1197, opened 2021-06-09
[OC-1775] Reminder isn't triggered when "Minutes to remind before task" is 0

freddidierRTE:
Release note : 

BUG : 

* OC-1775 Reminder isn't triggered when "Minutes to remind before task" is 0


AlexGuironnetRTE (approved):
Small question (see comment) other than that, tests are OK.

PR #1203, opened 2021-06-10
[OC-1746] Upgrade spring libraries to last version

freddidierRTE:
Release note :+1: 

Task : 

[OC-1746] Migrate to springboot v2.5.0

[OC-1746]: https://opfab.atlassian.net/browse/OC-1746

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
     id "maven-publish"
     id "signing"
     id "org.owasp.dependencycheck" version "6.2.1"
-    id 'org.springframework.boot' version '2.4.6'
+    id 'org.springframework.boot' version '2.5.1'
     id 'org.sonarqube' version '3.2.0' apply false
 }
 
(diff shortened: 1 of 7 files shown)


AlexGuironnetRTE:
We also need to update the documentation here (and make sure that we point to a section of the Spring doc that still exists): src/docs/asciidoc/deployment/configuration/configuration.adoc

AlexGuironnetRTE (changes requested):
We also need to update the documentation here (and make sure that we point to a section of the Spring doc that still exists): src/docs/asciidoc/deployment/configuration/configuration.adoc

Also, the Spring Boot version needs to be updated for the Spring Boot plugin in as well in build.gradle (and ideally we should get rid of the other legacy declarations in the classpath, but I think I logged an issue for that).

AlexGuironnetRTE (approved):
Tests OK (dev and docker mode)

PR #1365, opened 2021-07-01
Fix #1311 Moving Karate and Cypress to separate stages

### LINKED ISSUE #1311: Travis - move Cypress and Karate to distinct stages
Create distinct stages or jobs for cypress and karate tests so we can see at a glance what’s wrong in a failing build, and maybe restart just the job that caused a problem.

Use workspaces so docker images can be reused for Cypress and Karate tests (https://docs.travis-ci.com/user/using-workspaces)

This will also avoid building docker images twice (once in test-sonar and once in docker stage).

If we choose to run them as separate jobs in a common “Integration tests” stage, they could run in parallel => see if it takes less time than sequentially or not.

AlexGuironnetRTE:

TODO Update pipeline doc (hooks etc.)
Docker push should use workspace as well
ui build runs again (rather than being taken from cache) when generating the docker images... 2m40 out of 5m25 (logged #1413)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ install:
   - node --version
   - sdk version
   - javac -version
+  - docker info
 jobs:
   include:
     - stage: test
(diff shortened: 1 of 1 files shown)


AlexGuironnetRTE:
Closing until we have a concurrent plan on Travis

PR #1406, opened 2021-07-07
#1256 Clarify sound when switching from feed to monitoring

### LINKED ISSUE #1256: Clarify sound when switching from feed to monitoring
Actual behavior :

-->  configure sound for alarm 
--> do not show alarm on feed (via severty filter)
--> send an alarm card 

Result : no sound

--> go in monitoring screen 
--> send an alarm card 

Result : It rings 

--> go in archives 
--> send an alarm card 
Result : It rings 

--> go in feed
--> go in archives 
Result : no sound 

WANTED behavior : 

The sound shall be activated depending **only on feed filters** 

The problem is that monitoring filter use the same filters as the feed (via ngrx), so we need to make an other mechanism of filtering for monitoring filters . Try to avoid ngrx for filtering ?

quinarygio:
Fix #1256 

Release notes

In Bugs section:
#1256 : Clarify sound when switching from feed to monitoring

### DIFF
--- a/ui/main/src/app/services/lightcards.service.ts
+++ b/ui/main/src/app/services/lightcards.service.ts
@@ -84,7 +84,7 @@ export class LightCardsService {
  // even if the flow is still continuing.
  // To do that we combine a debounce waiting for the end of the flow and an interval to get the card list every second 
 
-    private getLightCards(): Observable<any> {
+    public getLightCards(): Observable<any> {
         return merge(this.getLightCardsInterval(), this.getLightCardDebounce());
     }
 
@@ -112,7 +112,7 @@ export class LightCardsService {
 
 // --------------------
 
-    private filterLightCards(lightCards: LightCard[], filters: Filter[]): LightCard[] {
+    public filterLightCards(lightCards: LightCard[], filters: Filter[]): LightCard[] {
         if (filters && filters.length > 0) {
             return lightCards.filter(card => Filter.chainFilter(card, filters));
         }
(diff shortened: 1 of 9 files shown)


ui/main/src/app/services/lightcards.service.ts
```diff
@@ -77,6 +83,33 @@ export class LightCardsService {
...
+            this.getFoundLightCards()
+        ]
+        ).pipe(
+            map(results => {
+                return results[1]
+                    .sort(this.compareFn(results[0].sortBySeverity, results[0].sortByRead));
+            }
+            )
+        ).subscribe((lightCards) => this.foundAndSortedLightCards.next(lightCards));
+    }
```
freddidierRTE, line 98: We do not need to sort the card for the monitoring filter ?

freddidierRTE:
The difference between Filter and SearchFilter is not clear , we should find another way to distinguish thinks 

This solution is always filtering in background even if do not have the monitoring screen available or used , when having a lot of cards it may created unnecessary overload

Solution proposed : create these filters in the monitoring component and name the effects in ngrx with monitoring in the names

freddidierRTE (changes requested):

AlexGuironnetRTE (approved):
Code refactoring OK
Manual tests OK:
- No changes to sound notification behaviour for feed screen
- For monitoring screen, sounds are played based on feed filters only, the Service, Process, and Process Status filters present on the Monitoring screen are not taken into account. One notable exception is the business period, which is the one selected for the monitoring screen (using the timeline buttons) and which is not synced with the one for the feed. This is not in the scope of this issue, but I'm not sure if it's an optimal user experience so I created #1441 to discuss it.

Issue #1411, opened 2021-07-08
Advertise project adopters


AlexGuironnetRTE:
This is a requirement for the graduated stage of LF Energy projects.

- [x] Add an ADOPTERS.adoc file
- [x] Add adopters logo to website

AlexGuironnetRTE:
For the website part I pushed the changes directly so it got integrated in the website directly, I figured it would be easier to review this way but if you have any remarks or ideas I can make changes.
The adopters pages is accessible through the `Community` menu or with this link https://opfab.github.io/pages/adopters.html

I'm wondering whether Adopters shouldn't be its own menu item in the navbar actually, what do you think?

PR #1442, opened 2021-07-13
Initializing ADOPTERS file (#1411)

AlexGuironnetRTE:
No need to mention it in release notes.


AlexGuironnetRTE:
I cancelled the build because I only added one text file (which isn't even published to the website) so there's really nothing to build.

PR #1449, opened 2021-07-15
#1270 Synchronize JWT data with database

### LINKED ISSUE #1270: Synchro JWT data with database
Create a new endpoint users/synchronizeWithToken 

-  If user does not exist create it in database (this mechanism replace the actual mechanism where the front create the user via a PUT request) 

-  If entities is configured to be in token : 
   Update user entities  membership in opfab database if necessary 
   If an entity does not exist just log a warning in user service 

-  If groups is configured to be in token : 
    Update user groups membership  in opfab database if necessary 
    If a group does not exist just log a warning in user service 

The endpoint shall be call just after login and the front shall wait the result to continue loading.

quinarygio:
Fix #1270 

Release notes

In Tasks section:
#1270 Synchronize JWT token data with database

### DIFF
--- a/services/core/users/src/main/java/org/opfab/users/controllers/UsersController.java
+++ b/services/core/users/src/main/java/org/opfab/users/controllers/UsersController.java
@@ -12,9 +12,16 @@
 package org.opfab.users.controllers;
 
 import lombok.extern.slf4j.Slf4j;
+
+import org.opfab.springtools.configuration.oauth.jwt.JwtProperties;
+import org.opfab.springtools.configuration.oauth.jwt.groups.GroupsProperties;
+import org.opfab.springtools.configuration.oauth.jwt.groups.GroupsMode;
 import org.opfab.springtools.error.model.ApiError;
 import org.opfab.springtools.error.model.ApiErrorException;
+import org.opfab.users.configuration.oauth2.UserExtractor;
 import org.opfab.users.model.*;
+import org.opfab.users.repositories.EntityRepository;
+import org.opfab.users.repositories.GroupRepository;
 import org.opfab.users.repositories.UserRepository;
 import org.opfab.users.repositories.UserSettingsRepository;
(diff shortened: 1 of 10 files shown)


freddidierRTE:
- In UserController , the method : `public User updateUser(HttpServletRequest request, HttpServletResponse response, String login, User user) throws Exception {`  may be removed as it is only used by the front to create user

quinarygio:
> * In UserController , the method : `public User updateUser(HttpServletRequest request, HttpServletResponse response, String login, User user) throws Exception {`  may be removed as it is only used by the front to create user

    Maybe it would be better to keep separated methods for creating and updating a user, to be more consistent with REST verbs and to avoid the creation of wrong users when trying to update a user passing a wrong username by mistake.

services/core/users/src/main/java/org/opfab/users/controllers/UsersController.java
```diff
@@ -193,6 +204,62 @@ public Void deleteUser(HttpServletRequest request, HttpServletResponse response,
...
+        if (existingUser == null) {
+            log.debug(String.format(USER_CREATED, login));
+            userService.createUser(user);
+        } else {
+            boolean updatedFromToken = false;
+            if (groupsProperties.getMode() == GroupsMode.JWT) {
+                updatedFromToken = !existingUser.getGroupSet().equals(new HashSet<String>(user.getGroups()));
+            }
+            if (!updatedFromToken && jwtProperties.gettingEntitiesFromToken) {
+                updatedFromToken = !new HashSet<String>(existingUser.getEntities()).equals(new HashSet<String>(user.getGroups()));
```
freddidierRTE, line 251: it shall be :  
updatedFromToken = !new HashSet<String>(existingUser.getEntities()).equals(new HashSet<String>(user.**getEntities())**); ?

freddidierRTE (changes requested):

PR #1483, opened 2021-07-21
Correct responsive problem with cards examples containing charts (#1479)

freddidierRTE:
Release note 

Bug 

Correct responsive problem with cards examples containing charts (#1479)


AlexGuironnetRTE (approved):
Test OK

PR #1485, opened 2021-07-21
Make travis build fail fast if docker images can't be generated

### LINKED ISSUE #1260: Karate tests are run even if the actual build (back/front/docker) has failed
This build shows that even though the ui build failed, and thus the dockerTagXXXX task wasn't executed, the Karate tests are run:

Link to build: https://travis-ci.com/github/opfab/operatorfabric-core/builds/223924956

This is a waste of time as the build is not going to pass anyway, and it means that the karate tests are being executed on the latest SNAPSHOT version published to DockerHub rather than on the current code, which can be misleading.

Moving the Karate tests to a separate stage or adding a condition on the success of the building phase could solve this.

AlexGuironnetRTE:
Fixes #1260 

No need to mention it in release notes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ jobs:
       script:
         - docker login --username ${DOCKER_CLOUD_USER} --password ${DOCKER_CLOUD_PWD}
         - docker-compose -f src/main/docker/test-environment/docker-compose.yml up -d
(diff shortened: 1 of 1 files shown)


AlexGuironnetRTE:
Test OK, build does stop immediately in case of failure in the `./gradlew --build-cache copyDependencies test jacocoTestReport dockerTag${OF_VERSION}` command
See https://travis-ci.com/github/opfab/operatorfabric-core/builds/233357456

PR #1548, opened 2021-07-27
Fixing logout and adding test (#1529)

### LINKED ISSUE #1529: Revert logout behaviour
Harmonizing the configuration caused the logout not to work properly (just redirecting to the /ui page).

AlexGuironnetRTE:
Fixes #1529 

In bug section: Fixing logout and adding test (#1529)


AlexGuironnetRTE:
@freddidierRTE I assigned it to you because while working on develop (so without this "fix") on something else, I happened to use the logout and it seemed to work as intended (even though the logout-url parameter in the web-ui from the network screen was "/ui/").... By "as intended" I mean that it navigated to http://localhost:2002/#/feed, which was showing the login/password input boxes... If I tried visiting http://localhost:2002/ in another tab, same thing. The session storage was empty.
I was in PASSWORD mode. Am I missing something ?

Also I noticed that the local storage is not cleared on log out, so for example user2 logging after user1 would have the filters, reminders etc. from user1... I'm logging a bug.

AlexGuironnetRTE:
Oh. Just noticed that the token is still in the local storage...

freddidierRTE:
it works with /ui/ but it does not logout from Keyckloack SSO so it is better to use keycloack url to have a better example

PR #1694, opened 2021-09-07
Add getDisplayContext() method to the template to know in which context the rendering is done (#1647)

### LINKED ISSUE #1647: Add a method for template to know in which context the rendering is done
Add a method templateGateway.getDisplayContext()

Possible responses : 
- realtime
- archives
- preview

Add in documentation and see how to use it in current examples

quinarygio:
Fix #1647 

Release notes:

In Featues section:
#1647 Add getDisplayContext() method to the template to know in which context the rendering is done

### DIFF
--- a/ui/main/src/assets/js/templateGateway.js
+++ b/ui/main/src/assets/js/templateGateway.js
@@ -7,13 +7,13 @@
  * This file is part of the OperatorFabric project.
  */
 
-
 const templateGateway = {
     opfabEntityNames : null, 
     childCards: [],
     userAllowedToRespond : false,
     userMemberOfAnEntityRequiredToRespond : false,
     entitiesAllowedToRespond: [],
+    displayContext: '',
 
     setEntityNames: function(entityNames){
         this.opfabEntityNames = entityNames;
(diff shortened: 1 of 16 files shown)


ui/main/src/assets/js/templateGateway.js
```diff
@@ -7,13 +7,15 @@
  * This file is part of the OperatorFabric project.
  */
 
+const ContextEnum = {REALTIME: "REALTIME", ARCHIVE: 'ARCHIVE', PREVIEW: 'PREVIEW', UNKNOWN: 'UNKNOWN2'}
```
freddidierRTE, line 10: remove UNKNOWN value

freddidierRTE:
I would prefer not exposing the enum outside of operator fabric . So i propose :  

- Declare the enum in an internal class and in the template we use the string value (as it is done for setScreenSize()) 
- The internal class for the enum shall be in model directory and name it templateGateway.model.ts 
- Values shall be in lowercase (to be coherent with setScreenSize()) 
- Enum name shall be DisplayContext

PR #1729, opened 2021-09-14
Optimize light cards loading when browser is slow

freddidierRTE:
In release note

Task :

Optimize light cards loading when browser is slow


AlexGuironnetRTE (approved):
All good when reading the code. Didn't manage to reproduce slow browser though.

Issue #1737, opened 2021-09-15
Add spinner when loading very heavy card


OlivierVoron:
In some of our use cases, as cards are generated automatically by a business process, some cards could be very heavy to load in the card feed/archives or monitoring taking more than 3 seconds to load.

It will be useful to have a spinner that shows to the user that the card is loading/rendering after a few seconds.

To cancel the loading of this card, the user may click and load another card or have a specific button for it ?

PR #1752, opened 2021-09-17
Add spinner when card loading take more than one second (#1737)

freddidierRTE:
Release note :

Feature : 

#1737 : Add spinner when card loading take more than one second

### DIFF
--- a/ui/main/src/app/modules/cards/components/card-details/card-details.component.html
+++ b/ui/main/src/app/modules/cards/components/card-details/card-details.component.html
@@ -1,0 +1,26 @@
+<!-- Copyright (c) 2021, RTE (http://www.rte-france.com)              -->
+<!-- See AUTHORS.txt                                                       -->
+<!-- This Source Code Form is subject to the terms of the Mozilla Public   -->
+<!-- License, v. 2.0. If a copy of the MPL was not distributed with this   -->
+<!-- file, You can obtain one at http://mozilla.org/MPL/2.0/.              -->
+<!-- SPDX-License-Identifier: MPL-2.0                                      -->
+<!-- This file is part of the OperatorFabric project.                      -->
+
+<div *ngIf="card && cardState" style="font-size:13px;" [hidden] = "cardLoadingInProgress" >
+  <of-detail   [cardState]="cardState" [card]="card" [childCards]="childCards"
+             [currentPath]="_currentPath" [parentModalRef]="parentModalRef" [screenSize]="screenSize" [displayContext]="displayContext">
+  </of-detail>
+</div>
+
+<div *ngIf="cardLoadingInProgress" [class.opfab-card-loading]="!isSmallscreen()"  [class.opfab-card-loading-small-screen]="isSmallscreen()">
+
+  <div class="opfab-card-loading-spinner">
+  <em  class="fas fa-spinner fa-spin opfab-slow-spinner" ></em>
+  </div>
+  <div style="text-align: center;font:16px;font-weight: bold;margin-top:10px" translate>
+      feed.cardLoadingInProgress
+  </div>
+
+  <br/>
+  <br/>
+</div>
--- a/ui/main/src/app/modules/cards/components/card-details/card-details.component.ts
+++ b/ui/main/src/app/modules/cards/components/card-details/card-details.component.ts
@@ -12,29 +12,23 @@ import {Card} from '@ofModel/card.model';
 import {Store} from '@ngrx/store';
 import {AppState} from '@ofStore/index';
 import * as cardSelectors from '@ofStore/selectors/card.selectors';
+import * as feedSelectors from '@ofStore/selectors/feed.selectors';
 import {ProcessesService} from '@ofServices/processes.service';
 import {Subject} from 'rxjs';
 import {takeUntil} from 'rxjs/operators';
 import {UserService} from '@ofServices/user.service';
 import {selectCurrentUrl} from '@ofStore/selectors/router.selectors';
 import {AppService} from '@ofServices/app.service';
-import {State as CardState, State} from '@ofModel/processes.model';
-import { NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
+import {State} from '@ofModel/processes.model';
+import {NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import {DisplayContext} from '@ofModel/templateGateway.model';
 
 @Component({
     selector: 'of-card-details',
-    template: `
-
-            <div *ngIf="card && cardState" style="font-size:13px;">
-                <of-detail   [cardState]="cardState" [card]="card" [childCards]="childCards"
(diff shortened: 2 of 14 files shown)


ui/main/src/app/modules/cards/components/card-details/card-details.component.html
```diff
@@ -0,0 +1,26 @@
...
+<!-- SPDX-License-Identifier: MPL-2.0                                      -->
+<!-- This file is part of the OperatorFabric project.                      -->
+
+<div *ngIf="card && cardState" style="font-size:13px;" [hidden] = "cardLoadingInProgress" >
+  <of-detail   [cardState]="cardState" [card]="card" [childCards]="childCards"
+             [currentPath]="_currentPath" [parentModalRef]="parentModalRef" [screenSize]="screenSize" [displayContext]="displayContext">
+  </of-detail>
+</div>
+
+<div *ngIf="cardLoadingInProgress" [class.opfab-card-loading]="!isSmallscreen()"  [class.opfab-card-loading-small-screen]="isSmallscreen()">
```
AlexGuironnetRTE, line 15: Why did you choose to use use the `hidden` attribute for the first div and an ngIf for the second div?
I'm asking as I think ngIf and hidden are not the same thing for cypress for example.

ui/main/src/assets/i18n/en.json
```diff
@@ -113,7 +113,8 @@
     "at": "at",
     "from" : "from",
     "moreCards": "cards more",
-    "lastResponse": "Last response received on"
+    "lastResponse": "Last response received on",
+    "cardLoadingInProgress":"Loading card in progress ..."
```
AlexGuironnetRTE, line 117:
```suggestion
    "cardLoadingInProgress":"Loading card ..."
```

I think the `ing` already conveys that the action is currently being done. Alternatively, if we want to use `in progress` then what precedes has to be some kind of noun, so `Card loading in progress`.

ui/main/src/app/modules/cards/components/card-details/card-details.component.ts
```diff
@@ -96,8 +94,30 @@ export class CardDetailsComponent implements OnInit, OnDestroy {
...
             });
+        this.checkForCardLoadingInProgressForMoreThanOneSecond();
+
+    }
+
+    // we show a spinner on screen if card loading take more than 1 second
+    checkForCardLoadingInProgressForMoreThanOneSecond() {
+        this.store.select(feedSelectors.selectLightCardSelection) 
+            .pipe(takeUntil(this.unsubscribe$))
+            .subscribe((cardId) => {   // a new card has been selected and will be download
```
AlexGuironnetRTE, line 105:
```suggestion
            .subscribe((cardId) => {   // a new card has been selected and will be downloaded 
```

ui/main/src/app/modules/cards/components/card-details/card-details.component.ts
```diff
@@ -96,8 +94,30 @@ export class CardDetailsComponent implements OnInit, OnDestroy {
...
+    }
+
+    // we show a spinner on screen if card loading take more than 1 second
+    checkForCardLoadingInProgressForMoreThanOneSecond() {
+        this.store.select(feedSelectors.selectLightCardSelection) 
+            .pipe(takeUntil(this.unsubscribe$))
+            .subscribe((cardId) => {   // a new card has been selected and will be download 
+                this.currentSelectedCardId = cardId;
+                setTimeout(() => {
+                    if (this.currentSelectedCardId == cardId) // the selected card has not change in between
```
AlexGuironnetRTE, line 108:
```suggestion
                    if (this.currentSelectedCardId == cardId) // the selected card has not changed in between 
```

AlexGuironnetRTE (changes requested):
Minor remarks on the code (see comments).
Tests ok with contingenciesBig + regular 2G as advised, except one thing:

If I click on the big card then rapidly (before it's finished loading) I change my mind because I'm not patient ;-) and click the plain message card I get the text "loading card in progress" (and interestingly, not the spinner), and after around a second, the card is displayed.
Whereas, if I open the message card first thing after opening the application, it loads instantly.

It might be confusing for users if they see the spinner and have to wait for what is clearly a "light" card (just 2 lines of text)? And really big cards could block others ?
Also it's weird that the text is showing in this case without the spinner.

What do you think?

ui/main/src/app/modules/cards/components/card-details/card-details.component.html
freddidierRTE: generally i use *ngIf by default and hidden when *ngIf does not match my needs 

On the first div , it is needed to keep the component instantiate between updates so i use hidden . Otherwise when we are watching a card in fullscreen mode and the card is updated , il goes back in smallscreen mode as the component is destroy and recreated . 

For the second div , it's my default choice 

*ngIf by default seems the choice recommended by angular : 
https://stackoverflow.com/questions/51317666/when-should-i-use-ngif-over-hidden-property-and-vice-versa/51317774

ui/main/src/assets/i18n/en.json
freddidierRTE: ok done

ui/main/src/app/modules/cards/components/card-details/card-details.component.ts
freddidierRTE: Done

freddidierRTE:
> If I click on the big card then rapidly (before it's finished loading) I change my mind because I'm not patient ;-) and click the plain message card I get the text "loading card in progress" (and interestingly, not the spinner), and after around a second, the card is displayed.
> Whereas, if I open the message card first thing after opening the application, it loads instantly.
> 
> It might be confusing for users if they see the spinner and have to wait for what is clearly a "light" card (just 2 lines of text)? And really big cards could block others ?
> Also it's weird that the text is showing in this case without the spinner.
> 
> What do you think?

I do not manage to reproduce the problem of "no spinner but text" , i had seen it once the first time i click on the card on dev mode  after i switch form git branch but never after I guess this is due to css caching in the browser .

For the fact that when you click on a big card and change rapidly for a small card you get a spinner : it is not a problem as the spinner will just be there for  1s. I try to change this behavior but i manage to have it work half time with really small card so i would prefer keeping it as is to avoid adding to much complexity in the mechanism

AlexGuironnetRTE:
> I do not manage to reproduce the problem of "no spinner but text" , i had seen it once the first time i click on the card on dev mode after i switch form git branch but never after I guess this is due to css caching in the browser .

OK that makes sense. I don't think I had it all the time either.

> For the fact that when you click on a big card and change rapidly for a small card you get a spinner : it is not a problem as the spinner will just be there for 1s. I try to change this behavior but i manage to have it work half time with really small card so i would prefer keeping it as is to avoid adding too much complexity in the mechanism

OK then, no I agree it's not worth making it more complex

PR #1765, opened 2021-09-20
Monitoring screen : apply the sort to the export file (#1743)

vlo-rte:
Release notes : 

In Features section : 

#1743 Monitoring screen : apply the sort to the export file


quinarygio:
Sorting is not applied to export when there is a specific monitoring configuration loaded.

quinarygio (approved):

Issue #1761, opened 2021-09-20
Archives screen : Add an option to activate / desactivate collapsable updates 


freddidierRTE:
![image](https://user-images.githubusercontent.com/58251627/133971031-aae48842-43f4-462f-ab1a-4ecc5ba32e90.png)

AlexGuironnetRTE:
This feature should not be present in the Logging screen.

PR #1767, opened 2021-09-20
Archives screen : Add an option to activate / desactivate collapsible…

vlo-rte:
Release notes : 

In Features section :

#1761 Archives screen : Add an option to activate / desactivate collapsible updates

### DIFF
--- a/src/test/cypress/cypress/integration/Archives.spec.js
+++ b/src/test/cypress/cypress/integration/Archives.spec.js
@@ -21,6 +21,8 @@ describe ('Archives screen tests',function () {
         cy.get('#opfab-archives-cards-list').find('.opfab-archives-table-line').should('have.length',6);
         // No plus icon is displayed
         cy.get('#opfab-archives-cards-list').find('.opfab-archives-icon-plus').should('not.exist');
+        // No minus icon is displayed
+        cy.get('#opfab-archives-cards-list').find('.opfab-archives-icon-minus').should('not.exist');
 
         // No card detail is displayed
         cy.get('of-card-detail').should('not.exist');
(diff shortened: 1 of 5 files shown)


src/test/cypress/cypress/integration/Archives.spec.js
```diff
@@ -111,5 +121,8 @@ describe ('Archives screen tests',function () {
         // We check there is a plus icon in place of the minus icon
         cy.get('#opfab-archives-cards-list').find('.opfab-archives-table-line').first().find('.opfab-archives-icon-plus').should('have.length', 1);
         cy.get('#opfab-archives-cards-list').find('.opfab-archives-table-line').first().find('.opfab-archives-icon-minus').should('not.exist');
+
+        // We click collapsible updates
+        cy.get('#opfab-archives-collapsible-updates').click({force: true});
     })
```
freddidierRTE, line 127: Add a  test without collapsible mode and check there is no + and the number of cards visible is correct

ui/main/src/app/modules/archives/archives.component.ts
```diff
@@ -115,6 +119,12 @@ export class ArchivesComponent implements OnDestroy, OnInit {
         ).subscribe(() => this.setDateFilterBounds());
     }
 
+    toggleCollapsibleUpdates() {
+        this.isCollapsibleUpdatesActivated = !this.isCollapsibleUpdatesActivated;
+        localStorage.setItem('opfab.isCollapsibleUpdatesActivated', String(this.isCollapsibleUpdatesActivated));
+        this.sendQuery(0);
```
freddidierRTE, line 125: Do not send the query until user click on search to be coherent with other input field

freddidierRTE (changes requested):
do not call  this.loadUpdatesByCardId(requestID) when the collapsible is desactivated to avoid unecessay processing

vlo-rte:
> do not call this.loadUpdatesByCardId(requestID) when the collapsible is desactivated to avoid unecessay processing

Done.

ui/main/src/app/modules/archives/archives.component.ts
vlo-rte: The checkbox is not part of the filters, I think it is more user-friendly, as the checkbox and the search button are a little bit apart, it can be "boring" to click the one then the other each time.
So I propose to not modify this point if you agree ?

src/test/cypress/cypress/integration/Archives.spec.js
vlo-rte: Done.

ui/main/src/app/modules/archives/archives.component.ts
freddidierRTE: OK for me

PR #1792, opened 2021-09-23
Load and cache all processes at startup (#1758)

### LINKED ISSUE #1758: Too many unnecessary requests on startup
When there is a lot of card with same process , the browser is making  a lot of time the same request to get process definition 

See how to avoid it . 

image

quinarygio:
Fix #1758 

Release notes

In Tasks section:
#1758 Load and cache all processes at startup (#1758)

### DIFF
--- a/ui/main/src/app/services/processes.service.ts
+++ b/ui/main/src/app/services/processes.service.ts
@@ -76,7 +76,8 @@ export class ProcessesService {
                         if (this.processes.length === 0) {
                             console.log(new Date().toISOString(), 'WARNING : no processes configured');
                             this.translationsLoaded.next(null);
-                            } else {
+                        } else {
+                            this.loadAllProcessesInCache();
                             this.loadAllTranslations();
                             console.log(new Date().toISOString(), 'List of processes loaded');
                         }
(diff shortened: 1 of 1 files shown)


ui/main/src/app/services/processes.service.ts
```diff
@@ -109,6 +110,12 @@ export class ProcessesService {
                 );
     }
 
+    private loadAllProcceseData() {
+        this.processes.forEach(process => {
+            this.queryProcess(process.id, process.version).subscribe();
```
freddidierRTE, line 115: I think , you do not need to make queries to the back as the information is already present in the process object . 
So just doing : 
```
 this.processes.forEach(process => {
this.processCache.set(`${process.id}.${process.version}` ,process); 
...
```
would do the job ? 

In this case , the method shall be rename `loadAllProcessesInCache()"

freddidierRTE (changes requested):

ui/main/src/app/services/processes.service.ts
quinarygio: Done

Issue #1813, opened 2021-09-27
OutOfMemory or mongoDB exception using archive screen 


freddidierRTE:
When using collapse mode , if there is big cards in archives the server can crash with outOfMemory Exception or the request can throw a HTTP 500 error code due to mongDB memory limits.

Issue #1332, opened 2021-06-30
Refactor shared objects management


freddidierRTE:
Domain objects that need to be used by other services or outside code are made available under client, as  plain classes.

For example: client/actions/build/classes/java/main/org/lfenergy/operatorfabric/actions/model/Action.class

They are also generated as interfaces in the service that “owns” them, for example:

services/core/actions/build/classes/java/main/org/lfenergy/operatorfabric/actions/model/Action.class

This caused an issue (OC-727) in the action service because it received an Action object from a REST call to the Thirds service and didn’t know what to make of it since as far as it was concerned Action was an abstract class.

In addition, it means that the services need to create concrete classes implementing these interfaces (extra code, extra maintenance): for example ActionData implementing Action.

That’s also a problem when these classes are duplicated accross services with different implementations (even though it’s limited by the common interface): 

For example: 

org/lfenergy/operatorfabric/actions/model/ActionData.java

org/lfenergy/operatorfabric/thirds/model/ActionData.java

→ One has constructors and not the other.

In addition to maintenance issues, this adds complexity for new developers on the project.

At the moment, no one in the team has been able to tell why this choice was made.

 

Proposed course of action

Try and refactor the code so each object is only defined/generated once, as a concrete class, under client, and have the services depend on the client packages.

If we manage to go through with it, all the better, if not and we find there really is a good reason/blocking point that means we need to keep using interfaces, we should at least write down why in the documentation and call the interfaces and the client objects differently.

freddidierRTE:
too old issue

PR #1814, opened 2021-09-27
Optimization to avoid OutOfMemory or mongoDB exception using archive screen (#1813) 

freddidierRTE:
In release note : 

Bug : 

#1813 Optimization to avoid OutOfMemory or mongoDB exception using archive screen

### DIFF
--- a/services/core/cards-publication/src/main/modeling/swagger.yaml
+++ b/services/core/cards-publication/src/main/modeling/swagger.yaml
@@ -681,9 +681,9 @@ paths:
       description: >-
         get archived cards matching the criteria given as parameters. Results are limited to the cards that
         the calling user is allowed to see (based on the card recipients). For performance reasons, the response 
-        does not contains all lightCards field, the returns value are : id, uid,publisher, processVersion,process,
+        does not contain all lightCard fields, the returned fields are : id, uid,publisher, processVersion,process,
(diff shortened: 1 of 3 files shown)


services/core/cards-publication/src/main/modeling/swagger.yaml
```diff
@@ -680,7 +680,10 @@ paths:
       summary: get archived cards matching given criteria
       description: >-
         get archived cards matching the criteria given as parameters. Results are limited to the cards that
-        the calling user is allowed to see (based on the card recipients).
+        the calling user is allowed to see (based on the card recipients). For performance reasons, the response 
+        does not contains all lightCards field, the returns value are : id, uid,publisher, processVersion,process,
```
AlexGuironnetRTE, line 684:
```suggestion
        does not contains all lightCards fields, the returns value are : id, uid,publisher, processVersion,process,
```

AlexGuironnetRTE, line 684:
```suggestion
        does not contains all lightCards field, the returned values are : id, uid,publisher, processVersion,process,
```

AlexGuironnetRTE (approved):
Test ok, no crashes with 600 "big cards".
No regressions on Archives or Monitoring.
As discussed, a specific return type for this request (to avoid null fields) could be implemented as part of https://github.com/opfab/operatorfabric-core/issues/1332 as it's not a priority now.

PR #1815, opened 2021-09-28
Replay sound notifications until user interacts with app (#1808)

### LINKED ISSUE #1808: Current sound notifications are too short and can be missed
Sound notifications are currently emitted by the browser. They are emitted only once on the arrival of the associated card, and the sounds played are all rather short.
This might cause the operator to miss some of the sounds, if they are on the phone for example.

Proposed enhancement: Sounds for new cards will be re-played at regular intervals until the operator click something on the OperatorFabric interface (indicating that they have come back to their desk). 

To be discussed: 

- Should it apply for all severities?
- What interval? 5 minutes? Should it be configurable?

AlexGuironnetRTE:
Fixes #1808 

In release notes:

#1808: Replay sound notifications until user interacts with app
+
See documentation (//TODO Add link) for the relevant configuration properties.

### DIFF
--- a/config/docker/ui-config/web-ui.json
+++ b/config/docker/ui-config/web-ui.json
@@ -4,6 +4,8 @@
   "checkPerimeterForResponseCard": true,
   "checkPerimeterForSearchFields": false,
   "showUserEntitiesOnTopRightOfTheScreen": true,
+  "settings.replayInterval": 10,
+  "settings.replayEnabled": true,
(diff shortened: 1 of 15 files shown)


freddidierRTE:
This option of repeating sound shall be activate by user via the settings , by default the sound is not repetitive 

Why not a short value by default , like 5s for repetition ?

Issue #1790, opened 2021-09-23
Script for translation migration in DB


freddidierRTE:
Migration of existing data : we shall provide a script to fill in the value in titleTranslated and summaryTranslated based an title and summary field and config.json of the processes.

PR #1854, opened 2021-10-06
Script for translation migration in DB (#1790)

vlo-rte:


freddidierRTE:
Move the directory from src/migration-opfab3 to src/tooling/migration-opfab3

Issue #1830, opened 2021-09-30
To have a different list of Tags for the Archives and for the Logging


SaschaEschmann:
For instance for the OPCSTA project, we have : 

- 51 Tags in the "Archives" which are all the Tags
- Only 8 are necessary in the "Logging" screen but the 51 are available for the user (could be confusing)

PR #1870, opened 2021-10-11
To have a different list of Tags for the Archives and for the Logging…

vlo-rte:
#1830 To have a different list of Tags for the Archives and for the Logging


freddidierRTE:
test ok 

Just add a migration document in src/docs/asciidoc/resources/migration_guide_to_2.11.adoc

to precise that there is now a distinct configuration between archive and logging 

Reference it in src/docs/asciidoc/resources/index.adoc

PR #1992, opened 2021-10-27
Authorize sending response from entity emitting user card (#1873)

### LINKED ISSUE #1873: Authorize sending response from entity emitting user card
Linked to #1569, it will be useful if the emitter could also answer to their questions, sent to other parties. It will allow sharing comment and eventual change of mind regarding the response to address, and better interactivity between the emitter and the other parties.

quinarygio:
Fix #1873 

Release notes
In Features section:
#1873 : Authorize sending response from entity emitting user card. Use 'response.state.emittingEntityAllowedToRespond' field in bundle config.json to enable it.

### DIFF
--- a/services/core/businessconfig/src/main/java/org/opfab/businessconfig/model/ResponseData.java
+++ b/services/core/businessconfig/src/main/java/org/opfab/businessconfig/model/ResponseData.java
@@ -37,6 +37,8 @@ public class ResponseData implements Response {
     private Boolean lock;
     private String state;
     private List<String> externalRecipients;
+    @Builder.Default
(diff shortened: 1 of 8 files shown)


vlo-rte (changes requested):
Can you update the documentation with this new feature please ? I think we can add it here : https://opfab.github.io/documentation/current/reference_doc/#_configure_the_response_in_config_json

Can you add a karate test with a bundle which have the value set to true and just check the value is taken into account ? This way, we will have the two cases : false (like you did) and true.

vlo-rte (approved):

Issue #1991, opened 2021-10-27
Add a spinner when loading card list on logging


freddidierRTE:
If the loading of the card list take more than one second , show a spinner on the screen

Issue #1990, opened 2021-10-27
Add spinner when loading card on archives


freddidierRTE:
If the loading of the card list take more than one second , show a spinner on the screen

PR #2019, opened 2021-11-04
Add a spinner when loading logging cards (#1991)

olivierPigeon-RTE:
In Release notes, Tasks section :
"#1990 : Add a spinner when loading logging cards"

### DIFF
--- a/src/test/cypress/cypress/integration/StateTypeTests.spec.js
+++ b/src/test/cypress/cypress/integration/StateTypeTests.spec.js
@@ -67,6 +67,8 @@ describe('State type tests', function () {
 
             cy.loginOpFab('operator1', 'test');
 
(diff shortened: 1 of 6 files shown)


olivierPigeon-RTE:
Duplicated lines are not new to this pr and were already existing, so this commit did not add duplicated code

vlo-rte (changes requested):
When there are already cards displayed, we can see this : 
image

I suppose this is not the wanted behaviour ?

vlo-rte (changes requested):
The problem is not solved because when there are already displayed results, the spinner doesn't appear anymore although it takes more than 1 second. Can you correct it please ?

vlo-rte (approved):

Issue #2043, opened 2021-11-09
When no process group define , double select field for process (in archive and logging screens)


freddidierRTE:
![image](https://user-images.githubusercontent.com/58251627/140876956-0fbc9ee2-7143-4b8f-b21a-1a6c77947b75.png)

PR #2044, opened 2021-11-09
Correct bug when no process group define (#2043)

freddidierRTE:
In release note :

BUGS: 

When no process group define , double select field for process (in archive and logging screens) #2043


AlexGuironnetRTE (approved):
Tests OK

PR #2055, opened 2021-11-09
Add cypress test for usercard (#1891)

### LINKED ISSUE #1891: Add cypress test for usercard
Use operator1
Open usercard 
Send a message card to Control room 1 and control room 2  
Check preview
Check connected user see the card in feed
Check operator2 receive the card
Update card via operator1
Check update is received by operator2
Delete card via operator 1 
Check card is not visible anymore for operator1 and operator 2

quinarygio:
Fix #1891

Release notes
In Tasks section:
#1891 : Add cypress test for usercard 

### DIFF
--- a/config/docker/nginx-cors-permissive.conf
+++ b/config/docker/nginx-cors-permissive.conf
@@ -160,7 +160,7 @@ server {
     proxy_pass http://cards-consultation:8080/;
     proxy_set_header X-Forwarded-For $remote_addr;
   }
-  location /cardspub/cards/userCard {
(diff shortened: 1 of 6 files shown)


AlexGuironnetRTE:
The "detached from dom" error appeared once on the CI on a particularly slow build (I think there were several in parallel), but I couldn't reproduce it afterwards.
So I'm merging this as is (just removed the `waitDefaultTimes` in the test) and if it comes back we can fix it by adding a check that it contains "Hello" on the command so it is retried until it is stable, then type.

Issue #1532, opened 2021-07-26
Generate Avro task gives warning about ignored property


AlexGuironnetRTE:
The build gives the following warning: 

> > Task :client:cards-client-data:generateAvroJava
> Ignored the org.opfab.avro.Card.publishDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.lttd.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.startDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.endDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Timespan.start.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Timespan.end.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.publishDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.lttd.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.startDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
> Ignored the org.opfab.avro.Card.endDate.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.

Is that a problem?

JeroenGommans:
It is not nice. I fixed it and also made sure you can now send a startDate with a Java Instant as parameter (previously you needed to convert to a Long).

PR #2061, opened 2021-11-11
Updated AVRO files to better reflect CardPublicationData.

JeroenGommans:
Updated AVRO files to better reflect CardPublicationData. 
Fixed AVRO compiler warnings  (fixes Issue #1532)
Updated to avro plugin v 1.2.1

Signed-off-by: Jeroen Gommans <jeroen@anjersoftwaredesign.nl>


AlexGuironnetRTE (approved):
Thanks a lot ! And you're right about dates, using Instants will be more coherent with what the sample client app does.

Thanks for the PR!

Issue #2090, opened 2021-11-18
The LTTD value does not change when a User Card is modified


anouarguesrami:
If the LTTD is set in the bundle via  "templateGateway.getSpecificCardInformation()"  and "lttdVisible" is false, when the User Card is modified the LTTD value doesn't reset. Eg.:
LTTD = 30 min
After 10 min the card is modified
LTTD = 20 min instead of 30 min

PR #2099, opened 2021-11-19
Correct bug  #2090 : The LTTD value does not change when a User Card is modified

freddidierRTE:
In release note 

Bugs : 

#2090 The LTTD value does not change when a User Card is modified


AlexGuironnetRTE (approved):
Test ok

PR #2112, opened 2021-11-23
Fixing detached DOM error (#1891)

AlexGuironnetRTE:
Nothing in release notes

Adding the "distinctUntilChanged" helped as the template was no longer re-rendered when the state value didn't actually change, but the test was still flaky, failing on the CI with a detached DOM error.

After investigating for quite a while, I wasn't able to pinpoint where the unwanted re-creations of the component were coming from, but as @freddidierRTE suggested adding a debounceTime step between state changes and template loading seems to fix the problem consistently (ran it several times on the CI, no failure).
So we're keeping this solution, as in any case it's better not to fire template requests until things have settled down.

### DIFF
--- a/src/test/cypress/cypress/integration/UserCard.spec.js
+++ b/src/test/cypress/cypress/integration/UserCard.spec.js
@@ -63,7 +63,7 @@ describe('User Card ', function () {
       })
 
 
-    /*describe('Edit user card', function () {
+    describe('Edit user card', function () {
 
       it('Edit User card from operator1', ()=>{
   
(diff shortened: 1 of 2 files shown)


AlexGuironnetRTE:
Adding the "distinctUntilChanged" helped as the template was no longer re-rendered when the state value didn't actually change, but the test was still flaky, failing on the CI with a detached DOM error.

After investigating for quite a while, I wasn't able to pinpoint where the unwanted re-creations of the component were coming from, but as @freddidierRTE suggested adding a debounceTime step between state changes and template loading seems to fix the problem consistently (ran it several times on the CI, no failure).
So we're keeping this solution, as in any case it's better not to fire template requests until things have settled down.

Issue #2015, opened 2021-11-04
Adding cypress tests for feature "Use user perimeter to present process/state list in UI"


vlo-rte:
Related to the issue "Use user perimeter to present process/state list in UI (#1908)"

PR #2113, opened 2021-11-23
Adding cypress tests for feature "Use user perimeter to present proce…

vlo-rte:
Release notes : 

In Tasks section : 

#2015 : Adding cypress tests for feature "Use user perimeter to present process/state list in UI"


freddidierRTE:
Duplicate the test with operator4 
Add a test with admin account , shall see message "No process/state available"
Add a test with operator 1 but no processGroup

vlo-rte:
> Duplicate the test with operator4 Add a test with admin account , shall see message "No process/state available" Add a test with operator 1 but no processGroup

Done.

Issue #2027, opened 2021-11-05
Admin UI multiselect are not sorted


quinarygio:
When creating or editing a user/entity/group/perimeter, multiselect dropdown lists are not sorted.
Example:
![image](https://user-images.githubusercontent.com/19426449/140534877-2ee308f6-98b6-4fa2-a68f-fe4f3d66dc98.png)

To be checked with users if it is ok to apply alphabeticl order or not.

PR #2120, opened 2021-11-24
Admin UI multiselect are not sorted (#2027)

vlo-rte:
For perimeters, the select dropdown list for the states is sorted but without using the prefix. To be discussed maybe ?

Release notes : 

In Tasks section : 

#2027 : Admin UI multiselect are not sorted


freddidierRTE:
the state are in the form stateID - stateName , i think we should change it for stateName(stateID)  so do not touch it , i will create a new issue

Issue #2028, opened 2021-11-05
Admin screen : pagination and displayed results not coherent after a delete


vlo-rte:
In the admin screens, if you delete a group or a perimeter (and maybe entity/user ?) which is displayed on the second page, then you see the results of the first page whereas the pagination indicates you are still on the second page.

freddidierRTE:
I observed the same behaviour when you just modify a user

PR #2159, opened 2021-11-30
Admin screen : pagination and displayed results not coherent after a …

vlo-rte:
Release notes : 

In Bugs section : 

#2028 : Admin screen : pagination and displayed results not coherent after a delete

### DIFF
--- a/ui/main/src/app/modules/admin/components/table/admin-table.directive.html
+++ b/ui/main/src/app/modules/admin/components/table/admin-table.directive.html
@@ -14,7 +14,7 @@
     <ag-grid-angular
             (gridReady)="onGridReady($event)"
             [gridOptions]="gridOptions"
-            [rowData]="rowData | async"
+            [rowData]="rowData"
             class="opfab-ag-grid-theme"
     >
     </ag-grid-angular>
--- a/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
+++ b/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
@@ -106,7 +106,7 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
   // ag-grid configuration objects
   public gridOptions;
   public gridApi;
-  public rowData: any;
+  public rowData: any[];
 
   public page: number = 1;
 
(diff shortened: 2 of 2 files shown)


AlexGuironnetRTE (changes requested):
The results are coherent after a delete, however it's not the case after a creation.

Steps to reproduce:
- Go to the Groups page for example, set the number of items per page to 5 and create/delete groups until you have exactly 5 groups
- Create a sixth group : as visible below, the pagination button shows that we are on page 2 but still displays page 1.

image

Also, this causes a new "GET" request to be sent to the back after each data change, rather than having the front update its local copy of the data by itself... But given that the number of changes through the Admin UI are limited, I don't think it's a problem.

AlexGuironnetRTE:
As discussed with @vlo-rte, our tests show that the behaviour can be different: 
- depending on whether there is still a "page 2" after the deletion
- depending on whether we do the creation and deletion immediately after, or if we refresh the page/tab between the two

AlexGuironnetRTE:
@vlo-rte You were right about the request thing, it's already the case on develop it wasn't introduced by your changes.

vlo-rte:
I've just made an update to correct the bug you mentioned.

AlexGuironnetRTE (changes requested):
This is really weird. This new version fixes the problem I described in my previous review, but only the first time!!

- I start with 5 users.
- I reload the app to make sure I start from a "fresh" cache
- In the users tab, I change the page size to 5.
- I add a user: results number is now 6, a button "2" appears in the pagination but we remain on page 1 and the table displays the lines for page 1 => all is as it should be, it fixes what I pointed out in the previous review.
- Then, I go to page 2 and delete the newly created user.
- After confirming, the page 2 disappears, we are back on page 1 and the table displays page 1 content. So far, so good.
- Then because I wanted to be sure, I create a user again. But this time, the behaviour is incorrect: the pagination says we're on page 2 but the lines displayed are page 1.

I thought I was going crazy so I recorded it on video if you want. I tried looking at the code a little but so far I have no idea why the second time would behave differently...

AlexGuironnetRTE:
> This is really weird. This new version fixes the problem I described in my previous review, but only the first time!!
> 
>     * I start with 5 users.
> 
>     * I reload the app to make sure I start from a "fresh" cache
> 
>     * In the users tab, I change the page size to 5.
> 
>     * I add a user: results number is now 6, a button "2" appears in the pagination but we remain on page 1 and the table displays the lines for page 1 => all is as it should be, it fixes what I pointed out in the previous review.
> 
>     * Then, I go to page 2 and delete the newly created user.
> 
>     * After confirming, the page 2 disappears, we are back on page 1 and the table displays page 1 content. So far, so good.
> 
>     * Then because I wanted to be sure, I create a user again. But this time, the behaviour is incorrect: the pagination says we're on page 2 but the lines displayed are page 1.
> 
> 
> I thought I was going crazy so I recorded it on video if you want. I tried looking at the code a little but so far I have no idea why the second time would behave differently...

No one else in the team can reproduce this problem so we're considering it ok for now.

Issue #2136, opened 2021-11-26
Monitoring screen : Answers filtering based on id instead of name


freddidierRTE:
In monitoring, the column answser is filtered using the id of the entity and not the entity name 

![image](https://user-images.githubusercontent.com/58251627/143584571-45674061-6f8a-4974-95d0-4746afc9e6d0.png)

PR #2160, opened 2021-11-30
Monitoring screen : Answers filtering based on id instead of name (#2136)

vlo-rte:
Release notes : 

In Bugs section : 

#2136 : Monitoring screen : Answers filtering based on id instead of name


AlexGuironnetRTE (approved):
Tests OK, code OK

Issue #1889, opened 2021-10-13
Add cypress for monitoring screen 


freddidierRTE:
send 6 test cards 
check you have the right number of card 
check you can open a card 
check you can filter by processGroup / process / state 
check the reset button
check filter and sorting table

PR #2175, opened 2021-12-03
Add cypress tests on monitoring screen (#1889)

olivierPigeon-RTE:

### DIFF
--- a/src/test/cypress/cypress/integration/Monitoring.spec.js
+++ b/src/test/cypress/cypress/integration/Monitoring.spec.js
@@ -3,7 +3,9 @@ describe ('Monitoring screen tests',function () {
 
     before('Set up configuration', function () {
         cy.loadTestConf();
+        cy.send6TestCards();
     });
+
 
     it('Check composition of multi-filters for process groups/processes/type of state for operator1', function () {
         cy.loginOpFab('operator1', 'test');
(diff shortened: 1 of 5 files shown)


src/test/cypress/cypress/integration/Monitoring.spec.js
```diff
@@ -107,4 +108,289 @@ describe ('Monitoring screen tests',function () {
...
+        // Access monitoring screen
+        cy.get('#opfab-navbar-menu-monitoring').click();
+
+        // Should have the correct number of cards (5 because one card is set not to be visible in monitoring)
+        cy.get('#opfab-monitoring-table-grid').find('.ag-center-cols-container').find('.ag-row').should('have.length', 5);
+
+        // Filter on IN PROGRESS cards
+        cy.get('#opfab-typeOfState').click();
+        cy.get('#opfab-typeOfState').contains('IN PROGRESS').should('exist');
+        cy.get('#opfab-typeOfState').find('.lazyContainer > li').first().click();
```
freddidierRTE, line 155: Why not using a select by value , like cy.get(#opfab-typeOfState).select(xxx).

https://docs.cypress.io/api/commands/select#Value

src/test/cypress/cypress/integration/Monitoring.spec.js
```diff
@@ -107,4 +108,289 @@ describe ('Monitoring screen tests',function () {
...
     })
+
+
+
+    it('Check monitoring cards reception', function () {
+
+        cy.loginOpFab('operator1','test');
+        cy.send6TestCards();
+
+        // We move to archives screen
```
freddidierRTE, line 119: comment is // We move to **monitoring** screen

src/test/cypress/cypress/integration/Monitoring.spec.js
```diff
@@ -107,4 +108,289 @@ describe ('Monitoring screen tests',function () {
         cy.loadTestConf();
         cy.reload();
     })
+
+
+
+    it('Check monitoring cards reception', function () {
+
+        cy.loginOpFab('operator1','test');
+        cy.send6TestCards();
```
freddidierRTE, line 117: Send 6 cards only once instead of in each it() (in the before method) , it will save time

freddidierRTE (changes requested):

src/test/cypress/cypress/integration/Monitoring.spec.js
olivierPigeon-RTE: there is no "select" in the html so cy.select(...) will not work here

PR #2179, opened 2021-12-03
Permit multiple entity to edit the same card (#1646)

### LINKED ISSUE #1646: Permit multiple entity to edit the same card
Add field entitiesAllowedToEdit in card as an array of string representing entities id 

When user see a card in the feed , he can see the button to edit and can edit the card if : 
- he is member of an the publisher entity 
- or  he is member of an entity in entitiesAllowedToEdit (do not forget to take into account , groups of entities) 

Implements the check in the back.  

To delete the card , we keep the actual rules , only the user member of the publisher entity can delete the card 

When a user update a card , the publisher is set to his entity 

The entitiesAllowedToEdit value can be set by user card template via the getSpecificCardInformation method . By default the value is not set

quinarygio:
Fix #1646 

Release notes

In Features section:
 #1646 : Permit multiple entity to edit the same card


src/test/resources/bundles/userCardExamples2/template/usercard_message.handlebars
```diff
@@ -22,6 +22,7 @@
         summary : {key : "message.summary"},
 		title : {key : "message.title"},
         data : {message: message},
+        entitiesAllowedToEdit: ['ALLCONTROLROOMS'],
```
freddidierRTE, line 25: i would only touch the defaultProcessV1 process so remove this line and put ALLCONTROLROOMS in src/test/resources/bundles/defaultProcess_V1/template/usercard_message.handlebars

ui/main/src/app/services/user-permissions-.service.ts
```diff
@@ -42,23 +42,45 @@ export class UserPermissionsService {
...
+    if ((card.publisherType === 'ENTITY') && (user.userData.entities.includes(card.publisher) 
+      || (!!card.entitiesAllowedToEdit && this.isUserInEntityAllowedToEditCard(user, card)))) {
+        permission = this.checkUserWritePerimeter(user, card);
+    }
+    return permission;
+  }
+
+  private checkUserWritePerimeter(user: UserWithPerimeters, card: Card) : boolean {
+    let permission = false;
+    user.computedPerimeters.forEach(perim => {
```
freddidierRTE, line 65: rename file to ui/main/src/app/services/user-permissions.service.ts  (removing the - )

freddidierRTE:
Add some cypress test 

- Send the message card via script 
- Check card is not editable by a user member of an entity not allowed to edit 
- Edit the card  and update it with a user member of an entity allowed to edit 
- Create a new message 
- Edit the new card  and update it with a user member of an entity allowed to edit

freddidierRTE (changes requested):

freddidierRTE:
I do not see any use of src/test/resources/cards/cypress/userCard/message2.json in cypress test 

In cypress test : 

```
      it('Cannot edit card from not allowed entity', ()=>{
        cy.loginOpFab('operator2','test');
        cy.get('of-light-card').should('have.length',1);
        cy.get('of-light-card').eq(0).click()
          .find('[id^=opfab-feed-light-card]')
          .invoke('attr', 'data-urlId')
          .then((urlId) => {
              cy.hash().should('eq', '#/feed/cards/'+urlId);
              cy.get('#opfab-card-edit').should('exist');
        });
```
The #opfab-card-edit shall not exist ?

Issue #2181, opened 2021-12-06
Add an option for user to not fitler timeline 


freddidierRTE:
In the feed filter configuration , add on option to apply or not feed filters to  timeline : 

![image](https://user-images.githubusercontent.com/58251627/145004097-56451318-9080-4479-98f8-3a88f2dbd2c9.png)

PR #2182, opened 2021-12-06
Add an option for user to not fitler timeline (#2181)

freddidierRTE:
Remove ngrx for filtering and sorting lightcards in feed 

Use filter.service.ts to manage filter instead of ngrx
Use sort.service.ts to manage sorting instead of nrgx 
Store services related to lightcards in "services/lightcards" repository 

In release note : 

Under feature :

#2181 Add on option for user to not fitler timeline


AlexGuironnetRTE (approved):
Tests OK:
- Display of the button depending on web-ui property
- Test with all available filters
- Tests in realtime, week etc.

No remarks on code

Issue #2197, opened 2021-12-07
Remove unused tag feature


freddidierRTE:
### Discussed in https://github.com/opfab/operatorfabric-core/discussions/2187

<div type='discussions-op-text'>

<sup>Originally posted by **freddidierRTE** December  6, 2021</sup>
The idea is to remove part of tags features which is not used and using a component which is not compatible with last version of angular (13) 

**Remove tag search in feed**

![image](https://user-images.githubusercontent.com/58251627/144852914-fa8f6743-d1b1-41cf-88fb-f4b541a65512.png)

**Remove tag preference in settings screen** 

![image](https://user-images.githubusercontent.com/58251627/144853028-4ff70cbc-7120-4733-99ac-8add12e5a117.png)

**But **KEEP TAGS  list in archives and logging screen** for searching cards** 

![image](https://user-images.githubusercontent.com/58251627/144853127-9ce437a1-f5bf-492d-8ce5-df271dc86933.png)
</div>

PR #2204, opened 2021-12-08
Remove unused tag feature (#2197)

freddidierRTE:
Release note : 

Feature : 
- #2197 Remove unused tag feature


AlexGuironnetRTE (approved):
Tested on localhost:2002, both on firefox & chrome

- No error messages in console
- Basic tests OK (sending/receiving cards, navigation, etc.)
- Tested on top of existing Mongo database (with users who had tags in their settings) => the field stays present in the database but is not returned by the API
Maybe mention it in the migration guide ?
- No remaining unused translation assets

PR #2216, opened 2021-12-09
Remove "core" subdirectory and adapt build accordingly

### LINKED ISSUE #2035: Custom docker-compose Gradle tasks don't work
The services.gradle build script contains code to dynamically create gradle tasks for docker-compose operations (for example `gradle composeUp` for `docker-compose up`). They're meant to be run from each service.

However, they don't work currently as they rely on a docker-compose.yml file under `{project directory for the service}/src/main/docker/` which doesn't exist (anymore ?). They're not documented and I can't really see the point of starting a single service as a docker container as:
- It would mean duplicating configuration between config/docker and the service-specific docker-composes
- I would have to check but I'm pretty sure services launched in that way wouldn't communicate out of the box as they wouldn't share a common network (as is the case between containers from the same docker-compose file).

=> Does anyone see a use for this or should we remove it ?

As shown by the snippet below, there is one interesting thing though: in the case of up or start, they take care of re-creating the volume data from the version-controlled one to make sure we always start with the same bundles for example.

See services/services.gradle

```
        Map[] commands = [
                [command: ["up"], dependsOn: "copyWorkingDir"],
                [command: ["down"]],
                [command: ["logs", "-f"]],
                [command: ["start"], dependsOn: "copyWorkingDir"],
                [command: ["stop"]]
        ]
        for (Map c : commands) {
            createComposeTasks(c, project)
        }
```
and
```
void createComposeTasks(Map taskDef, Project prj){
    def cmdName = taskDef.command[0].capitalize()
    def baseargs = ["-f", "$prj.projectDir/src/main/docker/docker-compose.yml"]
    def bashargs = ((baseargs as List) << (taskDef.command as List)).flatten()
//    println "generating compose$cmdName task with $bashargs"
    def task = prj.task([type: Exec],"compose$cmdName", {
        executable "docker-compose"
        args bashargs
//      dependsOn prj.tasks.copyWorkingDir
    })

    if(taskDef.dependsOn){
        task.dependsOn prj.tasks[taskDef.dependsOn]
    }
}
```

### LINKED ISSUE #2196: Gradle build - Move all services directly under services and remove core

### LINKED ISSUE #2215: Remove unused swagger custom tasks
We have custom tasks that are meant to help with setting up / debugging Swagger generation:

- swaggerHelp : display help regarding swagger configuration options for java => this configuration documentation is available on the repo
- debugSwaggerOperations : produces some report on the swagger generation (couldn't get it to work on develop, it just hangs forever)

However, they're (currently) never used either locally or by the CI build, so I think it's not worth keeping them as our build is already complex enough.

AlexGuironnetRTE:
Fixes #2196 
Fixes #2035 
Fixes #2215 

I also took this opportunity of removing unused / obsolete tasks in the build (#2035, #2215).

I isolated the moving of files from services/core/XXX to services/XXX in a separate commit because otherwise it makes it difficult to look at the changes.

In release notes, under tasks:

* #2196: Remove "core" subdirectory and adapt build accordingly
* #2035, #2215: Remove obsolete Gradle tasks


AlexGuironnetRTE:
I checked, the front unit tests are still run this time ;-)

Issue #2209, opened 2021-12-08
Offer the possiblity to hide the recipient list in usercard 


freddidierRTE:
Add a parameter in config.json , recipientVisible 

Example : 

```
      "userCard" : {
				"template" : "usercard_incidentInProgress",
				"severityVisible" : true,
				"startDateVisible" : false,
				"endDateVisible" : false,
                                "lttdVisible" : false,
                                "recipientVisible" : false,
                                "recipientList" : [{"id": "ALLCONTROLROOMS", "levels": [0,1]}, {"id": "ENTITY4"}]
			},
```

Default value : true 

Business rule : 

If recipientVisible == true , nothing change 
if recipientVisible == false : 
 - if recipientList is empty , entityRecipient = [] 
 - else entityRecipient = recipientList

Use examples : 
- IT incident : recipientVisible = false /    "recipientList" : [{"id": "ALLCONTROLROOMS", "levels": [0,1]}, {"id": "ENTITY4"}]
- Task (src/test/resources/bundles/userCardExamples3/config.json) , "recipientVisible" = false

Doc : https://opfab.github.io/documentation/current/reference_doc/#_recipients

Cypress test : src/test/cypress/cypress/integration/UserCard.spec.js

Add in karate test : src/test/api/karate/businessconfig/getABusinessconfig.feature 
 bundle : src/test/api/karate/businessconfig/resources/bundle_api_test/config.json

PR #2247, opened 2021-12-15
Offer the possiblity to hide the recipient list in usercard (#2209)

olivierPigeon-RTE:

In release notes, features section :
#2209 : Offer the possiblity to hide the recipient list in usercard

### DIFF
--- a/services/businessconfig/src/main/modeling/swagger.yaml
+++ b/services/businessconfig/src/main/modeling/swagger.yaml
@@ -537,6 +537,8 @@ definitions:
         type: boolean
       lttdVisible: 
         type: boolean
+      recipientVisible: # tells whether to display recipient dropdown when creating a user card
+        type: boolean
       recipientList:
         type: array
         items:
--- a/src/test/resources/bundles/userCardExamples3/config.json
+++ b/src/test/resources/bundles/userCardExamples3/config.json
@@ -14,7 +14,8 @@
       "color": "#0070C0",
       "userCard" : {
         "template" : "usercard_task",
-        "lttdVisible" : false
+        "lttdVisible" : false,
+        "recipientVisible" : false
       },
       "templateName": "task",
       "styles": [
--- a/ui/main/src/app/modules/usercard/usercard.component.ts
+++ b/ui/main/src/app/modules/usercard/usercard.component.ts
@@ -104,10 +104,11 @@ export class UserCardComponent implements OnDestroy, OnInit {
     public displaySendingCardInProgress = false;
 
     modalRef: NgbModalRef;
-    severityVisible = true;
-    startDateVisible = true;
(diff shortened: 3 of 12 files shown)


src/test/resources/bundles/userCardExamples3/config.json
```diff
@@ -14,7 +14,8 @@
       "color": "#0070C0",
       "userCard" : {
         "template" : "usercard_task",
-        "lttdVisible" : false
+        "lttdVisible" : false,
+        "recipientVisible" : false
```
AlexGuironnetRTE, line 18: I understand why we need a test with an empty recipientList, so OK for modifying the corresponding template under API, but not on this one because otherwise it makes the task usercard a bit useless to play with as it's only received by the sender themselves.

ui/main/src/app/modules/usercard/usercard.component.ts
```diff
@@ -450,6 +451,7 @@ export class UserCardComponent implements OnDestroy, OnInit {
             this.startDateVisible = (userCard.startDateVisible === undefined) ? true : userCard.startDateVisible;
             this.endDateVisible = (userCard.endDateVisible === undefined) ? true : userCard.endDateVisible;
             this.lttdVisible = (userCard.lttdVisible === undefined) ? true : userCard.lttdVisible;
+            this.recipientVisible = (userCard.recipientVisible === undefined) ? true : userCard.recipientVisible;
```
AlexGuironnetRTE, line 454: It's not specific to `recipientVisible`, but it looks like we're defining the default to be `true` 3 times:
- Line 111 when declaring the field
- Here, it the case where it's a usercard
- Line 463 if it's not a usercard

services/businessconfig/src/main/modeling/swagger.yaml
```diff
@@ -537,6 +537,8 @@ definitions:
         type: boolean
       lttdVisible: 
         type: boolean
+      recipientVisible: 
+        type: boolean
```
AlexGuironnetRTE, line 541: It could be interesting to have a short description here like for other properties, so people wanting to send cards see at once what this property impacts without having to go look in the adoc documentation.
From the name alone it could mean displaying the recipients in the card rather than displaying the recipients dropdown.

AlexGuironnetRTE (changes requested):
Tests all OK, but a few requested changes in code and doc. Let me know if anything's unclear!

ui/main/src/app/modules/usercard/usercard.component.ts
olivierPigeon-RTE: Ok, I'll let the default value of all fields at this place then (lines 449-456), and remove it from other places

AlexGuironnetRTE (approved):

Issue #2257, opened 2021-12-16
Permit to access to log api only to members of admin group


freddidierRTE:
The end point actuator/loggers/** shall be accessible for each service only for members of group admin 

Add karate tests for each service as a new karate file in src/test/api/karate/admin

For each service : 
- check an access with a not admin user , it shall return an error code 
- send a request to set log to DEBUG level 
- send a request to get log and check it is DEBUG level
- send a request to set log to INFO level (to avoid having log level to DEBUG for following tests) 
- send a request to get log and check it is INFO level

example to get a log level : 

curl  http://localhost:2104/actuator/loggers/org.opfab -H "Authorization: Bearer $token" -H "Content-type:application/json"

example to set a log level to DEBUG : 

curl -i -X POST http://localhost:2104/actuator/loggers/org.opfab -H "Authorization: Bearer $token" -H 'Content-Type: application/json' -d '{"configuredLevel": "DEBUG"}'

PR #2311, opened 2021-12-22
Permit to access to log api only to members of admin group (#2257)

quinarygio:
Release notes
In Task section:
 #2257 : Permit to access to log api only to members of admin group


freddidierRTE:
Add the same config/test for newly created service external-devices

freddidierRTE (changes requested):

Issue #2270, opened 2021-12-17
Admin screen : Incoherence between pagination and displayed results after deleting a filter


vlo-rte:
For example, if you are on a second page, you put a filter, you delete your filter, first page of results is displayed whereas pagination indicates you are still on second page.

![image](https://user-images.githubusercontent.com/57365872/146536132-014f25d6-1866-4e4f-8c9d-d3a8c7a1e05b.png)

PR #2323, opened 2021-12-29
Admin screen : Incoherence between pagination and displayed results after deleting a filter (#2270)

vlo-rte:
Release notes : 

In Bugs section : 

#2270 : Admin screen : Incoherence between pagination and displayed results after deleting a filter

### DIFF
--- a/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
+++ b/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
@@ -216,6 +216,11 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
     return this.translateService.instant(this.i18NPrefix + headerIdentifier);
   }
 
+  onFilterChanged(event) {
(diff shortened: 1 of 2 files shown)


ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -216,6 +216,11 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
     return this.translateService.instant(this.i18NPrefix + headerIdentifier);
   }
 
+  onFilterChanged(event) {
+    this.page = 1;
+    this.gridApi.paginationGoToPage(this.page-1);
```
freddidierRTE, line 221: why not this.gridApi.paginationGoToPage(**0**) ?

vlo-rte: True. Done.

PR #2424, opened 2022-01-21
Automatically send response from entity emitting user card (#1569)

### LINKED ISSUE #1569: Automatically send response from entity emitting user card
Currently, if a user from entity A sends a user card expecting answers for which entity A (among others) is allowed/required to respond, someone then has to manually respond on behalf of entity A. There are some use cases where the response from the entity initiating the card could be sent along with the initial card.
For example, an entity providing several timeslots for a maintenance can be assumed to be OK with all of them at the time of sending (even if that can change later).

Expected behaviour:
- The user creates the user card. In the preview tab, they see the card as it would be if they had already answered it: no inputs boxes, and their answer is already visible in any content based on child cards (for example a table listing entities next to their response).
- After clicking OK, the user is taken back to the feed with the card details displayed. The button is "Modify answer" and their answer is taken into account in the template just as if they had actually answered. The entity is green in the card detail header.
- If need be, they can modifiy this answer just like a normal answer.
- Response shows up in archives / monitoring / logging ?

quinarygio:
Fix #1569 

Release notes:
In Features section:
#1569 : Add the possibility to automatically send response from entity emitting user card 


src/docs/asciidoc/reference_doc/user_cards.adoc
```diff
@@ -179,6 +179,40 @@ The recipient field can be hidden using the attribute `recipientVisible` : if hi
 Once a user card has been sent it can be edited by a user member of the publisher entity.
 It is possible to allow other entities to edit the card by specifing the 'entitiesAllowedToEdit' card field.
 
+== Send response automatically (experimental feature)
+
+It is possibiìle to configure a template to automatically send a response when sending a user card expecting an answers from one of the entities of the emitting user.
```
freddidierRTE, line 184: it is **possible**

src/test/cypress/cypress/integration/UserCard.spec.js
```diff
@@ -98,6 +98,52 @@ describe('User Card ', function () {
...
+      cy.get("of-card-detail").should('exist');
+
+      // Feed preview show response from current entity
+      cy.get('#opfab-feed-lightcard-hasChildCardFromCurrentUserEntity').should('exist');
+      // Card preview show responses list
+      cy.get("#childs-div").should('not.be.empty');
+      // Response table has 1 header and 1 row 
+      cy.get("#childs-div").find('tr').should('have.length', 2);
+      // Cancel sending
+      cy.get('#opfab-usercard-btn-refuse').click();
```
freddidierRTE, line 141: In this test or in a new one  , send the card and check it is received in the feed open it and see response from child card is present

freddidierRTE (changes requested):

PR #2432, opened 2022-01-24
Add endpoint to list all versions of a business process (#2401)

### LINKED ISSUE #2401: Only latest uploaded bundle is shown
In the `getProcesses` API for Businessconfig, it is unclear the response only returns the latest uploaded version per ID. I was expecting to find _all_ process, including previous versions.

image

https://github.com/opfab/operatorfabric-core/blob/41d1e0090d109326d116367b428988cfaf4d8a9f/services/businessconfig/src/main/java/org/opfab/businessconfig/services/ProcessesService.java#L91-L93

My suggestion is to either:

1. update the API documentation for this endpoint to explicitly say "List latest uploaded version of existing processes"
2. add an endpoint to retrieve all versions of a given processId
3. both

I'm happy to contribute if you think above solution is feasible.

quinarygio:
Fix #2401 

Release notes
In Features section:
#2401 : Add endpoint to list all versions of a business process

### DIFF
--- a/services/businessconfig/src/main/modeling/swagger.yaml
+++ b/services/businessconfig/src/main/modeling/swagger.yaml
@@ -16,8 +16,8 @@ schemes:
 paths:
   /businessconfig/processes:
     get:
-      summary: List existing processes
-      description: List existing processes
+      summary: List latest version of existing processes
+      description: List latest version of existing processes
       operationId: getProcesses
       produces:
         - application/json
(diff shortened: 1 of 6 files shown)


src/test/api/karate/businessconfig/getBusinessconfigHistory.feature
```diff
@@ -0,0 +1,50 @@
...
+    When method GET
+    Then status 401
+
+  Scenario: List not existing Businessconfig history
+
+    # Check bundle
+    Given url opfabUrl + '/businessconfig/processes/history/not_existent'
+    And header Authorization = 'Bearer ' + authToken
+    When method GET
+    Then status 200
```
vlo-rte, line 48: The endpoint businessconfig/processes/{processId} returns 404 in case of process not existing, I think this new endpoint should also return 404 in this case

services/businessconfig/src/main/modeling/swagger.yaml
```diff
@@ -114,6 +114,28 @@ paths:
...
+          type: string
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Process'
+        '401':
+          description: Authentication required
```
vlo-rte, line 138: as mentioned previously, we should also return 404 I think

services/businessconfig/src/main/modeling/swagger.yaml
```diff
@@ -114,6 +114,28 @@ paths:
           description: Not found  
         '500':
           description: Unable to delete process
+  '/businessconfig/processes/history/{processId}':
```
vlo-rte, line 117: If we have a process called 'history', it can be a problem I think. What do you think of this endpoint /businessconfig/processhistory/{processId} ?

vlo-rte (changes requested):

vlo-rte (approved):

Issue #2558, opened 2022-02-15
No sound on external device when session ended


freddidierRTE:
When session ended , the UI can not send anymore the request to the external-device service as the token has expired 

Solution : close session before token expiration 

The time to close session before real end is a parameter in web-ui.json and default value is one minute 

See regularCheckTokenValidity() in ui/main/src/app/services/authentication/authentication.service.ts

PR #2570, opened 2022-02-16
No sound on external device when session ended (#2558)

vlo-rte:
Release notes : 

In bugs section : 

#2558 : No sound on external device when session ended


ui/main/src/app/services/authentication/authentication.service.ts
```diff
@@ -154,10 +154,12 @@ export class AuthenticationService {
      * @return true if the expiration date stored in the `localestorage` is still running, false otherwise.
      */
     verifyExpirationDate(): boolean {
+        const secondsToCloseSession = this.configService.getConfigValue('secondsToCloseSession', 60);
```
freddidierRTE, line 157: I would prefer having this const global and initialize in   assignConfigurationProperties()

PR #2598, opened 2022-02-18
Solve bug -  Business menu: clicking on different menus always displays the first page loaded in iframe

freddidierRTE:
Nothing in release note as bug was introduce on the current release

### DIFF
--- a/ui/main/src/app/modules/businessconfigparty/iframedisplay.component.ts
+++ b/ui/main/src/app/modules/businessconfigparty/iframedisplay.component.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/businessconfigparty/iframedisplay.component.ts
```diff
@@ -46,30 +48,31 @@ export class IframeDisplayComponent implements OnInit, OnDestroy {
...
+  private addRoutingToUrl(url: string): string {
           // Capture everything between the last part of the menu and the first occurrence of question mark. (to avoid messing with parameters).
           // example: /menu1/entry1/deeplink?someParam=param turns into "/deeplink"
-          const additionalRouting = this.router.url.match(`.*${paramMap.get('menu_id')}\/${paramMap.get('menu_entry_id')}(\/[^?]*)`);
-          return additionalRouting ? additionalRouting[1] : '';
-        }),
-        map(routing => url += routing));
+          const additionalRouting = this.router.url.match(`.*${this.menu_id}\/${this.menu_entry_id}(\/[^?]*)`);
+          const to_add = additionalRouting ? additionalRouting[1] : '';
+          return url += to_add;
```
quinarygio, line 75: Should'nt it be : "return url +add"  ?
This would solve also sonar code smells

quinarygio (changes requested):

Issue #2557, opened 2022-02-15
Start even if translation file is not present for a declared language


freddidierRTE:
Starting opfab with a reference of none existing translation declared in web-ui prevent opfab to start . 

Instead , ignore the error , add just a warning in console.log , for example "Language ... is not supported"

PR #2628, opened 2022-02-24
Start even if translation file is not present for a declared language #2557

vlo-rte:
Release notes : 

In Tasks section : 

#2557 : Start even if translation file is not present for a declare language

### DIFF
--- a/ui/main/src/app/app.component.ts
+++ b/ui/main/src/app/app.component.ts
@@ -18,7 +18,7 @@ import {selectIdentifier} from '@ofSelectors/authentication.selectors';
 import {ConfigService} from '@ofServices/config.service';
 import {TranslateService} from '@ngx-translate/core';
 import {catchError, skip} from 'rxjs/operators';
-import {merge} from 'rxjs';
+import {merge, Observable, Subject} from 'rxjs';
 import {I18nService} from '@ofServices/i18n.service';
 import {CardService} from '@ofServices/card.service';
 import {UserService} from '@ofServices/user.service';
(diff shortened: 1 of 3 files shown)


ui/main/src/app/app.component.ts
```diff
@@ -168,14 +168,14 @@ export class AppComponent implements OnInit {
...
-        this.loadTranslationForMenu();
-        this.launchAuthenticationProcess();
+
+      localeRequests$.forEach(localeRequest$ => {
+          localeRequest$.subscribe(result => {
+            this.loadTranslationForMenu();
+          });
       });
+      this.launchAuthenticationProcess();
     }
```
quinarygio, line 178: I have a doubt:  in this way we are not waiting for all the locales to be loaded before starting.  
Maybe keeping the merge() and removing ".pipe(skip(localeRequests$.length - 1))" would solve the issue.

ui/main/src/app/services/i18n.service.ts
```diff
@@ -52,8 +52,11 @@ export class I18nService {
 
     public loadLocale(locale: string): Observable<any> {
         return this.httpClient.get(`${I18nService.localUrl}${locale}.json`).pipe( tap({
-            next: translation => this.translate.setTranslation(locale, translation, true),
-            error: error => console.log(new Date().toISOString(),`Error : impossible to load locale ${I18nService.localUrl}${locale}.json`)
+            next: translation => {
+                this.translate.setTranslation(locale, translation, true);
+                console.log(new Date().toISOString(), 'opfab translation loaded for locale:', locale);
```
quinarygio, line 57: Should we start using the new OpfabLoggerService instead of console log?

quinarygio (changes requested):

ui/main/src/app/services/i18n.service.ts
vlo-rte: You're right. Done.

ui/main/src/app/app.component.ts
vlo-rte: Yes, with my solution, we are not waiting for all locales to be loaded before starting.
The problem with the solution you propose is sometimes all languages are not completely loaded, with your solution as soon as we have one result the subscribe is finished.

Finally, I think we've found the solution, with the link you found : https://stackoverflow.com/questions/41734921/rxjs-wait-for-all-observables-in-an-array-to-complete-or-error.

quinarygio (approved):

Issue #2581, opened 2022-02-18
In settings, control the value "Interval between sound replays" 


freddidierRTE:
Value shall be a number > 0

PR #2631, opened 2022-02-25
In settings, control the value "Interval between sound replays" (#2581)

vlo-rte:
Release notes : 

In Bugs section : 

#2581 : In settings, control the value "Interval between sound replays"


quinarygio:
Error message when entering wrong characters is not localized, but maybe we can do it in another issue..

vlo-rte:
> Error message when entering wrong characters is not localized, but maybe we can do it in another issue..

No problem. You're right, Ive made an update for having a localized message. Thank you.

quinarygio (approved):

Issue #2592, opened 2022-02-18
Administration screen : create new user/entity/group with existing id is permitted whereas it should not


freddidierRTE:
it you do so you erase the existing object with the same id (for example user with same login)

It should not be possible to create an object with existing id

PR #2633, opened 2022-02-25
Administration screen : create new user/entity/group with existing id is permitted whereas it should not (#2592)

vlo-rte:
Release notes : 

In Bugs section : 

#2592 : Administration screen : create new user/entity/group with existing id is permitted whereas it should not

### DIFF
--- a/ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
+++ b/ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
@@ -8,7 +8,7 @@
  * This file is part of the OperatorFabric project.
  */
 
-import {FormControl, FormGroup, Validators} from '@angular/forms';
+import {AsyncValidatorFn, FormControl, FormGroup, Validators} from '@angular/forms';
 import {Component, Input, OnInit} from '@angular/core';
 import {User} from '@ofModel/user.model';
 import {NgbActiveModal} from '@ng-bootstrap/ng-bootstrap';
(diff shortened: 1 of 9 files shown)


ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
```diff
@@ -126,14 +128,24 @@ export class EditUserModalComponent implements OnInit {
     });
   }
 
+  checkExistingLogin() {
+    if (!! this.userForm.value['login'])
+      this.crudService.getAllUsers().subscribe(users => {
+        if (users.filter(user => user.login === this.userForm.value['login']).length)
+          this.isExistingLogin = true;
+      });
```
quinarygio, line 136: set this.isExistingLogin = false when no matching user is found

quinarygio:
Validaton is now performed when focusing out of input field.  It would be nice to implement it as a custom async validator and check the value during text input with a debounce time.
see:
https://stackoverflow.com/questions/44183917/how-to-debounce-async-validator-in-angular-4-with-rxjs-observable/49403153
https://stackoverflow.com/questions/36919011/how-to-add-debounce-time-to-an-async-validator-in-angular-2

quinarygio (changes requested):

ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
vlo-rte: Done.

quinarygio (approved):

PR #2703, opened 2022-03-16
Add child cards in archive detail cards (#2560)

### LINKED ISSUE #2560: Add child cards in archive detail cards
As for card in feed , process child cards when rendering a card in archives 

We need to retrieve child cards from mongoDB taking into account that a child card : 
- can be updated 
- can persist or not when updating parent card depending on the value of keepChildCards 

For the child update , the rule is : take the last update (revelant to the parent card )

To ease the search in db , i could be interesting to store the date when a card is deleted (adding a deletionDate parameter in the archive card) . So to search child card , we can do : 

Having the parent card take all card with parentCardId == parentcard.id and (deletionDate == null OR deletionDate > parentcard.publishDate) and ( parent.deletionDate ===null OR publishDate < parent.deletionDate )

quinarygio:
Fix #2560 

Release notes
In Features section:
#2560 : Add child cards in archive detail cards 

### DIFF
--- a/services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/ArchivedCardRoutesConfig.java
+++ b/services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/ArchivedCardRoutesConfig.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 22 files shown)


freddidierRTE:
I made some test , the solution with null / not exist is not working well in fact . It works well for old child cards but not for old archives parent cards

You endup with old archived parent card showing child cards of current parent card : It is because for old archived card deletetion date is null in the java object  , and so it is consider as a current card  (with ``` if (parentCard.getDeletionDate() == null) { ``` )

So i propose another solution , ask  to set the value deletionDate to 0 for all archived cards after migration  (via a mongo command) . And in the code do not get child card when parent card has deletion date == 0

Then we do not need anymore to insert null in the field by default 

The command to launch is to be put in a migration document

Issue #2712, opened 2022-03-18
Add the possibility for the usercard template to set the list of recipient choices or the list of recipients


freddidierRTE:
The use of configuration  in state (config.json) to set these lists is to be set as deprecated either for setting the list of recipient choices for user or for setting the recipient list when dropdown list is not visible

PR #2714, opened 2022-03-18
Add the possibility for the usercard template to set the list of recipient choices or the list of recipients

freddidierRTE:
Add in release note

Feature : 
- #2712 Add the possibility for the usercard template to set the list of recipient choices or the list of recipients


src/docs/asciidoc/reference_doc/user_cards.adoc
```diff
@@ -110,6 +110,7 @@ The following card fields can be set via the object `card` in the object returne
 * secondsBeforeTimeSpanForReminder
 * severity (in case it is not visible from the user , when `severityVisible` set to false in `config.json`)
 * data
+* entityRecipients (in case recipient list is not visible form the user)
```
quinarygio, line 113: from instead of form

quinarygio (changes requested):

PR #2739, opened 2022-03-22
Add the possiblity to automatically pin card (#2692)

### LINKED ISSUE #2692: Add the possiblity to automatically pin card
image

Pin card are always present even if it is not in business period 

When clicking on a  pinned card , it open the detail (like when clicking on a card on the feed) 

**Automatic mode :** 

A card is pinned if : 
- in the state definition  , there is the value automaticPinWhenAcknowledged set to true 
- the card is acknowlegded
- if endDate exist : endDate < currentdate 

 To do : 
- use "Network contingency" example as example for pin card 
- Cypress test
- Add in documentation

**Future issue for information : Manual Mode** 

When clicking on the cross , the card is unpinned
when clicking on the pin icon top/right the  card is pinned 

When user connect , the ui get the pinned card from the back 

Implementation for manual mode 

- Pin card list (uids) are stored in a new mongo collection 
- An API permit to access the user pin cards , the api is on card consultation service

quinarygio:
Fix #2692

Release notes
In Features section:
#2692 : Add the possiblity to automatically pin card 

### DIFF
--- a/services/businessconfig/src/main/java/org/opfab/businessconfig/model/ProcessStatesData.java
+++ b/services/businessconfig/src/main/java/org/opfab/businessconfig/model/ProcessStatesData.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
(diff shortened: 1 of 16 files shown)


freddidierRTE:
image

- add a padding on the right
- the card color shall be a the same "level" (top and bottom is not align) as the background card color

freddidierRTE (changes requested):

Issue #2728, opened 2022-03-21
External devices adaptation


freddidierRTE:
When defining external devices, it is necessary sometimes to adress the same device (IP:PORT) with a different signal config.
This is to address the fact that we have more than one loudspeakers and want to activate a precise one. 

With this configuration, the request for the watchdog is sent more than once to the device.

--> Improve it by sending it only one request at the required frequency even if there is different config for the same device.

PR #2763, opened 2022-03-28
External devices adaptation (#2728)

vlo-rte:

### DIFF
--- a/services/external-devices/src/main/java/org/opfab/externaldevices/services/DevicesService.java
+++ b/services/external-devices/src/main/java/org/opfab/externaldevices/services/DevicesService.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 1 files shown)


services/external-devices/src/main/java/org/opfab/externaldevices/services/DevicesService.java
```diff
@@ -93,14 +90,21 @@ public void sendSignalForUser(String opFabSignalKey, String userLogin) throws Ex
 
     @Scheduled(cron = "${operatorfabric.externaldevices.watchdog.cron:*/5 * * * * *}", zone = "UTC")
     public void sendWatchdog() {
-        if(externalDevicesWatchdogProperties.getEnabled()) {
+        if (externalDevicesWatchdogProperties.getEnabled()) {
+
+            Map<String, Boolean> hostAndPortMap = new HashMap<>();
```
freddidierRTE, line 95: use HashSet which is more adapted in this case (no need to store a Boolean)

freddidierRTE (changes requested):

services/external-devices/src/main/java/org/opfab/externaldevices/services/DevicesService.java
vlo-rte: Done.

Issue #2668, opened 2022-03-09
Do not permit to have the same user connected twice 


freddidierRTE:
When a user connect itself and is already connected , he shall receive a popup to confirm his connection.

Popup text : "You have already a session open with this user , if you open this connection it will close the existing session" 

When he confirm , the old connection is closed 

This shall be an option to activate on an instance basis . 

Conception : before opening the subscription , call an endpoint to check if there is already a subscription 

If user confirm , open subscription , the back will close the old subscription sending a specific message to the old connection

PR #2779, opened 2022-03-29
Disable a user to have several sessions at the same time

olivierPigeon-RTE:

- In release note :
  -  In chapter :  Features
  -  Text : #2668  Do not allow the same user to be connected twice at the same time


freddidierRTE:
add a chapter to in https://opfab.github.io/documentation/current/deployment/#_business_service_specific_configurations

To precise the definition of the parameter checkIfUserIsAlreadyConnected

Issue #2855, opened 2022-04-07
Solve aggrid deprecated feature


freddidierRTE:
![image](https://user-images.githubusercontent.com/58251627/162161333-55847acf-04b0-4b91-8501-70becef0735c.png)
```
       cons:warn ❖  AG Grid: since v27.0, `suppressCellSelection` has been replaced by `suppressCellFocus`.
       cons:warn ❖  AG Grid: As of v27, registering components via grid property frameworkComponents is deprecated. Instead register both JavaScript AND Framework Components via the components property.
```

Found in "Check monitoring cards reception" test in src/test/cypress/cypress/integration/Monitoring.spec.js

PR #2880, opened 2022-04-12
Solve aggrid deprecated feature (#2855)

freddidierRTE:
Nothing in release note


ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -47,7 +47,7 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
       context: {
         componentParent: this
       },
-      frameworkComponents : {
+      components : {
```
olivierPigeon-RTE, line 50: replace frameworkComponents by components line 402 too

PR #2930, opened 2022-04-21
Deal with groups of entities in acknowledged footer (#2848)

### LINKED ISSUE #2848: Deal with groups of entities in acknowledged footer
When we send a card to a group of entities , the acknowledged footer contains only the group and we do not show the sub entities

The idea would be to extract all the child entities (recursively)

quinarygio:
Fix #2848

- In release note :
  -  In chapter :  Features
  -  Text : #2848 : Deal with groups of entities in acknowledged footer


freddidierRTE:
Bug with update in realtime  : 
- connect as operator1_fr with  browser  1 
- Send a card to French Control Center only
- Open the card in the feed : you see four control center in orange on the footer
- Open a new browser , connect as user operator2_fr , ack the card 
- go back to browser one : you still see four control center in orange on the footer , you should see control center fr south in green
- if you close and repopen the card you finally see control center fr south in green


I guess pb is within method addAckFromSubscription

freddidierRTE (changes requested):

PR #3023, opened 2022-05-18
Choose your entities on login (#2861)

### LINKED ISSUE #2861: Choose your entities on login
When user is attach to more than one entities being allowed to send card , he should confirm the entities is acting one at that time when login .

The choice is made using the activity area screen as a popup

The option can be activated via a configuration in web-ui.json (not activated by default)

quinarygio:
Fix #2861

- In release note :
  -  In chapter :  Features
  -  Text : #2861: Choose your entities on login

### DIFF
--- a/config/docker/ui-config/web-ui.json
+++ b/config/docker/ui-config/web-ui.json
@@ -3,6 +3,7 @@
   "environmentColor": "blue",
   "showUserEntitiesOnTopRightOfTheScreen": true,
   "externalDevicesEnabled": true,
+  "selectActivityAreaOnLogin" : false,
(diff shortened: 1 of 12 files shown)


freddidierRTE:
- Add a cypress test 
- Add the parameter  selectActivityAreaOnLogin in web-ui.json (set it to false)
- Add a larger pop-up to take into account the fact that the number of column can be greater , the size of the usercard popup for example 
- Do not ask for confirm when saving config (when opening the component via the right menu , it should keep the confirm pop up) 
- Change the title of the popup to "Choose you activity area"  (when opening the component via menu , the actual title shall remain)

Issue #3001, opened 2022-05-16
External devices configuration screen : "all users devices are already configured" displayed with slow network


vlo-rte:
With slow network,  when you try to add another user device configuration, you see briefly "all users devices are already configured"
(and then you have user login , and a few seconds later, devices multi-select)

PR #3081, opened 2022-05-31
Add a spinner when loading external devices list in modal (#3001)

olivierPigeon-RTE:

- In release note :
  -  In chapter :  Bugs
  -  Text : #3001 : Add a spinner when loading external devices list in modal


freddidierRTE:
We shall not see the spinner when editing a user . In fact we should not have to load the list of user if in edition mode.

image

Issue #3005, opened 2022-05-16
Add a message in card detail when trying to load an none existing cards


freddidierRTE:
Message : card does not exist

PR #3111, opened 2022-06-07
Add text when the card is not found (#3005)

olivierPigeon-RTE:

- In release note :
  -  In chapter :  Tasks
  -  Text : #3005 :  Add text when trying to open a card that is not found in the feed

### DIFF
--- a/src/test/cypress/cypress/integration/FeedTests.spec.js
+++ b/src/test/cypress/cypress/integration/FeedTests.spec.js
@@ -7,10 +7,15 @@
  * This file is part of the OperatorFabric project.
  */
 
-describe ('FeedScreen tests',function () {
+describe('FeedScreen tests', function () {
+
+    function tryToLoadNonExistingCard() {
+        cy.visit('#/feed/cards/thisCardDoesNotExist');
+
+        cy.get('#opfab-feed-card-not-found').should('exist');
+    }
 
     before('Set up configuration', function () {
-
         // This can stay in a `before` block rather than `beforeEach` as long as the test does not change configuration
         cy.resetUIConfigurationFiles();
 
@@ -25,8 +30,7 @@ describe ('FeedScreen tests',function () {
 
 
     it('Check card reception and read behaviour', function () {
-
-        cy.loginOpFab('operator1_fr','test');
+        cy.loginOpFab('operator1_fr', 'test');
 
         // Set feed sort to "Date" so the cards don't move down the feed once they're read
         cy.get('#opfab-feed-filter-btn-sort').click();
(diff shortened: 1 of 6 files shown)


ui/main/src/app/modules/cards/components/card-details/card-details.component.html
```diff
@@ -24,4 +25,6 @@
 
   <br/>
   <br/>
-</div>
\ No newline at end of file
+</div>
+
+<div *ngIf="cardNotFound" id="opfab-feed-card-not-found" style="font-size:3em; text-align:center;" translate>feed.cardNotFound</div>
```
freddidierRTE, line 30: Smaller font and text should be more down (it is too close to the timeline)

src/test/cypress/cypress/integration/Translation.spec.js
```diff
@@ -351,6 +334,21 @@ describe('Test translations', function () {
         checkNotificationSeverityTexts('Alarm', 'Actie', 'Conform', 'Informatie');
         checkAknowledgementTexts('Bevestigen', 'Alles', 'Bevestigd', 'Niet bevestigd');
         checkDateFilterTexts('Ontvangstdatum', 'START', 'EIND', 'Filters toepassen op tijdlijn', 'Opnieuw installen');
-    })
+    });
 
-})
\ No newline at end of file
+    it('Check Feed translations', function () {
```
freddidierRTE, line 339: it('Check translation for non-existent card' ...

src/test/cypress/cypress/integration/FeedTests.spec.js
```diff
@@ -61,15 +64,16 @@ describe ('FeedScreen tests',function () {
...
             .as('firstCardUrlId')
             .then((urlId) => {
-                cy.hash().should('eq', '#/feed/cards/'+urlId);
+                cy.hash().should('eq', '#/feed/cards/' + urlId);
                 cy.get('of-card-details').find('of-detail');
             });
+        cy.get('#opfab-feed-card-not-found').should('not.exist');
 
-
+        tryToLoadNonExistingCard();
```
freddidierRTE, line 76: I think it is not necessary to have this test as it's redundant with tests in translation.spec.ts

olivierPigeon-RTE: I did 2 separate tests in 2 different places that have different goals : 
* the one in FeedTest tests the card that was displayed before trying to load a non existing card is not displayed anymore and is replaced by a text (the value of the text is not tested here). And the same way, when we load an existing card, the texts disappears and the real card is shown
* the test in Translation chacks the value of the string in the message

I think these are 2 very different things, this is why I did separate tests in separate files

freddidierRTE: ok

Issue #3105, opened 2022-06-03
Sort alphabetically state rights in perimeter administration table


freddidierRTE:
Sort alphabetically the states right values : 

![image](https://user-images.githubusercontent.com/58251627/171802818-a6f2d4ad-eb3e-4a39-b15e-26561e6ca97d.png)

PR #3115, opened 2022-06-08
Sort alphabetically state rights in perimeter administration table (#3105)

vlo-rte:
I've also sorted alphabetically the columns "groups" and "entities" (in users management), and "perimeters" column (in groups management).

- In release note :
  -  In chapter :  Features
  -  Text : #3105 : Sort alphabetically state rights in perimeter administration table

### DIFF
--- a/ui/main/src/app/modules/admin/components/cell-renderers/array-cell-renderer.component.ts
+++ b/ui/main/src/app/modules/admin/components/cell-renderers/array-cell-renderer.component.ts
@@ -37,7 +37,7 @@ export class ArrayCellRendererComponent implements ICellRendererAngularComp {
                 .getValue()
                 .map((code) => {
                     // This entails that the items that need to be rendered have an `id` and a `name` property.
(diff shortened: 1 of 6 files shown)


ui/main/src/app/modules/admin/components/cell-renderers/array-cell-renderer.component.ts
```diff
@@ -49,6 +49,8 @@ export class ArrayCellRendererComponent implements ICellRendererAngularComp {
                     }
                 })
                 .join(', ');
+
+            this._nameValues = this._nameValues.split(', ').sort().join(', ');
```
freddidierRTE, line 53: Why not : 
```
                        return code;
                    }
                })
                .sort()
                .join(', ');
        } else {
```

vlo-rte: You're right. Done.

Issue #3101, opened 2022-06-02
Remove environment choice in setSecurityAuthFlow.sh


freddidierRTE:
Remove environment choice in src/test/resources/uiConfig/setSecurityAuthFlow.sh as we now use the same web-ui.json file between dev and docker env (the one in the config/docker/ui-config repository)

PR #3122, opened 2022-06-09
Remove environment choice in setSecurityAuthFlow.sh (#3101)

olivierPigeon-RTE:

- In release note :
  -  In chapter : Tasks
  -  Text : #3101 : Remove environment choice in setSecurityAuthFlow.sh


freddidierRTE:
No need to put information in the migration guide as it is only a dev tool , migration guide is dedicated to migration of end user instances

PR #3139, opened 2022-06-14
Add possibility to use multiselect as simple select (#3126)

### LINKED ISSUE #3126: Add possibility to use multiselect as simple select
In opfab.js , add a configuration option to act as simple select (option multiple set to false in virtual select) 

Change init function signature to : 

init : function(config) , config having the following structure  : 
```
{
id : string 
options : array of values  
multiple : boolean 
}
```

Modify documentation accordingly  

Use the component in process example : 
image

### LINKED ISSUE #3143: Missing translations for opfab.multiSelect on startup
Translations for opfab.multiSelect component should be loaded on startup after loading locales.

quinarygio:
Fix #3126 Fix #3143

- In release note :
  -  In chapter :  Features
  -  Text : #3126 : Add possibility to use multiselect as simple select
 
  -  In chapter :  Bugs
  -  Text : #3143 Missing translations for opfab.multiSelect on startup

### DIFF
--- a/src/docs/asciidoc/reference_doc/template_description.adoc
+++ b/src/docs/asciidoc/reference_doc/template_description.adoc
@@ -622,14 +622,19 @@ OperatorFabric provides a multiselect component based on https://sa-si-dev.githu
   </div>
 
  <script>
-  myMultiSelect = opfab.multiSelect.init("my-multiselect",[
+  myMultiSelect = opfab.multiSelect.init({
+                id: "my-multiselect",
+                options: [
                 { label: 'Choice A', value: 'A' },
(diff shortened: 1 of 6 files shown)


freddidierRTE:
1) I do not know if the bug was present before but in edition mode the status of the state is not set to the current value .

image


2) Add a search feature to the select : https://sa-si-dev.github.io/virtual-select/#/examples?id=with-search-box  (it is not activate by default when there multiple is set to false ) 
Add it as an option , set it to true for "State of the process" and false for "Status of the state"

3) Add a cypress test using it

PR #3150, opened 2022-06-16
Avoid loosing acknowlegments when connection lost (#3083)

### LINKED ISSUE #3083: Loosing acknowlegments when connection lost
When user lost connection due for example to a network pb , the subscription closes and reopen as soon as connection is possible . 
The card that arrived during the connection loss are retrieve by the ui via a request for card received during the period .

This solution does not take into account acquittement received during the "lost connection period" , so we may lost ack when we have connection lost .
To solve this , add a lastAckDate field in mongo card collection so we can retrieve card that was acknowledged during the "lost connection period" via this field.

quinarygio:
Fix #3083

- In release note :
  -  In chapter :  Bugs 
  -  Text : #3083 : Avoid loosing acknowlegments when connection lost


freddidierRTE:
Add an index on mongoDB for the new field , use annotation @Indexed

PR #3188, opened 2022-06-20
Refacto Time service in DateTimeFormatter service

freddidierRTE:
Nothing in release note


ui/main/src/app/common/utilities.ts
```diff
@@ -58,4 +60,29 @@ export class Utilities {
...
+       if (!ngbDateTime.date) return null;
+       dateFromNgb.setFullYear(ngbDateTime.date.year);
+       dateFromNgb.setDate(ngbDateTime.date.day);
+       dateFromNgb.setMonth(ngbDateTime.date.month-1);
+       dateFromNgb.setHours(ngbDateTime.time.hour);
+       dateFromNgb.setMinutes(ngbDateTime.time.minute);
+       dateFromNgb.setSeconds(ngbDateTime.time.second);
+       return dateFromNgb.valueOf();
+    }
+
```
vlo-rte, line 77: I remember a bug we had before. If I remember well, there is a risk to have for example : 
if current date is 2022/02/15 and the datetime parameter is 2022/03/31
Then the function will try to set 2022/02/31 (line 70) and it will crash no ?

ui/main/src/app/common/utilities.ts
```diff
@@ -58,4 +60,29 @@ export class Utilities {
...
     }
+
+
+    public static convertNgbDateToEpochDate(ngbDateTime: DateTimeNgb) : number {
+       const dateFromNgb = new Date();
+       if (!ngbDateTime) return null;
+       if (!ngbDateTime.date) return null;
+       dateFromNgb.setFullYear(ngbDateTime.date.year);
+       dateFromNgb.setDate(ngbDateTime.date.day);
+       dateFromNgb.setMonth(ngbDateTime.date.month-1);
```
vlo-rte, line 71: I remember a bug we had before. If I remember well, there is a risk to have for example :
if current date is 2022/02/15 and the datetime parameter is 2022/03/31
Then the function will try to set 2022/02/31 (line 70) and it will crash no ?

freddidierRTE: i made a test , it does not crash . 

If you set for example  the date time to 2022/02/31 , it will return the date 2022/02/01 (epoch)

vlo-rte (approved):

Issue #3146, opened 2022-06-16
Usercard  : Replace single filter component


freddidierRTE:
Replace all single filter  with new multiselect component in user card 

Add on option to set multiple to false in multi-select component  (default value is true)

PR #3205, opened 2022-06-22
Usercard : Replace single filter component (#3146)

quinarygio:

- In release note :
  -  In chapter :  Tasks
  -  Text : #3146  : Usercard : Replace single filter component


vlo-rte (changes requested):
As discussed today, at the daily, I propose you to add the type of the parameters for functions and also to add the return type in the declaration of them (except when it is void).

quinarygio:
> As discussed today, at the daily, I propose you to add the type of the parameters for functions and also to add the return type in the declaration of functions (except when it is void).

Done

vlo-rte (approved):

PR #3243, opened 2022-06-29
Provide a way to access all cards in archives/logging screens for admin user (#3145)

vlo-rte:
I decided to name new parameter "adminMode" and not "isAdminMode" in the request to the back to stay coherent with the existing parameter "latestUpdateOnly"... 
 
- In release note :
  -  In chapter : Features
  -  Text : #3145 : Provide a way to access all cards in archives/logging screens for admin user

### DIFF
--- a/services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/ArchivedCardCustomRepositoryImpl.java
+++ b/services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/ArchivedCardCustomRepositoryImpl.java
@@ -64,11 +64,13 @@ public class ArchivedCardCustomRepositoryImpl implements ArchivedCardCustomRepos
 
     public static final String LATEST_UPDATE_ONLY = "latestUpdateOnly";
 
+    public static final String ADMIN_MODE = "adminMode";
+
     public static final int DEFAULT_PAGE_SIZE = 10;
 
     private static final List<String> SPECIAL_PARAMETERS = Arrays.asList(
             PUBLISH_DATE_FROM_PARAM, PUBLISH_DATE_TO_PARAM, ACTIVE_FROM_PARAM, ACTIVE_TO_PARAM,
-            PAGE_PARAM, PAGE_SIZE_PARAM, CHILD_CARDS_PARAM, LATEST_UPDATE_ONLY);
+            PAGE_PARAM, PAGE_SIZE_PARAM, CHILD_CARDS_PARAM, LATEST_UPDATE_ONLY, ADMIN_MODE);
 
 
     private static final String ARCHIVED_CARDS_COLLECTION = "archivedCards";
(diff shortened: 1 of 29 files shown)


services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/ArchivedCardCustomRepositoryImpl.java
```diff
@@ -308,6 +318,17 @@ private boolean isLatestUpdateOnlyParamPresent(MultiValueMap<String, String> par
         return false;
     }
 
+    private boolean isAdminModeParamPresent(CurrentUserWithPerimeters currentUserWithPerimeters,
```
freddidierRTE, line 321: I would name it checkIfInAdminMode as the method is not only checking the param but as well checking the user is admin 
I would put a log with warning level if we want to be in admin mode but we are not member of the admin group

services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/UserUtilitiesCommonToCardRepository.java
```diff
@@ -58,7 +58,11 @@ default Flux<T> findByInitialParentCardUid(ReactiveMongoTemplate template, Strin
     default List<Criteria> computeCriteriaToFindCardByIdWithUser(String id, CurrentUserWithPerimeters currentUserWithPerimeters) {
         List<Criteria> criteria = new ArrayList<>();
         criteria.add(Criteria.where("_id").is(id));
-        criteria.add(computeCriteriaForUser(currentUserWithPerimeters));
+
+        boolean isCurrentUserMemberOfAdminGroup = ((currentUserWithPerimeters.getUserData().getGroups() != null) &&
+                                                   (currentUserWithPerimeters.getUserData().getGroups().contains("ADMIN")));
```
freddidierRTE, line 63: i would factorize this method as we need it at least in two places 
adding the method inside CurrentUserWithPerimeters class 
and maybe a unit test

freddidierRTE:
Add karate tests :
- check with admin account either in admin mode or not , i guess in file : src/test/api/karate/cards/fetchArchivedCardsWithParams.feature
- check to get a card in archives with admin even if card is not in perimeter , to check he sees the card

freddidierRTE (changes requested):

services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/ArchivedCardCustomRepositoryImpl.java
vlo-rte: Done.

services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/UserUtilitiesCommonToCardRepository.java
freddidierRTE: After discussion not needed yet to factorize

vlo-rte:
> Add karate tests :
> 
>     * check with admin account either in admin mode or not , i guess in file : src/test/api/karate/cards/fetchArchivedCardsWithParams.feature

Done.

Issue #3167, opened 2022-06-17
Missing spinner in usercard


vlo-rte:
In usercard, when we change service or process or state, there is no spinner or message showing the template is loading, so sometimes the template is not the good one.
Add also a spinner after clicking "see before sending" button because it can take time too.

(see issue #3010)

PR #3264, opened 2022-07-05
Add spinners in user card creation screen (#3167)

olivierPigeon-RTE:

- In release note :
  -  In chapter :  Tasks
  -  Text : #3167 Add spinner in user card creation screen

### DIFF
--- a/src/test/cypress/cypress/integration/UserCard.spec.js
+++ b/src/test/cypress/cypress/integration/UserCard.spec.js
@@ -171,7 +171,7 @@ describe('User Card ', function () {
       cy.loginOpFab('operator1_fr', 'test');
       cy.get('#opfab-navbarContent').find('#opfab-newcard-menu').click();
       cy.usercardSelectService('User card examples');
-      cy.usercardSelectProcess('Message or question');;
+      cy.usercardSelectProcess('Message or question');
       cy.usercardSelectState('Message');
       cy.get('#opfab-usercard-startdate-choice').should("exist");
       cy.get('#opfab-usercard-enddate-choice').should("not.exist");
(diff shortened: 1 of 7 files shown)


src/test/cypress/cypress/integration/UserCard.spec.js
```diff
@@ -190,6 +190,30 @@ describe('User Card ', function () {
     })
   })
 
+  describe('Test spinners', function () {
+    it('Check spinner appears when card template is loading', () => {
+
+      cy.delayRequestResponse('/businessconfig/**/*');
```
freddidierRTE, line 196: Should work with /** only

src/test/cypress/cypress/integration/UserCard.spec.js
```diff
@@ -190,6 +190,30 @@ describe('User Card ', function () {
...
+
+      cy.loginOpFab('operator1_fr', 'test');
+      cy.get('#opfab-navbarContent').find('#opfab-newcard-menu').click();
```
freddidierRTE, line 199: check here the spinner as well 
   cy.checkLoadingSpinnerIsDisplayed();
    cy.checkLoadingSpinnerIsNotDisplayed();

ui/main/src/app/modules/share/card-detail/card-detail.component.html
```diff
@@ -9,5 +9,9 @@
 
 
 
-<link *ngFor="let cssUrl of hrefsOfCssLink" [href]="cssUrl" rel="stylesheet" type="text/css">
-<div id="opfab-card-template-detail" class="opfab-card-detail"  [innerHtml]="htmlContent" style="height: 100%;"></div>
\ No newline at end of file
+<of-spinner *ngIf="isLoading; else cardDetails"></of-spinner>
```
freddidierRTE, line 12: i would prefer using hidden feature for card detail : 
```
<div id="opfab-card-template-detail" class="opfab-card-detail"  [innerHtml]="htmlContent" [hidden]="isLoading"  style="height: 100%;"></div>
<of-spinner *ngIf="isLoading"></of-spinner>
```
This is to avoid setting innerHtml before ng template is creating and thus having the div not existing. I do not know if it can arise , but using hidden guarantee that the <div> exists .

ui/main/src/app/modules/usercard/usercard.component.html
```diff
@@ -48,21 +48,25 @@
                 (lttdFilterChange)="onLttdChange()"
                 #datesForm> </of-usercard-dates-form>
 
-    <div class="opfab-form-lighter" [innerHtml]="userCardTemplate"></div>
 
-    <br/>
+    <of-spinner *ngIf="isLoadingCardTemplate; else cardTemplate"></of-spinner>
```
freddidierRTE, line 52: same remark as previously , use hidden

freddidierRTE (changes requested):

Issue #3183, opened 2022-06-20
Usercard : error message not appropriate for users with no entities


vlo-rte:
When the user disconnect from all its entities, and then he tries to create a usercard, there is : 

![image](https://user-images.githubusercontent.com/57365872/174537169-2ea44791-eb8d-4b4c-bbf0-a934801e5857.png)

enlessly. Update the message to set an appropriate one.

PR #3276, opened 2022-07-06
Add an error message when a user without entity tries to send a card (#3183)

olivierPigeon-RTE:

- In release note :
  -  In chapter :  Tasks
  -  Text : #3183 :  Add an error message when a user without entity tries to send a card


src/test/cypress/cypress/integration/UserCard.spec.js
```diff
@@ -188,6 +188,35 @@ describe('User Card ', function () {
...
+    it('Users should not be able to send cards if it does not belong to any entity', () => {
+
+      cy.loginOpFab('operator1_fr', 'test');
+      
+      cy.get('#opfab-navbarContent').find('#opfab-newcard-menu').click();
+      cy.get('#opfab-usercard-creation').should('exist');
+      cy.get('#opfab-usercard-close').click();
+
+      // The user should not be able to send a card if it does not belong to any entity
+      cy.openOpfabActivityArea();
```
freddidierRTE, line 201: cy.openActivityArea

PR #3290, opened 2022-07-08
Remove test code

freddidierRTE:
Nothing in release note


quinarygio (changes requested):
Remove unused imports reported by sonar: 
import {NgbDate} from '@ng-bootstrap/ng-bootstrap';
import {DateTimeNgb} from '@ofModel/datetime-ngb.model';

PR #3300, opened 2022-07-11
Add an example for external web application for testing (#3253)

### LINKED ISSUE #3253: Add an example for external web application for testing
Add an external application example via url  /external/appExample define in nginx (dev + docker)

This point to the directory src/test/externalWebAppExample

It contains just an html file with opfab background color and with text "test application" 

This application should be accessible via "single menu entry menu" 

Add a cypress test to check it works

quinarygio:
Fix  #3253

- In release note :
  -  In chapter :  Features
  -  Text :  #3253 : Add an example for external web application for testing


freddidierRTE:
This application should be accessible via "single menu entry menu"
Modify accordingly src/test/cypress/cypress/integration/BusinessMenu.spec.js

PR #3303, opened 2022-07-11
Feed notification configuration : display problem if the user has no right for a process group (#3246)

### LINKED ISSUE #3246: Feed notification configuration : display problem if the user has no right for a process group
remove ReadOnly group for operator1_fr
modify Dispatcher group for having only cypress perimeter for this group
then you have : 
image

Then add taskExample perimeter to the Dispatcher group, you will see : 
 
image

quinarygio:
Fix #3246 

- In release note :
  -  In chapter :  Bugs
  -  Text : #3246 : Feed notification configuration : display problem if the user has no right for a process group


freddidierRTE:
it solve only part of the problem , we should not see process group in the following screen 
image

PR #3308, opened 2022-07-12
Externalize some cypress commands

freddidierRTE:
Nothing in release note


src/test/cypress/cypress/support/externalCommands.js
```diff
@@ -0,0 +1,14 @@
+export class externalCommands {
```
quinarygio, line 1: Add the copyright header

quinarygio (changes requested):

src/test/cypress/cypress/support/usercardCommands.js
```diff
@@ -0,0 +1,40 @@
+import {externalCommands} from './externalCommands';
```
quinarygio, line 1: Add the copyright header

quinarygio (changes requested):

PR #3333, opened 2022-07-15
Provide shared css/js to external applications (#3310)

### LINKED ISSUE #3310: Provide share css/js to external applications
Extract form ui/main/src/app/services/global-style.service.ts js code that could be used by external application to switch between night and day mode 

Add this js code in /ui/main/src/shared/js/opfab-application.js

Extract from /ui/main/src/scss/styles.scss , commun css rules between opfab and external app  ,and put it in file /ui/main/src/shared/opfab-application.css 

The common css are the css define in https://opfab.github.io/documentation/current/reference_doc/#opfab_template_style

Use this js/css in the external app example

quinarygio:
Fix #3310

- In release note :
  -  In chapter :  Tasks
  -  Text : #3310 : Provide shared css/js to external applications


freddidierRTE:
Add font files as share
Add  one block of component with classic opfab-bgcolor and one with opfab-bgcolor-darker
Add new chapter in documentation (https://opfab.github.io/documentation/current/reference_doc) named : External application integration . Explain you can use the shared files , refer to the list of component of templates and reference the application file example

freddidierRTE:
Font is not correct on login screen : 
image

Issue #3234, opened 2022-06-29
Refuse sending card if process or state does not exist 


freddidierRTE:
Return an HTTP error code and a message

PR #3451, opened 2022-08-12
Refuse sending card if process or state does not exist (#3234)

vlo-rte:
- In release note :
  -  In chapter :  Features
  -  Text : #3234 : Refuse sending card if process or state does not exist

### DIFF
--- a/client/businessconfig/src/main/java/org/opfab/businessconfig/model/AcknowledgmentAllowedEnum.java
+++ b/client/businessconfig/src/main/java/org/opfab/businessconfig/model/AcknowledgmentAllowedEnum.java
@@ -9,6 +9,8 @@
 
 package org.opfab.businessconfig.model;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 
@@ -32,4 +34,20 @@ public enum AcknowledgmentAllowedEnum {
     ONLY_WHEN_RESPONSE_DISABLED_FOR_USER("OnlyWhenResponseDisabledForUser");
 
     String value;
+
+    @Override
+    @JsonValue
+    public String toString() {
+        return String.valueOf(value);
+    }
+
+    @JsonCreator
+    public static AcknowledgmentAllowedEnum fromValue(String text) {
+        for (AcknowledgmentAllowedEnum b : AcknowledgmentAllowedEnum.values()) {
+            if (String.valueOf(b.value).equals(text)) {
+                return b;
+            }
+        }
+        return null;
+    }
 }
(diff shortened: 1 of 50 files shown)


services/cards-publication/src/main/java/org/opfab/cards/publication/controllers/CardController.java
```diff
@@ -56,6 +60,17 @@ public class CardController {
...
+
+
+        if ((user != null) && (!cardProcessingService.doesProcessStateExistInBundles(card.getProcess(),
+                card.getProcessVersion(), card.getState(), user.getUserData().getLogin()))) {
+            throw new ApiErrorException(
+                    ApiError.builder()
+                            .status(HttpStatus.BAD_REQUEST)
+                            .message(String.format(UNEXISTING_PROCESS_STATE, card.getProcess(), card.getState(), card.getProcessVersion(), card.getProcessInstanceId()))
+                            .build());
+        }
```
freddidierRTE, line 72: Should be done in cardProcessingService like other controls

src/test/api/karate/businessconfig/resources/packageBundles.sh
```diff
@@ -13,4 +13,12 @@ cd ..
 cd bundle_api_test_apogee 
 tar -czvf bundle_api_test_apogee.tar.gz config.json i18n.json css/ template/ 
 mv bundle_api_test_apogee.tar.gz ../
+cd ..
+cd bundle_defaultProcess_V1
+tar -czvf bundle_defaultProcess_V1.tar.gz config.json i18n.json css/ template/
+mv bundle_defaultProcess_V1.tar.gz ../
+cd ..
```
freddidierRTE, line 20: Must we really copy all files , could we just take the file form the resource directory ?

client/businessconfig/src/main/java/org/opfab/businessconfig/model/AcknowledgmentAllowedEnum.java
```diff
@@ -32,4 +34,20 @@ public enum AcknowledgmentAllowedEnum {
...
+    }
+
+    @JsonCreator
+    public static AcknowledgmentAllowedEnum fromValue(String text) {
+        for (AcknowledgmentAllowedEnum b : AcknowledgmentAllowedEnum.values()) {
+            if (String.valueOf(b.value).equals(text)) {
+                return b;
+            }
+        }
+        return null;
```
freddidierRTE, line 51: Why this is needed ?

freddidierRTE (changes requested):

client/businessconfig/src/main/java/org/opfab/businessconfig/model/AcknowledgmentAllowedEnum.java
vlo-rte: I can't figure out why I needed it. Weird...
(So I remove)

src/test/api/karate/businessconfig/resources/packageBundles.sh
vlo-rte: Done.

PR #3454, opened 2022-08-17
Use shared spinner component when loading opfab

### LINKED ISSUE #3362: Spinner for loading opfab : spinner gray and text white
See if we can replace it by the shared component.

quinarygio:
Fix #3362 

- In release note :
  -  In chapter :  Bugs 
  -  Text :  #3362 : Spinner for loading opfab : use shared spinner component

### DIFF
--- a/ui/main/src/app/modules/core/application-loading/application-loading.component.html
+++ b/ui/main/src/app/modules/core/application-loading/application-loading.component.html
@@ -6,16 +6,9 @@
 <!-- SPDX-License-Identifier: MPL-2.0                                      -->
 <!-- This file is part of the OperatorFabric project.                      -->
 
-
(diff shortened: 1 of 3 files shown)


vlo-rte (changes requested):
With slow network, first I see only the message "Application is loading ..." without spinner, and then after several seconds, the spinner appears. We used to have the spinner and the message displayed at the same time. Would it be possible to have the same behaviour ?

And also would it be possible to center the spinner and the message, like it was before ? 
Currently it looks like : 

image

vlo-rte:
I propose "padding-top: 15%;" instead of 20% , center is better I think.
I still have "Application is loading ...." displayed before the spinner (tested with chrome).

vlo-rte (approved):

Issue #3450, opened 2022-08-12
Geographical map: add a optional chart to display number of cards by severity


geppyz:
On the geographical map we would like to add a Graphical visualisation of the number of cards (grouped by severity) to be an optional extra.

<img width="1713" alt="Schermafbeelding 2022-08-12 om 13 48 50" src="https://user-images.githubusercontent.com/18466477/184305364-bbfa4d44-231a-4f0e-a8c6-a2c99c946aa6.png">

geppyz:
Current example:

<img width="1250" alt="Schermafbeelding 2022-08-12 om 15 59 25" src="https://user-images.githubusercontent.com/18466477/184322613-d5653c91-02f2-4229-9d64-d5eec986cc17.png">

PR #3455, opened 2022-08-18
Add Graph to Map and added labels and translation (#3450)

geppyz:
Signed-off-by: Gerben Danen <geppyz@gmail.com>

- In release note :
  -  In chapter :  Features
  -  Text : #3450

### DIFF
--- a/ui/main/package.json
+++ b/ui/main/package.json
@@ -45,6 +45,7 @@
     "bootstrap": "5.2.0",
     "chart.js": "3.9.1",
     "chartjs-adapter-date-fns": "2.0.0",
+    "chartjs-plugin-datalabels": "2.1.0",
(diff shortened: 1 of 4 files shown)


ui/main/package.json
```diff
@@ -45,6 +45,7 @@
     "bootstrap": "5.2.0",
     "chart.js": "3.9.1",
     "chartjs-adapter-date-fns": "2.0.0",
+    "chartjs-plugin-datalabels": "^2.1.0",
```
freddidierRTE, line 48: use the precise version of the component (this is to avoid delivering in a release a version that has not been fully tested) 
So :

`"chartjs-plugin-datalabels": "2.1.0"`

geppyz: Thanks, I changed it!

Issue #1318, opened 2021-06-30
Resource identifiers should only contain letters, digits, dashes or underscores


freddidierRTE:
Resources that are accessible through the API (Users, Groups, Entities, Perimeters, etc.) are usually reachable through a unique identifier (usually the "id" field, or "login" in the case of users)

If this id contains spaces or special characters it can cause issues in the API (when spaces are converted to %20 in urls for example) or for the persistence in Mongo, so we want to make sure that these identifiers will only contain:

 - letters
 - digits
 - dashes
 - underscores

There is already a validation in the UI, but we would also like to enforce it at the API level.

TODO

- [x] If a call is made to the API to create a resource with an identifier that is not compliant with this rule, it should return a bad request error with a message explaining the issue.
- [x] Add this requirement to the documentation (in the swagger.yml file?)
- [x] Add karate tests to test that it is correctly enforced
- [x] Make sure that all examples/default values from the documentation or getting started are compliant with this rule.

PR #3470, opened 2022-08-22
Resource identifiers should only contain letters, digits, dashes or underscores (#1318)

vlo-rte:
I've chosen to forbid uppercase letters in login and so, to delete calls to toLowerCase() functions to convert login. 
To be discussed maybe ?

- In release note :
  -  In chapter :  Features
  -  Text : #1318 : Resource identifiers should only contain letters, digits, dashes or underscores


vlo-rte:
After discussion, we authorize uppercase letters when creating a user via API, but it will be converted to lowercase.

freddidierRTE:
Simplify karate test via "Scenario outline feature"  : https://github.com/karatelabs/karate#data-driven-tests

PR #3507, opened 2022-09-02
Add export feature for admin screens (#3503)

### LINKED ISSUE #3503: Add export feature for admin screens
Add a button export (similar as for archives) that permit to export all data 

Do it at least for user list

quinarygio:
Fix #3503

- In release note :
  -  In chapter :  Features
  -  Text : #3503 : Add export feature for admin screens

### DIFF
--- a/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
+++ b/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
@@ -28,10 +28,12 @@ import {Group} from '@ofModel/group.model';
 import {Entity} from '@ofModel/entity.model';
 import {EntitiesService} from '@ofServices/entities.service';
 import {PerimetersCellRendererComponent} from '../cell-renderers/perimeters-cell-renderer.component';
+import {ExportService} from '@ofServices/export.service';
 
 @Directive()
 @Injectable()
 export abstract class AdminTableDirective implements OnInit, OnDestroy {
+
     constructor(
         protected translateService: TranslateService,
         protected confirmationDialogService: ConfirmationDialogService,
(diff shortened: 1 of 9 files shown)


vlo-rte:
Translation for titles of the columns in the export file is not taken into account. For example in french language, we have : 
image

vlo-rte:
I think we should display group name and entity name instead of ids, the same than what we display on the screen. Currently we have : 
image

vlo-rte:
For groups sheet : I think we should display YES/NO (etc...) instead of 1/0 for the column "realtime". I think we should also display the perimeters in alphabetical order (like it seems to be the case on the screen).
image

vlo-rte:
For perimeters sheet : I thinhk we should display the state name instead of its id. Currently we have : 
image

vlo-rte (changes requested):
Please see my comments.

ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -416,10 +418,80 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
...
+
+    private isStringArray(arr): boolean {
+        return arr.every((value) => {
+            return typeof value === 'string';
+        });
+    }
+
+    private getMappedName(id: string, renderer: string): string {
+        if (renderer) {
+            const celRenderer = new this.gridOptions.components[renderer].prototype.constructor();
```
freddidierRTE, line 489: cellRender instead of celRender

ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -416,10 +418,80 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
...
+    objectArrayToString(arr: any, renderer: string, data: any): string {
+        let json = JSON.stringify(arr);
+        if (json.length > 1) {
+            // Remove square brakets
+            json = json.substring(1,json.length -1)
+        }
+        return json;
+    }
+
+    private getExportHeader(header: string): string {
```
freddidierRTE, line 467: suggestion : this.getTranslatedHeaderName(header: string)

ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -416,10 +418,80 @@ export abstract class AdminTableDirective implements OnInit, OnDestroy {
...
+        else return this.objectArrayToString(arr, renderer, data);
+    }
+
+    private isStringArray(arr): boolean {
+        return arr.every((value) => {
+            return typeof value === 'string';
+        });
+    }
+
+    private getMappedName(id: string, renderer: string): string {
```
freddidierRTE, line 487: getNameFromId

PR #3538, opened 2022-09-09
Refuse cards if no translation file or no translation (#3178)

vlo-rte:
To be merged after https://github.com/opfab/operatorfabric-core/pull/3451 because branch FE-3178 was created from branch FE-3234. Thanks.

- In release note :
  -  In chapter :  Features
  -  Text : #3178 : Refuse cards if no translation file or no translation


freddidierRTE:
Do not check if option authorizeToSendCardWithInvalidProcessState is set to true in service configuration

Issue #3545, opened 2022-09-13
add a possibility to not close card when user acknowledge


freddidierRTE:
Add a parameter in state definition in config.json  named  closeCardWhenUserAcknowledge

Default value is true 

If value is false : 
- replace the actual button label with ACKNOWLEGDE instead of ACKNOWLEDGE AND CLOSE in card dettails
- do not close card when user acknowledge 

Add this option on process state card example : 

![image](https://user-images.githubusercontent.com/58251627/189862969-6968dbcc-942c-4c60-bc0c-d902e31451ef.png)

Add in karate test (modify an existing test or create a new one) 
Add in cypress tests 
Add in documentation:  https://opfab.github.io/documentation/current/reference_doc/#_acknowledgment_done_by_the_user

PR #3550, opened 2022-09-14
Add a possibility to not close card when user acknowledge (#3545)

quinarygio:

- In release note :
  -  In chapter :  Features
  -  Text : #3545 : Add a possibility to not close card when user acknowledge


freddidierRTE:
1) Bug : 
Go to process  card 
Ack the card 
The button should change to unacknowledge , it is not the case (you need to close and reopen the card to see it)

2) change value _closeCardWhenUserAcknowledge_ to  _closeCardWhenUserAcknowledge**s**_ (mistake of mine in the issue)

PR #3584, opened 2022-09-21
Add card sorting by start and end date (#3565)

### LINKED ISSUE #3565: Add card sorting by start and end date
Currently it is possible to sort cards by priority and creation date, but we are creating cards for future events and would like to sort the list so that the card which end date is closest is the first one, thus enabling operators to see the cards in the order they must be carried out, or were supposed to be carried out.

So I would like to ask, that it would be possible to sort cards by 1. start date and 2 end date

quinarygio:
Fix #3565

- In release note :
  -  In chapter :  Features
  -  Text : #3565 Add option to sort cards by start and end date

### DIFF
--- a/ui/main/src/assets/i18n/en.json
+++ b/ui/main/src/assets/i18n/en.json
@@ -171,9 +171,11 @@
     },
     "sort": {
       "title": "Sort by",
-      "severity": "Criticality then date",
-      "publish": "Date",
-      "unread": "Unread then date"
+      "severity": "Criticality then reception date",
+      "publish": "Reception date",
+      "unread": "Unread then reception date",
+      "startDate": "Start date",
+      "endDate": "End date"
     },
     "tips": {
       "acknowledged": "Acknowledged",
--- a/ui/main/src/assets/i18n/fr.json
+++ b/ui/main/src/assets/i18n/fr.json
@@ -171,9 +171,11 @@
     },
     "sort": {
       "title": "Triées par",
-      "severity": "Criticité puis date ",
-      "publish": "Date",
-      "unread": "Non lu puis date"
+      "severity": "Criticité puis date de réception",
+      "publish": "Date de réception",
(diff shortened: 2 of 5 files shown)


ui/main/src/assets/i18n/en.json
```diff
@@ -173,7 +173,9 @@
       "title": "Sort by",
       "severity": "Criticality then date",
       "publish": "Date",
-      "unread": "Unread then date"
+      "unread": "Unread then date",
+      "startDate": "Start date",
+      "endDate": "End date"
```
freddidierRTE, line 178: Replace : 
 "Date" -->  "Date of reception" 
 "Unread then" --> "Unread then date of reception"
 "Criticality then" --> "Criticality then date of reception"

ui/main/src/assets/i18n/fr.json
```diff
@@ -173,7 +173,9 @@
       "title": "Triées par",
       "severity": "Criticité puis date ",
       "publish": "Date",
-      "unread": "Non lu puis date"
+      "unread": "Non lu puis date",
+      "startDate": "Date de début",
```
freddidierRTE, line 177: "Date" --> "Date de réception"
"Non lu puis date" --> "Non lu  puis date de réception"
"Criticité puis date" --> "Criticité puis date date de réception"

ui/main/src/assets/i18n/nl.json
```diff
@@ -173,7 +173,9 @@
       "title": "Sorteer op",
       "severity": "Ernst en dan datum",
       "publish": "Datum",
-      "unread": "Ongelezen en dan datum"
+      "unread": "Ongelezen en dan datum",
+      "startDate": "Startdatum",
+      "endDate": "Einddatum"
```
freddidierRTE, line 178: Same comment as for other languages

freddidierRTE (changes requested):

Issue #3566, opened 2022-09-16
Add setting to automatically open the first card in list


Haigutus:
At the moment where the user opens UI no card is active. We would like to display to operator hallway the first card already open.

1. If operator opens the application, automatically the first card by default sorting will be displayed
2. If real-time time-frame is active, automatically after previous card has expired new first card will be open

This enables to set up operator fabric to large control screens, where no user input is required and allays the sorted list first entry card is displayed, without any user interaction to open it for the display.

freddidierRTE:
Interesting feature , i have already an easy implementation , i propose you to integrate it in next release schedule for next week .

It will be considered as an experimental feature (no documentation , no automatic test ...) 

For this first implementation , you will have to define in web-ui.json the user logins that is in this mode called "hallway mode"

```
  "settings": {
    "usersInHallwayMode":["operator1_fr"],
    "locale": "en",
    "dateTimeFormat": "HH:mm DD/MM/YYYY",
```

So you could test it , if it suits your needs and works well , we may deliver it as an official feature on a next release.

freddidierRTE:
@Haigutus : Does this feature suits your needs ?

freddidierRTE:
@Haigutus : the feature is now accessible via the settings of the user in release 4.5.0 , see migration guide : https://opfab.github.io/documentation/archives/4.5.0.RELEASE/docs/single_page_doc.html#_hallway_mode_setting

PR #3602, opened 2022-09-23
Add experimental hallway mode (#3566)

freddidierRTE:
- In release note :
  -  In chapter :  Features 
  -  Text : #3566 Add experimental hallway mod
  
To use it set users that are in this mode via  settings.userInHallwayMode  in web-ui.json


quinarygio:
When sending lot of cards using send6xN_TestCards.sh  the card detail correctly show the last received card, but in the feed some cards are marked as read, while others as unread.
image

We will see how to manage this after experimentation

quinarygio (approved):

PR #3690, opened 2022-10-10
Clean some code 

freddidierRTE:
Nothing in release note


ui/main/src/app/model/feed-filter.model.ts
```diff
@@ -8,61 +8,35 @@
...
-    /**
-     * apply the filter to the card, returns true if the card passes the filter, false otherwise
-     * @param card
-     */
-    applyFilter(card: LightCard): boolean {
-        if (this.active) {
-            return this.funktion(card, this.status);
-        }
-        return true;
+     static chainFilter(card: LightCard, next: Filter[]) {
```
quinarygio, line 16: Declare boolean return type

ui/main/src/app/model/feed-filter.model.ts
```diff
@@ -8,61 +8,35 @@
...
     }
 
     /**
      * Apply this filter to a card, then a chain of filter recursively.
      * The recursion stops when the card is filtered out
-     * @param card
-     * @param next
      */
-    chainFilter(card: LightCard, next: Filter[]) {
+     chainFilter(card: LightCard, next: Filter[]) {
```
quinarygio, line 24: Declare boolean return type

quinarygio (changes requested):

Issue #3638, opened 2022-09-27
In templates , with slow network, style is apply long after template processing


freddidierRTE:
Connect with operator1_fr
Set slow 3G network with chrome
Click on  card example "Message"
Close the card
Click again on card example .

The card appears and the style is applied later (text in blue) 

Solution : put style file in cache 

Do not use browser cache , use a specific cache like for template , cache should be updated at the same time as template cache 

It means css will have to  be injected  , example : https://stackoverflow.com/questions/15505225/inject-css-stylesheet-as-string-using-javascript

PR #3691, opened 2022-10-10
Solve bug : In templates , with slow network, style is apply long after template processing (#3638)

freddidierRTE:
Add a cache mechanism for css associated with templates

- In release note :
  -  In chapter :   Bugs 
  -  Text : 3628  Solve bug : In templates , with slow network, style is apply long after template processing

### DIFF
--- a/ui/main/src/app/store/effects/processes.effects.ts
+++ b/ui/main/src/app/store/effects/processes.effects.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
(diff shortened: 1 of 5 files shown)


ui/main/src/app/store/effects/processes.effects.ts
```diff
@@ -14,13 +14,15 @@ import {ProcessesActionTypes} from '@ofActions/processes.actions';
 import {catchError, debounce, map} from 'rxjs/operators';
 import {ProcessesService} from '@ofServices/processes.service';
 import {HandlebarsService} from 'app/modules/cards/services/handlebars.service';
+import {TemplateCssService} from '@ofServices/template-css.service';
```
quinarygio, line 17: Fix copyright

quinarygio (changes requested):

quinarygio (approved):

PR #3739, opened 2022-10-19
Make the output of waitForOpfabToStart.sh looks better

freddidierRTE:
Nothing in release note

### DIFF
--- a/bin/run_all.sh
+++ b/bin/run_all.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2018-2021, RTE (http://www.rte-france.com)
+# Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
 # See AUTHORS.txt
 # This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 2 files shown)


bin/run_all.sh
```diff
@@ -262,6 +262,6 @@ esac
 
 if [ "$waitForOpfabToStart" = true ]; then
     if [ $command = "start" ] || [ $command = "restart" ]; then 
-      ./bin/waitForOpfabToStart.sh
+      ./bin/waitForOpfabToStart.sh -w
```
quinarygio, line 265: Fix copyright

quinarygio (changes requested):

bin/waitForOpfabToStart.sh
```diff
@@ -42,29 +58,50 @@ case $key in
 esac
 done
 
+
+printWaitInProgress() {
+sleep 5 &
+pid=$!
+
+frames="/ | \\ -"
+while kill -0 $pid 2&>1 > /dev/null;
```
quinarygio, line 67: should be:  2>&1

quinarygio (changes requested):

Issue #3768, opened 2022-10-26
Search filter in the Card Feed


lhazenoot:
Add a search filter in the card feed that finds cards based on card title and summary (and maybe other contents of the cards). The search filter will be placed next to the feed-filter and feed-sort option. 

![Place to add search filter](https://user-images.githubusercontent.com/91895080/197967832-70711833-e829-4f75-8063-6947419123c0.png)

freddidierRTE:
Ok for me.  Add an option in web-ui.json to show/hide the search filter

PR #3790, opened 2022-11-04
Search filter in the Card Feed (#3768)

lhazenoot:

- In release note :
  -  In chapter :  Features
  -  Text : #3768

### DIFF
--- a/ui/main/src/app/services/lightcards/lightcards-feed-filter.service.ts
+++ b/ui/main/src/app/services/lightcards/lightcards-feed-filter.service.ts
@@ -17,13 +17,15 @@ import {SortService} from './sort.service';
 import {GroupedCardsService} from '@ofServices/grouped-cards.service';
 import {ConfigService} from '@ofServices/config.service';
 import {LogOption, OpfabLoggerService} from '@ofServices/logs/opfab-logger.service';
+import {SearchService} from '@ofServices/lightcards/search-service';
 
 @Injectable({
     providedIn: 'root'
 })
 export class LightCardsFeedFilterService {
-    private filteredAndSortedLightCards = new ReplaySubject(1);
+    private filteredAndSortedLightCards = new Subject();
     private filteredLightCards = new Subject();
+    private filteredAndSearchedLightCards = new ReplaySubject(1);
     private filteredLightCardsForTimeLine = new Subject();
     private onlyBusinessFilterForTimeLine = new Subject();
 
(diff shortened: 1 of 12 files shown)


ui/main/src/app/services/lightcards/lightcards-feed-filter.service.ts
```diff
@@ -96,6 +104,17 @@ export class LightCardsFeedFilterService {
             });
     }
 
+    private computeFilteredAndSearchedLightCards() {
+        combineLatest([this.searchService.getSearchChanges(), this.getFilteredLightCards()])
+            .pipe(map((results) => {
+                return this.searchService.searchLightCards(results[0], results[1]);
+            }))
+            .subscribe((lightCards) => {
+                this.logger.debug('Number of cards visible after searching : ' +  lightCards.length, LogOption.LOCAL_AND_REMOTE);
```
freddidierRTE, line 113: "Number of cards visible after **filtering and**  searching"

And remove the log just before to avoid to much log 
```
 this.logger.debug('Number of cards visible after filtering : ' +  lightCards.length ,LogOption.LOCAL_AND_REMOTE);
```

ui/main/src/app/services/lightcards/lightcards-feed-filter.service.ts
```diff
@@ -63,6 +67,10 @@ export class LightCardsFeedFilterService {
         return this.filteredLightCardsForTimeLine.asObservable();
     }
 
+    public getFilteredAndSearchedLightCards() : Observable<any> {
+        return this.filteredAndSearchedLightCards.asObservable();
+    }
+
     private computeFilteredLightCards() {
```
freddidierRTE, line 74: In ui/main/src/app/modules/feed/components/map/map.component.ts  line 65 getFilteredAndSortedLightCards shall be replace by   getFilteredAndSearchedLightCards

In ui/main/src/app/modules/feed/feed.component.ts line 68 getFilteredAndSortedLightCards shall be replace by getFilteredAndSearchedLightCards

So method public getFilteredAndSearchedLightCards() should not be neeeded anymore in this service

freddidierRTE:
Do not forget to udpate/create copyright text in files  (For existing ones , just add the line Copyright (c) 2022, Alliander (http://www.alliander.com) ) 

In your actual implementation , the timeline is not filtered by the serach field , is it the wanted behavior?

freddidierRTE (changes requested):

ui/main/src/app/services/lightcards/lightcards-feed-filter.service.ts
lhazenoot: If we do this than the sort function doesn't work anymore.

freddidierRTE: yes you are right , forgot my remark

freddidierRTE:
Could you rebase to solve the conflict ? 

In your actual implementation , the timeline is not filtered by the search field , is it the wanted behavior?

I will add an issue to add cypress tests for this feature

lhazenoot:
It not needed for us to filter the timeline, we just want to quick find the cards in the card list. I will rebase the branch right now.

Issue #3222, opened 2022-06-27
Grouped card have graphical bug in day mode


geppyz:
While using card grouping feature i noticed a graphical bug while useing the daytime theme. This bug is not present in the night theme.

[](url
<img width="459" alt="Schermafbeelding 2022-06-27 om 15 28 51" src="https://user-images.githubusercontent.com/18466477/175953673-bf84789d-c033-4e56-af08-5b43cb908139.png">
)

PR #3806, opened 2022-11-09
Grouped cards have graphical bug in day mode (#3222)

lhazenoot:

- In release note :
  -  In chapter :  Bugs
  -  Text : #3222


quinarygio (changes requested):
Please add/update copyright text in new/modified files (For existing ones , just add the line Copyright (c) 2022, Alliander (http://www.alliander.com/) )

Issue #3805, opened 2022-11-09
Improve display of multi-line card titles


lhazenoot:
Right now long card titles are broken in strange places (see photo for example) and that is not very readable, and also looks a bit odd. We would like to change the card title style so that words don't break up anymore, but are moved to a new line.

Improvement:
- Add a 'work-break' property to the card-title in the light-card.component.scss

How it looks like now:
![broken words](https://user-images.githubusercontent.com/91895080/200806078-7ac08369-5e58-4e19-9600-620c59d1e2b1.png)

What we want it looks like:
![Scherm­afbeelding 2022-11-09 om 11 26 53](https://user-images.githubusercontent.com/91895080/200806265-546e5590-a7a9-43de-a9ff-9925c59a21cf.png)

PR #3807, opened 2022-11-09
Improve display of multi-line card titles (#3805)

lhazenoot:

- In release note :
  -  In chapter :  Features
  -  Text : #3805


quinarygio (changes requested):
Not working whit a very long work in the title : 
image

Adding "overflow-wrap: anywhere;" style should fix the problem.

Issue #3826, opened 2022-11-14
Code injection in mutliselect component 


freddidierRTE:
HTML is interpreted in mutliselect component , see following example which lead to open a popup via code injection in entity name.

![image](https://user-images.githubusercontent.com/58251627/201673149-e63d5fc5-df2b-4959-85e1-1467b5197b55.png)

![image](https://user-images.githubusercontent.com/58251627/201672950-24f1fe06-2a14-497a-b559-ffa3a141b277.png)

In our use cases, all  the options lists came from trusted input so it is not a big issue,  but for template using the component there's not always  the guarantee the input is trusted. 

Solution sanitize option list in multiselect component

PR #3836, opened 2022-11-15
Solve bug : Code injection in mutliselect component (#3826)

freddidierRTE:
Nothing in release note (already present)


freddidierRTE:
PR Error

Issue #3848, opened 2022-11-17
Deeplink iFrame link bug


geppyz:
Some of the time a deeplink into a 'businessparty/iframe' link wont work. 
The link gets extra encoded and cannot be found by the iframedisplay component anymore. Also any url-parameters would be concatinated on the menu_entry_id and this would not be resolved.

Error in console would be:
"Error in business application redirection =  Error: No such menu entry."

PR #3849, opened 2022-11-17
Fix for encoded urls in iframe (#3848)

geppyz:
- In release note :
  -  In chapter : Bugs
  -  Text : #3848


ui/main/src/app/modules/businessconfigparty/iframedisplay.component.ts
```diff
@@ -68,6 +68,10 @@ export class IframeDisplayComponent implements OnInit, OnDestroy {
         });
     }
 
+    private removeEncodedParams(url: string): string {
+        return decodeURIComponent(url).split('?')[0];
+    }
```
freddidierRTE, line 73: This method is used only to remove encoded characters from the menu_entry_id , while do we need to split it ? 

I suggest to remove this method and add directly before : 
```
 .queryMenuEntryURL(this.menu_id, decodeURIComponent(this.menu_entry_id))
```

geppyz:
@freddidierRTE, I have not found a better way to do it yet. Could you merge it?

PR #3897, opened 2022-11-29
Logging: ag-grid table with filtering and paging on backend (#3793)

### LINKED ISSUE #3793: To filter the content of the Logging screen depending on the content
Dear all,

For our projects, it would be nice to have the possibility to update the table of the "Logging" screen depending on a free text filter.

An option would be to have the same searching option existing in the "Monitoring" screen.

Kind regards,

quinarygio:
Fix #3793 

- In release note :
  -  In chapter :  Features
  -  Text : #3793 : Logging:  add search filters on title and summary columns


freddidierRTE:
Is it possible to make a search ignoring uppercase/lowercase like in monitoring ? 
Add an index in titleTranslated and summaryTranslated in services/cards-publication/src/main/java/org/opfab/cards/publication/model/ArchivedCardPublicationData.java
The export feature shall take into account the filters 
Sometimes the filter input disappear if we type too slowly (it launch the request and remove the popup with the input field)

Issue #3773, opened 2022-10-27
Advanced  reminder feature using RFC 5545 recurrence structure


freddidierRTE:
Keep the actual reminder feature that will be deprecated 

Add a new one based on rfc5545 : https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10  and rrule lib : https://github.com/jakubroztocil/rrule

Add a new field in card named rrule  (at the card level not anymore in timespan as the actual one)

The field contains a json  representing the recurrence as define in the RFC (https://www.rfc-editor.org/rfc/rfc5545#section-3.8.5.3)

RFC string : FREQ=WEEKLY;COUNT=10;WKST=SU;BYDAY=TU,TH

In Json  
```
rrule : { 
  freq : "WEEKLY" 
  count: 10,
  wkst: ["SU"],
  byday: ["TU"]
}
```

This structure shall be compatible with rrule js lib 

Do not take into account timezone for this first implementation 

- [x] Modify back to include this structure

- Add this structure on the front  : 

   

- [x] Add a new service ui/main/src/app/services/rrule-reminder that will be an equivalent of ui/main/src/app/services/reminder . Store i n local storage as for the deprecated reminder but with another key (for example rruleReminderList)

   

- [x] Modify calendar to use this new reminder in // of the deprecated one 

   

- [x] Modify timeline to use this new reminder in // of the deprecated one  (in getAllCardsToDrawOnTheTimeLine() in ui/main/src/app/modules/feed/components/time-line/custom-timeline-chart/custom-timeline-chart.component.ts

   

- [x] Add a new task usercard using this new reminder , calling it Task Advanced . We can start with simple option and iterate to add more possibilities to be discuss 

   

- [x] Add in documentation

PR #3921, opened 2022-12-05
Advanced reminder feature using RFC 5545 recurrence structure (#3773)

vlo-rte:
I propose to add karate and cypress tests after the review of this PR, and in another issue.

- In release note :
  -  In chapter : Features
  -  Text : #3773 : Advanced reminder feature using RFC 5545 recurrence structure


freddidierRTE:
The publish date and startDate shall not be seen on timeline when recurrence is present as for classic task : 

image

The recurrence shal not start before startDate on timeline

image

When remind , the remind does not take into account the startDate and reminder before startDate

I guess solution is in ui/main/src/app/services/rrule-reminder/rrule-reminderUtils.ts

in 
```
export function getNextTimeForRepeating(card: Card, startingDate?: number): number {
    if (!startingDate) {
        startingDate = new Date().valueOf();
    }
```

When no start date it shall use startDate

PR #3939, opened 2022-12-14
Starting Refacto user service 

freddidierRTE:
Nothing in release note


services/users/src/main/java/org/opfab/users/services/CurrentUserWithPerimetersService.java
```diff
@@ -0,0 +1,93 @@
+package org.opfab.users.services;
```
quinarygio, line 1: Missing copyright

services/users/src/test/java/org/opfab/users/stubs/EntityRepositoryStub.java
```diff
@@ -0,0 +1,156 @@
+/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
```
quinarygio, line 1: Fix copyright : leave only 2022

services/users/src/test/java/org/opfab/users/stubs/UserServiceStub.java
```diff
@@ -0,0 +1,97 @@
+
+/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
```
quinarygio, line 2: Fix copyright : leave only 2022

quinarygio (changes requested):

quinarygio (approved):

PR #3940, opened 2022-12-14
Add READONLY opfab role (#3772)

### LINKED ISSUE #3772: Add a read only opfab role
When user is readonly : 
- he cannot respond to a card
- he cannot send cards 
- when he ack a card , it does not ack for his entities 

Add an example with an account named operator1_crisisRoom 
Add cypress/karate test 
Add documentation

quinarygio:
Fix #3772

- In release note :
  -  In chapter :  Features
  -  Text : #3772 : Add READONLY opfab role

### DIFF
--- a/client/users/src/main/java/org/opfab/users/model/OpfabRolesEnum.java
+++ b/client/users/src/main/java/org/opfab/users/model/OpfabRolesEnum.java
@@ -25,8 +25,9 @@ import com.fasterxml.jackson.annotation.JsonValue;
  */
 public enum OpfabRolesEnum {
     ADMIN("ADMIN"),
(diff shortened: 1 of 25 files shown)


freddidierRTE (changes requested):

freddidierRTE:
When user is read only and the card is ack at the entity level , it should be consider as ack at the user level for him otherwise he cannot ack the card 

To reproduce : 
Connect with operator1_fr , send a card message to everybody 
Connect with operator1_crisisRoom , try to ack the card --> the fact that the user ack the card is not visible for him , the card is still consider as unacknowledged

PR #3958, opened 2022-12-21
Fix card detail not visible in archives mode when using admin mode with role VIEW_ALL_ARCHIVED_CARDS (#3951)

### LINKED ISSUE #3951: Card detail not visible in archives mode when using admin mode with  role VIEW_ALL_ARCHIVED_CARDS
To reproduce 

Start a clean version of opfab
Load conf
Send the 6 tests cards
Connect as operator2_fr
Go in archives 
Make a search 
Open the message card  , you see the detail
Make a search in admin mode 
Open the Network contingency cards , you see an endless spinner instead of the card detail 

image

quinarygio:
Fix #3951

- In release note :
  -  In chapter :  Bugs 
  -  Text : #3951 : Fix card detail not visible in archives mode when using admin mode with role VIEW_ALL_ARCHIVED_CARDS


freddidierRTE:
Nothing in release note as the bug was not present in previous release

PR #3965, opened 2022-12-22
External app : changing the type of the returned object (#3915)

### LINKED ISSUE #3915: External app : changing the type of the returned object
In ExternalAppClientImpl.java, in the function callExternalHttpApplication(...), we call restTemplate.postForObject(externalRecipientUrl, requestBody, CardPublicationData.class); 
The type for the returned object should not be CardPublication.

Currently it probably works because empty body is converted to null object but it would be better to change this.
And add more logs please, to see what happens in case of problems.

quinarygio:
Fix #3915

- In release note :
  -  In chapter :  Tasks
  -  Text : #3915 : External app : changing the type of the returned object


freddidierRTE:
The change was to be made in  callExternalHttpApplication and not in  notifyExternalHttpApplication

But let the return type as modify (void) for notifyExternalHttpApplication and modify as well callExternalHttpApplication

Issue #3957, opened 2022-12-21
Migrate to full calendar V6


freddidierRTE:
Does not build with renovate : https://github.com/opfab/operatorfabric-core/pull/3956

PR #3977, opened 2022-12-27
Migrate to full calendar V6 (#3957)

vlo-rte:
Nothing in release notes.


vlo-rte:
The branch to merge is FE-3957-bis.

Issue #4006, opened 2023-01-10
Change opfabRole to permissions


freddidierRTE:
Move opfabRoles from the user object to the group object and rename it in "permissions"

To divide the work : 

- [x] 1) Modif the group object in the back to introduce the permissions field as an Array of Permissions . A permission is an Enum like for opfabRoles . Add in unit test  

- [x] 2) Add/Modify Karate test concerning group management in user service  --> Giovanni  
- [x] 3) Add the field in the array groups in admin screen  --> Clement  
- [x] 4) Modify cypress test for the array in admin screen --> Clement 
- [x] 5) Add the field in edit group modal --> Clement 
- [x] 6) Add/modify cypress test for edit group modal  --> Clement 
- [x] 7) Introduce the permission in userWithPerimters Object as a new field "permissions" , this new field should be build based on user groups , add/modify corresponding  unit tests --> Giovanni 
- [x] 8) Use the new Permission for viewing all archives   (Back + Front) --> Giovanni 
- [x] 9) Use the new permission for readonly (Back + Front) --> Giovanni 
- [x] 10) Use the new permission for admin (Back + Front) --> giovanni 
- [x] 11) Modify user service to use permission for security controls --> giovanni 
- [x] 12) Add the documentation and  remove documentation for opfabRoles --> Clement
- [x] 13) Remove admin screen for opfabRoles --> Clement 
- [x] 14) Remove usage of opfabRoles for menu 
- [x] 15) Remove opfabRoles form back 

Data default conf  : 

Group supervisor -->  Permission READONLY 
Group admin --> Permission ADMIN 
Group manager --> Permission VIEW_ALL_ARCHIVED_CARDS
Operator2 became member of group manager

PR #4029, opened 2023-01-13
Added the edit button and the cypress tests for permissions in the admin group table #4006

ClementBouvierN:
Signed-off-by: ClementBouvierN <clement.bouvierneveu@rte-france.com>


src/test/cypress/cypress/integration/Admin.spec.js
```diff
@@ -21,6 +21,7 @@ describe('AdmininstrationPages', () => {
     before('Set up configuration', function () {
         script.loadTestConf();
     });
+    /*
```
freddidierRTE, line 24: This test should not be commented

freddidierRTE (changes requested):

PR #4041, opened 2023-01-17
Refacto ui : externalize http calls to get processes information

freddidierRTE:
Nothing in release note 

The config service has been moved from app/business/config to app/business/services , this causes a lot of minor modifications due to change in import 

All http calls in ui/main/src/app/services/processes.service.ts are externalize via interfaces  : 

-  ui/main/src/app/business/server/process.server.ts 
-  ui/main/src/app/business/server/config.server.ts

The server interfaces are now in ui/main/src/app/business/server/
The implementation are now in ui/main/src/app/server

The handlebars tests implementation in  ui/main/src/app/modules/card/services/handlebars.service.spec.ts has been simplified  (the list of tests remain the same) 

The ui/main/src/app/services/processes.service.spec.ts has been simplified thanks to the new implementation.


ui/main/src/app/modules/usercard/selectStateForm/usercard-select-state-form.component.ts
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2023, RTE (http://www.rte-france.com)
```
quinarygio, line 1: should be: 2022-2023

quinarygio (changes requested):

vlo-rte (approved):

PR #4054, opened 2023-01-18
Refacto UI : move some business services to the new business repository

freddidierRTE:
Move date-time-formatter and gui service in business/services repository


ui/main/src/app/modules/logging/components/cell-renderers/time-cell-renderer.component.ts
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2023, RTE (http://www.rte-france.com)
```
vlo-rte, line 1: Should be 2022-2023

vlo-rte (changes requested):

vlo-rte (approved):

PR #4060, opened 2023-01-20
Refacto UI : modify http error management in server classes

freddidierRTE:
Add a ServerResponse class to manage error in business services without having to be coupled with angular HttpClient 

Nothing in release note

### DIFF
--- a/ui/main/src/app/business/services/processes.service.spec.ts
+++ b/ui/main/src/app/business/services/processes.service.spec.ts
@@ -7,19 +7,18 @@
  * This file is part of the OperatorFabric project.
  */
 
-
 import {ProcessesService} from './processes.service';
 import {getRandomAlphanumericValue} from '@tests/helpers';
 import {Process} from '@ofModel/processes.model';
 import {ConfigServerMock} from '@tests/mocks/configServer.mock';
 import {ProcessServerMock} from '@tests/mocks/processServer.mock';
+import {ServerResponse, ServerResponseStatus} from '../server/serverResponse';
 
 describe('Processes Services', () => {
-
     let processesService: ProcessesService;
     let processServerMock: ProcessServerMock;
     let configServerMock: ConfigServerMock;
-    
+
     beforeEach(() => {
         processServerMock = new ProcessServerMock();
         configServerMock = new ConfigServerMock();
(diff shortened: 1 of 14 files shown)


ui/main/src/app/business/services/processes.service.spec.ts
```diff
@@ -81,13 +80,22 @@ describe('Processes Services', () => {
         });
     });
 
-    describe('#queryProcess', () => {
-        const processDefinition = new Process('testPublisher', '0', 'businessconfig.label');
-        it('should load businessconfig from remote server', () => {
-            processServerMock.setProcessDefinition(processDefinition);
+    describe('Query a process', () => {
+        it('GIVEN a process exist WHEN query the process THEN process is return', () => {
```
vlo-rte, line 84: What do you think of 'GIVEN an existing process WHEN query the process THEN process is returned' ?

ui/main/src/app/business/services/processes.service.spec.ts
```diff
@@ -81,13 +80,22 @@ describe('Processes Services', () => {
...
+            const processDefinition = new Process('testPublisher', '0', 'businessconfig.label');
+            const serverResponse = new ServerResponse<Process>(processDefinition, null, null);
+            processServerMock.setResponseForProcessDefinition(serverResponse);
             processesService
                 .queryProcess('testPublisher', '0')
                 .subscribe((result) => expect(result).toEqual(processDefinition));
         });
+
+        it('GIVEN a process does not exist WHEN query the process THEN null return', () => {
```
vlo-rte, line 93: What do you think of 'GIVEN a non-existing process WHEN query the process THEN null is returned' ?

vlo-rte (changes requested):

ui/main/src/app/server/angular.server..ts
```diff
@@ -0,0 +1,28 @@
...
+    protected processHttpResponse<dataType>(httpResponse: Observable<dataType>): Observable<ServerResponse<dataType>> {
+        return httpResponse.pipe(
+            map((data) => {
+                return new ServerResponse(data, ServerResponseStatus.OK, '');
+            }),
+            catchError((error) => {
+                let serverStatus = ServerResponseStatus.UNKNOWN_ERROR;
+                if (error.status === 404) serverStatus = ServerResponseStatus.NOT_FOUND;
+                else if (error.status === 403) serverStatus = ServerResponseStatus.FORBIDDEN;
+
```
quinarygio, line 23: I wonder if it could be useful to log the error status when not  404 or 403

freddidierRTE: the log is already present in the network panel so i do not think we have to repeat it here

vlo-rte (approved):

PR #4068, opened 2023-01-23
Refacto event stream management

freddidierRTE:
Isolate the code that manage the realtime connection to opfab using an interface ui/main/src/app/business/server/opfabEventStream.server.ts implemented in ui/main/src/app/server/angularOpfabEventStream.server.ts

This class manage technically the connection (heartbeat management, recovery ..) 

The business management of the events is shared between 
- A new  class ui/main/src/app/business/services/opfabEventStream.service.ts  for the general events 
- The existing class ui/main/src/app/services/lightcards/lightcards-store.service.ts for the events regarding cards 

The lightcards-store does not access directly to the event stream server but uses the event stream **service** (opfabEventStream.service.ts)

The resulting card.service.ts is now mainly a gateway to the back for cards operations , it will be refactor to be coherent with new UI architecture in an another issue

The ui/main/src/app/services/lightcards/lightcards-store.service.ts becomes rather big so it may be cut in an another issue

Nothing in release note

### DIFF
--- a/ui/main/src/app/app.component.ts
+++ b/ui/main/src/app/app.component.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2018-2023, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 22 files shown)


vlo-rte (approved):

quinarygio (approved):

quinarygio (commented):
There is a conflicting file

Issue #4069, opened 2023-01-23
Missing zero-padding for Task Advanced card


vlo-rte:
Without zero padding, if the time is for example 07:08, it is "interpreted" as 7:8 and so it does not work with the display of the card and then in the timepicker when you edit the card.

PR #4070, opened 2023-01-23
Missing zero-padding for Task Advanced card (#4069)

vlo-rte:
Nothing in release notes because this bug is not present in the actual release yet.

### DIFF
--- a/ui/main/src/app/modules/card/services/handlebars.service.ts
+++ b/ui/main/src/app/modules/card/services/handlebars.service.ts
@@ -45,6 +45,7 @@ export class HandlebarsService {
         HandlebarsService.registerMergeArrays();
         HandlebarsService.registerConditionalAttribute();
         HandlebarsService.registerReplace();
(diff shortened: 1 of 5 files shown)


ui/main/src/app/modules/card/services/handlebars.service.ts
```diff
@@ -238,6 +239,12 @@ export class HandlebarsService {
         });
     }
 
+    private static registerPadStart() {
+        Handlebars.registerHelper('padStart', function (nbCharactersToAdd, characterForPadding, stringToPad) {
+            return String(stringToPad).padStart(nbCharactersToAdd, characterForPadding);
```
quinarygio, line 244: Maybe it would be better to rename the 'nbCharactersToAdd' param to someting like 'targetLength'.

vlo-rte: Yes you're totally right :+1:

quinarygio (approved):

PR #4109, opened 2023-01-27
Refacto UI : remove use of ngrx for token synchro

freddidierRTE:
Nothing in release note

### DIFF
--- a/ui/main/src/app/modules/core/application-loading/application-loading.component.ts
+++ b/ui/main/src/app/modules/core/application-loading/application-loading.component.ts
@@ -176,10 +176,18 @@ export class ApplicationLoadingComponent implements OnInit {
         this.store.select(selectIdentifier).subscribe((identifier) => {
             if (identifier) {
                 this.logger.info(`User ${identifier} logged`);
+                this.synchronizeUserTokenWithOpfabUserDatabase();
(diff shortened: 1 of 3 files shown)


ui/main/src/app/modules/core/application-loading/application-loading.component.ts
```diff
@@ -176,13 +176,21 @@ export class ApplicationLoadingComponent implements OnInit {
...
                 this.userLogin = identifier;
                 this.loadSettings();
             }
         });
     }
 
+    private synchronizeUserTokenWithOpfabUserDatabase() {
+        this.userService.synchronizeWithToken().subscribe({
+            next: () =>  this.logger.info("Synchronization of user token with user database done"),
+            error: () => this.logger.warn("Impossible to synchronize user token with user database")
```
quinarygio, line 190: Is it correct to just log a warn ? Previously in case of error in synchronization the user would have been logged out by calling "authService.clearAuthenticationInformation()"

freddidierRTE: yes that was the previous behavior . But i prefer continuing the loading process if not succeed as opfab can work without the synchro

quinarygio (approved):

Issue #4126, opened 2023-01-31
Show popup with list of cards on the geomap


lhazenoot:
Right now, If there are multiple cards with the same location, only one card is shown as a popup when clicking on the dot. We would like to see all cards with the same location. Therefore, we want the popup to show a list of all cards with the same location, just like the popup in the 'timeline calender'.  

Example for the new popup:
![multiple popup cards map (1)](https://user-images.githubusercontent.com/91895080/215791680-dc7ce473-f4ce-4c8a-9e10-4bbfff6cd482.png)

- Also need to update the cypress test for Geomap.

PR #4160, opened 2023-02-03
Show popup with list of cards on the geomap (#4126)

lhazenoot:

- In release note :
  -  In chapter :  Features
  -  Text : #4126


quinarygio (commented):
When there are many cards with the same location the popup is not fully visible and card list is truncated: 
image

Maybe it would be better to have a limit on the number cards shown in the popup and show 3 dots "..." Oherwise there should be a scrollbar

quinarygio (approved):

PR #4161, opened 2023-02-06
Play sound and show alert when a new card is not visible in feed (#4134)

### LINKED ISSUE #4134: Activate sound even if feed filters hide the new card
When user has set sounds in settings 
Sound should arise even if cards is hide on the feed due to feed/timeline filters
When a new sounded card arrive and is not visible on the feed , a specific alert message should be display in red as in the following screen shot 

image

The message :

in french  :  "Vous avez reçu une carte masquée par les filtres que vous avez activés (Timeline ou flux de cartes)"

in english  : "You have received a card hidden by the filters you have activated (Timeline or card feed)"

The message does not disappear automatically like classical alert messages , user shall click on the cross to remove it 

The message shall never be hidden by another message (in particular reload message) 

The red color shall be the same as the red color for alarm cards 

Use the actual alert service with a new message level : BUSINESS 

Add a cypress test

quinarygio:
Fix #4134

- In release note :
  -  In chapter :  Features 
  -  Text : #4134 : Play sound and show alert when a new card is not visible in feed

### DIFF
--- a/src/test/cypress/cypress/integration/SoundNotification.spec.js
+++ b/src/test/cypress/cypress/integration/SoundNotification.spec.js
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2021-2023, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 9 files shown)


src/test/cypress/cypress/integration/SoundNotification.spec.js
```diff
@@ -155,8 +157,70 @@ describe('Sound notification test', function () {
...
+        sendCardWithSeverityAlarm();
+        sound.checkNumberOfEmittedSoundIs(1);
+        feed.checkNumberOfDisplayedCardsIs(0);
+
+        cy.get('#div-detail-msg').find('span').eq(0).contains('You have received a card hidden by the filters you have activated (Timeline or card feed)');
+        // Check that alert message css class
+        cy.get('of-alert').find('.opfab-alert-business').should('exist');
+        cy.wait(6000);
+        // Check that alert message is not closed automatically after 6 seconds
+        cy.get('#div-detail-msg').should('exist');
```
freddidierRTE, line 211: simulate time instead of waiting 6sec using `cy.clock` and  `cy.tick` (See example inside the current test)

Issue #4183, opened 2023-02-07
refactor ui/main/src/app/services/user.service.ts to conform to the new UI Software architecture


freddidierRTE:
create  ui/main/src/app/business/server/user.server.ts to interact with the back instead of using httpClient inside the service

remove method queryAllEntities form the user.service.ts

PR #4187, opened 2023-02-07
Refacto user service into business code #4183

ClementBouvierN:
Signed-off-by: ClementBouvierN <clement.bouvierneveu@rte-france.com>

Nothing in release note

### DIFF
--- a/ui/main/src/app/business/services/user.service.ts
+++ b/ui/main/src/app/business/services/user.service.ts
@@ -1,5 +1,4 @@
-/* Copyright (c) 2020, RTEi (http://www.rte-international.com)
- * Copyright (c) 2021-2023, RTE (http://www.rte-france.com)
+/* Copyright (c) 2023, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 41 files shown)


ui/main/src/app/business/services/user.service.ts
```diff
@@ -8,87 +7,117 @@
...
     }
 
     synchronizeWithToken(): Observable<User> {
-        return this.httpClient.post<User>(`${this.userUrl}/users/synchronizeWithToken`, null);
+        return this.userServer.synchronizeWithToken().pipe(
+            map((userResponse) => {
+                if (userResponse.status === ServerResponseStatus.OK){
+                    return userResponse.data;
+                } else {
+                    this.handleServerResponseError(userResponse);
```
freddidierRTE, line 67: In this case , do not use handleServerResponse but the logger : 
logger.error("Impossible to synchronize token")

ui/main/src/app/business/services/user.service.ts
```diff
@@ -8,87 +7,117 @@
...
     }
 
     currentUserWithPerimeters(): Observable<UserWithPerimeters> {
-        return this.httpClient.get<UserWithPerimeters>(`${this.userUrl}/CurrentUserWithPerimeters`);
+        return this.userServer.currentUserWithPerimeters().pipe(
+            map((userResponse) => {
+                if (userResponse.status === ServerResponseStatus.OK){
+                    return userResponse.data;
+                } else {
+                    this.handleServerResponseError(userResponse);
```
freddidierRTE, line 80: In this case , do not use handleServerResponse but the logger : 
logger.error("Impossible to load user perimeter")

ui/main/src/app/business/services/user.service.ts
```diff
@@ -166,10 +195,28 @@ export class UserService extends CrudService {
...
     }
 
     willNewSubscriptionDisconnectAnExistingSubscription(): Observable<boolean> {
-        return this.httpClient.get<boolean>(`${this.willNewSubscriptionDisconnectAnExistingSubscriptionUrl}`);
+        return this.userServer.willNewSubscriptionDisconnectAnExistingSubscription().pipe(
+            map((userResponse) => {
+                if (userResponse.status === ServerResponseStatus.OK){
+                    return userResponse.data;
+                } else {
+                    this.handleServerResponseError(userResponse);
```
freddidierRTE, line 216: In this case , do not use handleServerResponse but the logger : 
logger.error("Impossible to check if new connection will disconnect existing subscription ")

freddidierRTE (changes requested):

Issue #4083, opened 2023-01-25
Refactor code regarding lightcards  to conform to the new UI Software architecture


freddidierRTE:
move ui/main/src/app/services/lightcards in ui/main/src/app/business/services/lightcards

Remove use of angular in ui/main/src/app/services/lightcards/sort.service.spec.ts

PR #4208, opened 2023-02-10
Refactor code regarding lightcards to conform to the new UI Software architecture (#4083)

freddidierRTE:
Nothing in release note


ui/main/src/tests/mocks/remote-logger.service-angular.mock.ts
```diff
@@ -0,0 +1,27 @@
+/* Copyright (c) 2022-2023, RTE (http://www.rte-france.com)
```
quinarygio, line 1: Copyright should be only for 2023

quinarygio (changes requested):

PR #4210, opened 2023-02-10
Show deleted card message in detail view (#4163)

### LINKED ISSUE #4163: Delete card message in detail view
When a detail card is on the screen and the ui receives a delete of the card , a message should be one the card saying the card is deleted instead of removing the card from the screen 

Message shall override the card , keeping the card visible lighter in background

quinarygio:
Fix #4163 

- In release note :
  -  In chapter :  Features 
  -  Text : #4163 : Show deleted card message in detail view

### DIFF
--- a/ui/main/src/app/business/services/card/selectedCard.service.ts
+++ b/ui/main/src/app/business/services/card/selectedCard.service.ts
@@ -9,7 +9,7 @@
 
 import {Injectable} from '@angular/core';
 import {Card} from '@ofModel/card.model';
-import {Observable, ReplaySubject} from 'rxjs';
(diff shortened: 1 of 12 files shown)


ui/main/src/app/business/services/card/selectedCard.service.ts
```diff
@@ -60,6 +62,15 @@ export class SelectedCardService {
     public getSelectCard(): Observable<SelectedCard> {
         return this.selectedCardWithChildrenChange.asObservable();
     }
+
+    public setSelectedCardDeleted(cardId: string) {
```
freddidierRTE, line 66: i would name it setCardDeleted as it is not always the selected card that is deleted

Issue #4251, opened 2023-02-21
Scrollable card list in popup of timeline


lhazenoot:
We often have a lot of cards at the same time, but we can only see a few cards in the popup of the timeline. We would like to see all the cards in the popup. Therefore, we would like to have a scrollbar in the popup, so we can see all the cards, or if not possible, then at least 100 cards in the popup. 

Some example: 
![Scherm­afbeelding 2023-02-21 om 13 00 21](https://user-images.githubusercontent.com/91895080/220341492-8d068395-cb8e-49af-927d-b4aa22f44b5a.png)

PR #4263, opened 2023-02-23
Scrollable card list in popup of timeline (#4251)

lhazenoot:
- In release note :
  -  In chapter :  Features
  -  Text : #4251


freddidierRTE:
Add alliander copyright to the files you modify

PR #4266, opened 2023-02-24
Fix business menu cypress test

quinarygio:
Nothing in release note


vlo-rte (approved):

vlo-rte:
I merge because it solves the pb on businessmenu cypress test, the build is failed because of another cypress test.

Issue #4264, opened 2023-02-23
Entities recipients for information : update archive card detail footer


vlo-rte:
We have to take into account the recipients and the recipients for information.

PR #4267, opened 2023-02-24
Entities recipients for information : update archive card detail footer (#4264)

vlo-rte:
Nothing in release notes.


freddidierRTE:
Add a line on the footer for entityRecipientsForInformation if exist

freddidierRTE:
Add a line on the footer for entityRecipientsForInformation if exist

PR #4272, opened 2023-02-27
Sorting strings doesn't take emojis into account anymore #4262

ClementBouvierN:
Signed-off-by: ClementBouvierN [clement.bouvierneveu@rte-france.com](mailto:clement.bouvierneveu@rte-france.com)

Nothing in release note


freddidierRTE:
to be correct as well in archives / logging / screen : 
image

PR #4282, opened 2023-03-03
sonarcloud code smells : Replace this instanceof check and cast with 'instanceof String string'

vlo-rte:
Nothing in release notes.

### DIFF
--- a/services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/IpAddressAuthorizationManager.java
+++ b/services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/IpAddressAuthorizationManager.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, RTE (http://www.rte-france.com)
+/* Copyright (c) 2021-2023, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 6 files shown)


services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/IpAddressAuthorizationManager.java
```diff
@@ -62,8 +62,8 @@ public Mono<AuthorizationDecision> check(Mono<Authentication> authenticationMono
 
     private User getAuthenticatedUser(Authentication authentication) {
         User userData = null;
-        if (authentication.getPrincipal()  instanceof CurrentUserWithPerimeters){
-            CurrentUserWithPerimeters current = (CurrentUserWithPerimeters) authentication.getPrincipal();
+        if (authentication.getPrincipal()  instanceof CurrentUserWithPerimeters currentUserWithPerimeters){
+            CurrentUserWithPerimeters current = currentUserWithPerimeters;
             userData = current.getUserData();
```
quinarygio, line 67: Is it necessary to define 'current' variable? Could we just use "userData = currentUserWithPerimeters .getUserData()" ?

quinarygio (changes requested):
Added comments about not needed variable definitions

services/cards-consultation/src/main/java/org/opfab/cards/consultation/configuration/webflux/IpAddressAuthorizationManager.java
vlo-rte: You're right. Done

tools/spring/spring-oauth2-utilities/src/main/java/org/opfab/springtools/configuration/oauth/WebSecurityChecks.java
```diff
@@ -56,11 +56,11 @@ public boolean checkUserIpAddress(Authentication authentication) {
...
         //but is String type if there is no authentication (jira : OC-655)
-        if (authentication.getPrincipal()  instanceof String) {
-            user = (String) authentication.getPrincipal();
+        if (authentication.getPrincipal()  instanceof String principal) {
+            user = principal;
         }
-        else if (authentication.getPrincipal()  instanceof CurrentUserWithPerimeters){
-            CurrentUserWithPerimeters current = (CurrentUserWithPerimeters) authentication.getPrincipal();
+        else if (authentication.getPrincipal()  instanceof CurrentUserWithPerimeters currentUserWithPerimeters){
+            CurrentUserWithPerimeters current = currentUserWithPerimeters;
```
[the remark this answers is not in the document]
vlo-rte: You're right. Done

quinarygio (approved):

PR #4306, opened 2023-03-09
Fix task duration not used in calendar (#4230)

### LINKED ISSUE #4230: Task duration is not used in calendar
Either correct or do not define task duration

quinarygio:
Fix #4230 

- In release note :
  -  In chapter :  Bugs 
  -  Text : #4230 : Fix task duration not used in calendar


vlo-rte (changes requested):
I think it does not work with "Task advanced example", the card is visible in calendar in month view but not in week and day view. Could you check please ?

vlo-rte (approved):

vlo-rte:
> I think it does not work with "Task advanced example", the card is visible in calendar in month view but not in week and day view. Could you check please ?

It's a mistake. It works. So I merge.

PR #4307, opened 2023-03-10
Fix browser stuck for user tryin to connect without authorized IP address

### LINKED ISSUE #4245: Browser stuck for user tryin to connect without authorized IP address
For example if you set for operator1_fr, the field "authorized IP address" with 1.2.3.4, then it will be impossible for operator1_fr to connect to opfab (wanted behaviour), but it will be impossible to connect with another user because the browser is "stuck" with the connection of operator1_fr.

quinarygio:
Fix #4245

- In release note :
  -  In chapter :  Bugs
  -  Text : #4245 : Fix browser stuck for user tryin to connect without authorized IP address

### DIFF
--- a/ui/main/src/app/modules/core/application-loading/application-loading.component.ts
+++ b/ui/main/src/app/modules/core/application-loading/application-loading.component.ts
@@ -197,12 +197,17 @@ export class ApplicationLoadingComponent implements OnInit {
         this.settingsService.getUserSettings().subscribe({
             next: (response) => {
                 if (response.status === ServerResponseStatus.OK){
-                    this.logger.info(new Date().toISOString() + `Settings loaded` + response.data);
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/core/application-loading/application-loading.component.ts
```diff
@@ -202,6 +202,11 @@ export class ApplicationLoadingComponent implements OnInit {
                     this.checkIfAccountIsAlreadyUsed();
                 } else {
                     if (response.status === ServerResponseStatus.NOT_FOUND) console.log(new Date().toISOString(), 'No settings for user');
+                    else if (response.status === ServerResponseStatus.FORBIDDEN) {
+                        console.log(new Date().toISOString(), 'Access forbidden when loading settings');
```
freddidierRTE, line 206: use logger

PR #4347, opened 2023-03-22
Moved the bundle storage one folder deeper FE-4320

ClementBouvierN:
Nothing in release note


freddidierRTE:
Add migration documentation

Issue #4344, opened 2023-03-21
Scrollable list in the geomap popup


lhazenoot:
Right now, the max number of cards displayed in the geomap popup is five. We would like to see all cards in the popup. Therefore, we would like to add a scroll bar in the popup, so we can display all the cards. Just like the timeline popup. 

Example:
![Geomap popup scroll](https://user-images.githubusercontent.com/91895080/226601298-5246a527-0a01-4ddb-a2a3-dfb57714401a.png)

PR #4357, opened 2023-03-27
Scrollable list in the geomap popup (#4344)

lhazenoot:
- In release note :
  -  In chapter :  Features
  -  Text : #4344

### DIFF
--- a/ui/main/src/app/modules/feed/components/map/map.component.scss
+++ b/ui/main/src/app/modules/feed/components/map/map.component.scss
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022, Alliander (http://www.alliander.com)
+/* Copyright (c) 2022-2023, Alliander (http://www.alliander.com)
 /* Copyright (c) 2018-2022, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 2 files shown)


ui/main/src/app/modules/feed/components/map/map.component.scss
```diff
@@ -21,6 +21,9 @@
     border-radius: 10px;
```
quinarygio, line 21: Could you just fix the copyright header:  Copyright (c) 2022-**2023**, Alliander

quinarygio (changes requested):

quinarygio:
In firefox cards title goes on newline : 
image

lhazenoot:
> In firefox cards title goes on newline : image

@quinarygio Hmm, that is weird! At mine, it just looks normal in firefox. Don't know if this is really a problem? Or can we just leave it for what it is?

Firefox geomap popup

quinarygio:
> > In firefox cards title goes on newline : image
> 
> @quinarygio Hmm, that is weird! At mine, it just looks normal in firefox. Don't know if this is really a problem? Or can we just leave it for what it is?
> 
> Firefox geomap popup

The issue occurs only when there is the scrollbar


> > In firefox cards title goes on newline : image
> 
> @quinarygio Hmm, that is weird! At mine, it just looks normal in firefox. Don't know if this is really a problem? Or can we just leave it for what it is?
> 
> Firefox geomap popup

The problem occurs only when there is the scrollbar

quinarygio:
> > @quinarygio Even with the scrollbar I don't get the issue. Also in Edge and Chrome it works normal. So, I have no clue what it could be. Do you want me to find a solution for it?
> 
> Lets wait for a second check by @freddidierRTE to see if it could be reproduced or if it's just me..

I will merge it as nobody else reproduced the issue

PR #4455, opened 2023-04-20
Add service cards-external-diffusion (#4396)

### LINKED ISSUE #4396: Add a new back services : cards-external-diffusion
This service shall be able  to send cards via mail when user is not connected and a card has been published  for him.

It shall send a mail per card the user would have been received if connected

The user mail shall be set in settings 
The user shall set in settings if he wants to receive mail or not 

**Mail format :** 
 
  Title : Prefix  - card title - card summary - start date - end date 
  --> Prefix is an instance parameter 
  
  Content : Fix text + http link to the card 
  --> Fix text is an instance parameter

 Create a new service called cards-external-diffusion 

Do not use swagger to generate code (for controller or data object)
Isolate business code from spring framework 

Find a way to make integration test (use a test mail server ? )

quinarygio:
Fix #4396 

- In release note :
  -  In chapter :  Features
  -  Text : #4396 : Add service cards-external-diffusion (#4396)


node-services/cards-external-diffusion/src/domain/server-side/opfabInterface.ts
```diff
@@ -0,0 +1,315 @@
...
+    public loadUsersData() {
+        this.fetchAllUsers();
+        this.userPerimeters.clear();
+        this.userSettings.clear();
+    }
+
+    public clearUserCache(login: string) {
+        this.userSettings.delete(login);
+        this.userPerimeters.delete(login);
+        this.fetchUser(login);
```
freddidierRTE, line 211: Maybe we do not need to fetch here , it will be fetch after if missing

This avoid pb when a lot of updates came from rabbit 

To discuss

freddidierRTE:
Add a chapter mail notification in https://opfab.github.io/documentation/current/reference_doc/  , a few lines to explain the feature
Update in documentation : https://opfab.github.io/documentation/current/deployment/#_service_port_table
Update in documentation : https://opfab.github.io/documentation/current/deployment/#_configuration
Add documentation for developer in https://opfab.github.io/documentation/current/dev_env/#_development_tools : use of mailhog and how to connect to it 


Possible to store the service in services instead of node-services ? 

Add sonar analysis (config in build.gradle)

freddidierRTE (changes requested):

node-services/cards-external-diffusion/src/cardsExternalDiffusion.ts
```diff
@@ -0,0 +1,159 @@
...
+    .setOpfabGetUsersUrl(config.get('opfab.usersUrl'))
+    .setOpfabGetUsersConnectedUrl(config.get('opfab.connectedUsersUrl'))
+    .setOpfabGetCardsUrl(config.get('opfab.cardsUrl'))
+    .setOpfabGetTokenUrl(config.get('opfab.getTokenUrl'))
+    .setAuthenticationService(authenticationService)
+    .setLogger(logger);
+
+opfabInterface.loadUsersData();
+
+const listener = new EventBusListener();
```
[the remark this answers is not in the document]
quinarygio: What about moving the EventbusListener into CardsExternalDiffusionService  instead ? Because it has to be started and stopped as the service and I see OpfabInterface just as an interface to other services.

node-services/cards-external-diffusion/src/domain/server-side/opfabInterface.ts
quinarygio: In fact 'fecthUser()' it will not be called if the user is missing from cache (normally the control takes the users list from the cache),  so it has to be called when removing a user from the cache.

PR #4459, opened 2023-04-21
Adjust log for card publication service

freddidierRTE:
- In release note :
  -  In chapter :  Tasks
  -  Text :  #4459 Adjust log for card publication service.


services/cards-publication/src/main/java/org/opfab/cards/publication/controllers/CustomExceptionHandler.java
```diff
@@ -17,6 +17,7 @@
 import org.springframework.http.HttpStatus;
 import org.springframework.http.HttpStatusCode;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.AccessDeniedException;
```
quinarygio, line 20: Fix Copyright

quinarygio (changes requested):

PR #4469, opened 2023-05-03
Sorted the names of the processes in the dashboard #4427

ClementBouvierN:
- In release note :
  -  In chapter :  Features
  -  Text : #4427 Sorted the names of the processes in the dashboard


ui/main/src/app/business/view/dashboard/dashboard.view.ts
```diff
@@ -64,7 +64,10 @@ export class Dashboard {
             processContent.name = process.name;
             statesContent.sort((obj1, obj2) => Utilities.compareObj(obj1.name, obj2.name));
             processContent.states = statesContent;
+
             if (processContent.states.length > 0) this.dashboardPage.processes.push(processContent);
+            
+            this.dashboardPage.processes.sort((obj1, obj2) => Utilities.compareObj(obj1.name, obj2.name));
```
freddidierRTE, line 70: Should be sort only once , when the foreach is done (and not each time we add a process)  ?

PR #4471, opened 2023-05-05
Load ui-menu.json via businessConfig service (#4441)

### LINKED ISSUE #4441: Load ui-menu.json via businessConfig service
To be able to dynamically reload menu , instead of having menu.json as a file in config (config/docker/ui-config/ui-menu.json) , access it via an endpoint from businessconfig (like for example processGroup.json) 

So add an endpoint to POST the file and a one to GET the file 
When the file is modify , it should send a signal to rabbit so the user could reload it (similar mechanism as other file like processGroup) 

Do not control the format of the menu file that is POST , consider it as json file for now ( like what is done when uploading business data #4387 ) 

Add documentation in migration file 
Modify documentation 

Add the load of the menu in the src/test/resources/loadTestConf.sh

vlo-rte:
- In release note :
  -  In chapter :  Features 
  -  Text : #4441 : Load ui-menu.json via businessConfig service


freddidierRTE:
issue cancelled

Issue #4417, opened 2023-04-14
Send heardbeat from ui to back


freddidierRTE:
In certain circumstances , the network is cut between the browser and the cards-consultation services but the connection does not close on the back , this situation has been observe in production for a duration of 1 hour . So the user is considered connected where as it is not the case 

Solution : 
Send every X minutes an HTTP heartbeat request from the UI to the back  . The endpoint shall cards/CardSubscriptionHeartBeat

If the back did not receive the heartbeat for Y minutes , it closes the connection 

By default X = 60s
and Y = 150 s 

Implementation : 

- Add a last connection date in services/cards-consultation/src/main/java/org/opfab/cards/consultation/services/CardSubscription.java
- Update it when receiving hearbeat 
- Add a new Thread in services/cards-consultation/src/main/java/org/opfab/cards/consultation/services/CardSubscriptionService.java

PR #4487, opened 2023-05-11
Send heartbeat from front to back #4417

ClementBouvierN:
- In release note :
  -  In chapter :  Tasks
  -  Text : #4417 The UI is now sending a heartbeat to the back


services/cards-consultation/src/main/java/org/opfab/cards/consultation/services/CardSubscriptionService.java
```diff
@@ -93,6 +101,34 @@ private void sendHeartbeatMessageInAllSubscriptions()
...
+                    Thread.currentThread().interrupt(); // Cf sonar java:S2142 "InterruptedException" should not be ignored
+                    return; 
+                }
+            log.debug("Comparing last heartbeat received");
+            cache.keySet().forEach(key -> {
+                CardSubscription sub = cache.get(key); 
+                if (sub != null) // subscription can be null if it has been evicted during the process of going throw the keys
+                {
+                    log.debug("Time since last heart beat : {}", Instant.now().getEpochSecond() - sub.getHeartbeatReceptionDate().getEpochSecond());
+                    if (Instant.now().getEpochSecond() - sub.getHeartbeatReceptionDate().getEpochSecond()  > 150) {
```
freddidierRTE, line 127: Use a configuration value instead of 150  

Name : heardbeatDelayInSecondsToConsiderUserDisconnected
Default value : 100

services/cards-consultation/src/main/java/org/opfab/cards/consultation/services/CardSubscriptionService.java
```diff
@@ -93,6 +101,34 @@ private void sendHeartbeatMessageInAllSubscriptions()
...
+                        this.evictSubscription(sub.getId());
```
freddidierRTE, line 128: add a log : 
```
 log.info("User with subscription id  {}  has not send heardbeat for more than {} seconds  disconnect it",key heardbeatDelayInSecondsToConsiderUserDisconnected) ; 
```

freddidierRTE (changes requested):

Issue #4454, opened 2023-04-20
Add scripts to change log level on services 


freddidierRTE:
add in src/test/resources/logs

Add a getLogs.sh to get log level : param server:port , login , pwd ( similar to other scripts) 
```
curl  http://opfab-publication.rte-france.com:2102/actuator/loggers/ROOT -H "Authorization: Bearer $token" -H "Content-type:application/json"
```
Add a setLogs.sh to set log level  , param : server:port , loglevel  , login , pwd  

Example of request to set logs  : 
```
curl -i -X POST http://opfab-publication.rte-france.com:2102/actuator/loggers/ROOT -H "Authorization: Bearer $token" -H 'Content-Type: application/json' -d '{"configuredLevel": "INFO"}
```

Add in documentation https://opfab.github.io/documentation/current/deployment/#_logging_administration  , the link to the scripts

PR #4519, opened 2023-05-24
Add scripts to change log level on services (#4454)

quinarygio:
- In release note :
  -  In chapter :  Tasks
  -  Text : #4454:  Add scripts to change log level on services

### DIFF
--- a/src/test/resources/logs/setLogs.sh
+++ b/src/test/resources/logs/setLogs.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Copyright (c) 2023, RTE (http://www.rte-france.com)
+# See AUTHORS.txt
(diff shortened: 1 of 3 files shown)


src/test/resources/logs/setLogs.sh
```diff
@@ -0,0 +1,31 @@
...
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+url=$4
+if [ -z $url ] 
+then
+	url="http://localhost"
+fi
+if [ -z $3 ]
+then
+    echo "Usage : setLogs host:port level user opfab_url"
```
vlo-rte, line 19: Same remark than previously, what do you think of adding an example of usage ?

vlo-rte (changes requested):

vlo-rte (approved):

PR #4523, opened 2023-05-26
Remove userRecipientsIds, groupRecipientsIds and entityRecipientsIds from CardOperation class

vlo-rte:


freddidierRTE:
Conflicts to resolve

PR #4548, opened 2023-05-30
Fix ack filter default value and reset (#4543)

### LINKED ISSUE #4543: Ack/Not ack filter : filter still activated even if you click "reset filters"
Steps to reproduce : 

set only ack cards
reset filter => only not ack are visible
then you go in archives
then you go back , the filter is really changed to ack only visible (and the filter icon is red again)

It also happens if you reload instead of going on the archives

quinarygio:
Fix #4543 

- In release note :
  -  In chapter :  Bugs 
  -  Text :#4543 :  Fix ack filter default value and reset


src/docs/asciidoc/deployment/configuration/web-ui_configuration.adoc
```diff
@@ -73,7 +73,7 @@ a|card time display mode in the feed. Values :
  - LTTD: displays card with lttd date;
  - NONE: nothing displayed.
 |feed.defaultSorting|"unread"|no|Default card sorting option in feed page, possible values are : "unread", "date", "severity", "startDate", "endDate".
-|feed.defaultAcknowledgmentFilter|"notack"|no|Default card acknowlegment filtering option in feed page, possible values are : "notack", "ack", "all", "none".
+|feed.defaultAcknowledgmentFilter|"notack"|no|Default card acknowlegment filtering option in feed page, possible values are : "notack", "ack", "all".
```
vlo-rte, line 76: notack => notAck

Issue #4608, opened 2023-06-14
computeShowDetailCardHeader() has a faulty condition


ClementBouvierN:
in the function, there is this test:
!this.cardState.showDetailCardHeader || this.cardState.showDetailCardHeader === true

which always return true

PR #4609, opened 2023-06-14
Fixed the showDetailCardHeader test #4608

ClementBouvierN:
- In release note :
  -  In chapter :  Bugs 
  -  Text : #4608 Fixed the showDetailCardHeader condition

### DIFF
--- a/src/test/resources/bundles/messageOrQuestionExample/config.json
+++ b/src/test/resources/bundles/messageOrQuestionExample/config.json
@@ -28,7 +28,6 @@
     "questionState": {
       "name": "Question ",
       "description": "Question ",
-      "showDetailCardHeader": false,
       "color": "#0070C0",
       "userCard" : {
 		"template" : "usercard_question",
@@ -49,7 +48,6 @@
     "confirmationState": {
       "name": "Confirmation",
       "description": "Please confirm",
-      "showDetailCardHeader": false,
       "color": "#0070C0",
       "userCard" : {
 		"template" : "usercard_confirmation",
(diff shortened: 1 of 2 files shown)


src/test/resources/bundles/defaultProcess_V1/config.json
```diff
@@ -85,6 +85,7 @@
         "state": "responseState"
       },
       "templateName": "question",
+      "showDetailCardHeader" : true,
```
vlo-rte, line 88: The default value is true so it is not needed, it "weighs down" unnecessarily the file.

src/test/resources/bundles/defaultProcess_V1/config.json
```diff
@@ -109,6 +110,7 @@
         "style"
       ],
       "acknowledgmentAllowed": "Always",
+      "showDetailCardHeader" : true,
```
vlo-rte, line 113: The default value is true so it is not needed, it "weighs down" unnecessarily the file.

src/test/resources/bundles/messageOrQuestionExample/config.json
```diff
@@ -28,7 +28,7 @@
     "questionState": {
       "name": "Question ",
       "description": "Question ",
-      "showDetailCardHeader": false,
+      "showDetailCardHeader": true,
       "color": "#0070C0",
```
vlo-rte, line 32: Same remark than previously. 
I propose to have only one 
"showDetailCardHeader" : true
in opfab, for automatic tests, to be sure there is no regression in the future.

src/test/resources/bundles/messageOrQuestionExample/config.json
```diff
@@ -49,7 +49,7 @@
     "confirmationState": {
       "name": "Confirmation",
       "description": "Please confirm",
-      "showDetailCardHeader": false,
+      "showDetailCardHeader": true,
```
vlo-rte, line 52: Same remark than previously. 
I propose to have only one 
"showDetailCardHeader" : true
in opfab, for automatic tests, to be sure there is no regression in the future.

vlo-rte (changes requested):

vlo-rte (approved):

Issue #4607, opened 2023-06-14
Add reload page button to the application-loading and connection-lost screen


lhazenoot:
Lately, it has been common for our OpFab to hang on the application loading screen, and we have to manually refresh the page to fix it (most of the time). Many of our users are unaware that they have to refresh manually, and they often come to us with complaints. Therefore, we would like to add a 'reload page' button when the application has not loaded after 10 seconds.

Secondly, we would also like to add the reload page button to the connection lost screen, as currently we also have to manually refresh this one.

freddidierRTE:
Ok for the loading 

But for the connection lost is shall reconnect automatically , if it does not reconnect it means the back is not fully available and reload will just end up with a spinner and the application "application is loading" message . 

If it does not reconnect automatically it means there is a bug in the mechanism and so we need to investiguate , do you have a way to reproduce it ?

vlo-rte:
Resolved by PR https://github.com/opfab/operatorfabric-core/pull/4610

PR #4610, opened 2023-06-14
Add reload page button to the loading screens (#4607)

lhazenoot:
- In release note :
  -  In chapter :  Features
  -  Text : #4607


freddidierRTE:
Could you use a translated value instead of a direct english text for the button ? 

Could you add some vertical space between the text "Application is loading ..." and the button (i guess something like 15 or 20 px shall be ok) 

image

PR #4646, opened 2023-06-26
Transfer card reminder management from ui to backend (#4586)

### LINKED ISSUE #4586: Transfert reminder management front code  to back
Today , the management of reminder is done in the front : a regular task check if a card has to be remind if yes it unack/unread the card 

A problem arise when reminding in a context of user connected at the same entity : each user independently remind.

You end up with the following situation : 
- "user1" of "entity1" has a card "card1" with a reminder at 10am , 5 minutes before 10am opfab unread/unack the card . User 1 ack the card after reading the remind. 
- user2 connect to entity1 at 10am  , as he connect to entiy one he see "card1" but as reminder is set , the front code unread/unack the card . But it is not wanted as "user1" already ack this remind  -> the front user2 do not know that the front of user1 already ack the reminder.

To solve the problem , we need to implement the reminder logic in the back . Each time a card is to be remind , the back send an update of the card so the ack and the read will be reseted an each user will see again the card . So coming back to the previous case , when user1 will ack the new card , it will automatically be ack for user 2 who will not see the remind. 

Consider that :
- The code of the reminder is based on rrule and  it seems that we have no real easy lib to process rrule in java
-  Existing typescript ui code to deal with reminder already implement an important part of the logic
-  We have already back services based on ts code (via node.js)

--> the code to deal with that shall be based on node.js (like external diffusion component) , it shall be a new component named : cards-reminder

The idea is to catch new card via rabbit and set a reminder in mongo for the new date to reminder , ie the next date to update the card  (comparing to the actual UI code, mongo act like the local storage of the browser)

WARNING : What happens in case the service is started after the publication service or is stopped when a card is send , does rabbit keep the card in the queue and the service get the card at startup ? 

The solution shall keep the deprecated way of managing reminder (i.e existing front code) 

To reset read/ack , a new endpoint shall be set in card-publication :  resetReadAndAcks . This endpoint will get the card from mongo and send an updated version of the card , publishdate shall be set to current time (this is necessary for the sound notification to work, otherwise it consider it is an old card).

To communicate between service , use a new account named "opfab_internal_account" 

The resetReadAndAcks shall only be accessible to opfab_internal_account

The endpoint shall be tested via Karate

quinarygio:
Fix #4586

- In release note :
  -  In chapter :  Features 
  -  Text : #4586 : Transfer card reminder management from ui to backend


freddidierRTE (changes requested):

node-services/cards-reminder/src/domain/client-side/configDTO.ts
```diff
@@ -0,0 +1,13 @@
...
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+export default class ConfigDTO {
+    public checkPeriodInSeconds = 0;
```
freddidierRTE, line 12: 5 seconds

node-services/cards-reminder/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
+}
```
freddidierRTE, line 3: to remove or add to .gitignore

freddidierRTE (changes requested):

Issue #4674, opened 2023-06-30
Solve sonar issue Do not use "Error" to declare a variable


freddidierRTE:
https://sonarcloud.io/project/issues?languages=ts&resolved=false&rules=typescript%3AS2137&id=org.lfenergy.operatorfabric%3Aoperatorfabric-core&open=AYcoafKwK07Qf5OT5PkQ

and  log the error

PR #4680, opened 2023-07-03
Solve sonar issue Do not use Error to declare a variable (#4674)

HubertDuchatelle:
nothing in release note


ui/main/src/app/authentication/auth-handler.ts
```diff
@@ -94,7 +94,7 @@ export abstract class AuthHandler {
     private decodeToken(token: string): any {
         try {
             return jwt_decode(token);
-        } catch (Error) {
+        } catch (err) {
```
freddidierRTE, line 97: Ajouter une ligne de log : this.logger.error(" Error in token decoding " + err);

PR #4683, opened 2023-07-03
Add a retry to solve falky test with business menu

freddidierRTE:
Nothing in release note


src/test/cypress/cypress/integration/BusinessMenu.spec.js
```diff
@@ -7,12 +7,11 @@
  * This file is part of the OperatorFabric project.
  */
```
quinarygio, line 8: Fix Copyright

quinarygio (changes requested):

Issue #4620, opened 2023-06-16
Add a ui screen to manage buisness data file 


freddidierRTE:
Add a way to : 
- download files
- upload files 
- delete files 

test cypress

PR #4690, opened 2023-07-05
Added Business Data Management Table #4620

ClementBouvierN:
- In release note :
  -  In chapter :  Features 
  -  Text : #4620 Added Business Data Management Table in the administration tab


freddidierRTE:
Need to be rebased 

Bug : 

Load test conf
Log as admin 
Download the example conf file 
Modify it
Update it via admin screen
Disconnect from admin
Connect as operator1_fr 
Create new card IT Incident the list of services is empty 

In fact the update delete the "service" file and create a new "service.json" file

freddidierRTE (changes requested):

PR #4711, opened 2023-07-11
Added a business data cypress test #4705

ClementBouvierN:
Nothing in release notes


freddidierRTE:
Check the download (see other download example) 
Check upload if possible

Issue #4641, opened 2023-06-21
Add a comment field in user object 


freddidierRTE:
This field shall be visible on the admin interface for user  when you edit user . It shall not be visible on the table

PR #4720, opened 2023-07-12
Added a comment attribute to users #4641

ClementBouvierN:
- In release note :
  -  In chapter :  Features 
  -  Text : #4641 When adding or editing a user, it's possible to see, add or edit a comment about the user

### DIFF
--- a/services/users/src/main/java/org/opfab/users/model/UserData.java
+++ b/services/users/src/main/java/org/opfab/users/model/UserData.java
@@ -37,6 +37,7 @@ public class UserData implements User {
     private String login;
     private String firstName;
     private String lastName;
+    private String comment;
 
     @JsonIgnore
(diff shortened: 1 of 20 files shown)


services/users/src/main/java/org/opfab/users/model/UserData.java
```diff
@@ -54,11 +55,17 @@ public UserData(User user){
...
         this.groupSet = new HashSet<>(user.getGroups());
         this.authorizedIPAddresses = new HashSet<>(user.getAuthorizedIPAddresses());
     }
 
+    @Override
+    public void setComment(String comment) {
+        this.comment = comment;
+    }
+
     public void addGroup(String group){
```
freddidierRTE, line 69: to see if removable

PR #4725, opened 2023-07-13
Do not consider character case when checking publisher is equal to lo…

HubertDuchatelle:
- In release note :
  -  In chapter :  Features
  -  Text : #4669 No longer consider case when checking login is equal to publisher / representative in CardPermissionControlService

### DIFF
--- a/services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
+++ b/services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
@@ -691,6 +691,29 @@ class CardProcessServiceShould {
         }
 
         @Test
+        void GIVEN_a_card_WHEN_card_is_send_with_a_login_case_different_than_publisher_THEN_card_is_accepted() {
+                User user = new User();
+                user.setLogin("DummyUser");
+                CurrentUserWithPerimeters sameUser = new CurrentUserWithPerimeters();
+                sameUser.setUserData(user);
+
+                ComputedPerimeter cp = new ComputedPerimeter();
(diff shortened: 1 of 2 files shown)


services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
```diff
@@ -734,6 +757,32 @@ void GIVEN_a_card_with_representative_dummyUser_WHEN_wrongUser_send_the_card_THE
...
+                cp.setState("state1");
+                cp.setRights(RightsEnum.RECEIVEANDWRITE);
+                List<ComputedPerimeter> list = new ArrayList<>();
+                list.add(cp);
+                sameUser.setComputedPerimeters(list);
+
+                CardPublicationData card = generateOneCard("IGNORED_PUBLISHER");
+                card.setPublisherType(PublisherTypeEnum.EXTERNAL);
+                card.setRepresentativeType(PublisherTypeEnum.EXTERNAL);
+                card.setRepresentative(currentUserWithPerimeters.getUserData().getLogin());
```
freddidierRTE, line 779: Should be for example card.setRepresentative("DUMMYUSER"), in your test login has the same case as representative

freddidierRTE (changes requested):

services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
HubertDuchatelle: resolved

Issue #4752, opened 2023-07-18
Add a message-or-question-list build-in template 


freddidierRTE:
Add a build-in template that provide to the user a list of predefined message or question 
The list of message/question came from a config file loaded as business data

the tag shall be \<opfab-message-or-question-list-usercard\>

PR #4831, opened 2023-07-27
Added built-in template with message selection #4752

ClementBouvierN:
In release note :
 In chapter : Features
   Text :  #4752 Added message selection from a JSON file as a built-in template


ui/main/src/app/business/buildInTemplates/message-or-question-list/usercard/message-or-question-listUserCardTemplateView.ts
```diff
@@ -0,0 +1,88 @@
...
+            summary: {key : "message_or_question_list.summary"},
+            startDate: new Date().valueOf(),
+            data: {
+                title: title,
+                message: message,
+                question: question
+            },
+            severity: severity
+        };
+
```
freddidierRTE, line 54: add field entitiesAllowedToRespond with the list of recipients when it is a question otherwise set value to ""

freddidierRTE (changes requested):

freddidierRTE:
Implement the edition/copy mode 

Add in documentation ,in https://opfab.github.io/documentation/archives/SNAPSHOT/docs/single_page_doc.html#_build_in_templates

Add a cypress test : send a card with question with a user and respond with another user

Issue #4368, opened 2023-03-30
add in archives view/export new field 


freddidierRTE:
Remove business period

Add
- entity recipient
- publisher 
- processus

PR #4850, opened 2023-08-02
add in archives view/export new field (#4368)

HubertDuchatelle:
- In release note :
  -  In chapter :  Features
  -  Text : #4368 In archives view / export : removed field business period, added fields publisher, entity recipients and process


vlo-rte:
Please can you add space between name of columns, in this example, "publication date" and "emitter" are too near 
image

vlo-rte:
Can you make an update to not have words broken ? In this example, we have 
"ELECTRICITY CONSUMPTION FORECAST" which is broken and also "Result of the maintenance is <OK>" which is also broken
image

vlo-rte (changes requested):
Please see my remarks.

vlo-rte:
In archives.spec.js, can you test also the "process column"  please ?

vlo-rte (changes requested):
In archives.spec.js, can you test also the "process column" please ?

HubertDuchatelle:
To discuss with Frédéric:
- words break and columns width

Issue #4768, opened 2023-07-19
Archives/Logging : ExpressionChangedAfterItHasBeenCheckedError in ng serve mode (#4768)


freddidierRTE:
Error in console when navigate to archives or logging screen in mode dev (port 4200)

![image](https://github.com/opfab/operatorfabric-core/assets/58251627/44e0a146-79a1-4a68-8ed9-bf47bc37dd42)

HubertDuchatelle:
For the archives/Logging screen, the issue occurs when setDateFilterBounds() is called in ngAfterViewInit(), this changes the value of publishMinDate from null to an object, which causes an error because of the binding in the template line 92 [minDate]="publishMinDate".

freddidierRTE:
Not reproduce , we use now native browser date picker so it may have sold the problem

PR #4863, opened 2023-08-16
Archives/Logging : ExpressionChangedAfterItHasBeenCheckedError in ng serve mode (#4768)

vlo-rte:
Nothing in release notes.


vlo-rte:
I close the PR because I realize that instead of dealing with the problem, I deleted it. The dateFilterBounds in the datetime components don't work anymore with my solution....

PR #4868, opened 2023-08-22
library xlsx not published anymore on npmjs

vlo-rte:
Why this PR : 
https://stackoverflow.com/questions/76101101/xlsx-package-vulnerabilities-found-but-there-is-no-newer-package
https://git.sheetjs.com/sheetjs/sheetjs/issues/2667
https://cdn.sheetjs.com/

Nothing in release notes.


freddidierRTE:
Use a precise version instead of latest

vlo-rte:
> Use a precise version instead of latest

Done.

Issue #4733, opened 2023-07-17
Add a script to migrate card to new recurrence structure


freddidierRTE:
add a js script to launch on mongoDB to migrate card with old recurrence structure (using recurrence field)  to new structure (rrule)

PR #4910, opened 2023-08-30
Add a script to migrate card to new recurrence structure (#4733)

vlo-rte:
Nothing in release notes.


freddidierRTE:
take into account the case with no month in recurence 
log the modification done

PR #4911, opened 2023-08-30
Add the possibility to set the publisher of the card via getUserResponse (#4794)

### LINKED ISSUE #4794: Add the possibility to set the publisher of the card via getUserResponse
In case user can respond to a card on behaf of more than one entity it should be possible to set via template code the entity to use

To do that the getUserResponse method of templates could return the publisher : 

Example : 

` opfab.currentCard.registerFunctionToGetUserResponse(() => {
                var confirm = document.getElementById('resp_confirm').value;
                const message = document.getElementById('resp_message').value;
                const responseCardData = { 
                        confirm: confirm,
                        message: message
                    };
                return {
                    valid: true,
                    publisher : "ENTITY1"
                    responseCardData: responseCardData
                };

            });`

quinarygio:
Fix #4794

- In release note :
  -  In chapter :  Features
  -  Text : #4794 :  Add the possibility to set the publisher of the card via getUserResponse

### DIFF
--- a/src/test/cypress/cypress/integration/ResponseCard.spec.js
+++ b/src/test/cypress/cypress/integration/ResponseCard.spec.js
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2021-2023, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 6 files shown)


src/test/cypress/cypress/integration/ResponseCard.spec.js
```diff
@@ -606,4 +606,57 @@ describe('Response card tests', function () {
...
+        cy.get('#opfab-card-details-entities-choice-selector').should('exist');
+        cy.get('#opfab-card-details-entities-choice-selector').find('.vscomp-option-text').should('have.length', 4);
+        
+
+        cy.get('#opfab-card-details-entity-choice-btn-cancel').click();
+
+        cy.get('#resp_message').invoke('val', 'Cypress test for response publisher');
+        card.sendResponse();
+
+        //Response was sent withiout showing popup for the entities choice
```
freddidierRTE, line 651: typo **without**

freddidierRTE:
Add the new field in documentation : https://opfab.github.io/documentation/archives/SNAPSHOT/docs/single_page_doc.html#_design_the_question_form_in_the_template

PR #4956, opened 2023-09-05
Solve Filesystem path, filename, or URI manipulation   (Ref audit : CID:29156)

freddidierRTE:


services/businessconfig/src/main/java/org/opfab/businessconfig/services/ProcessesService.java
```diff
@@ -687,4 +680,15 @@ public void deleteAllBusinessData() throws IOException {
         File dataDirectory = new File(resourcePath.toString());
         FileUtils.cleanDirectory(dataDirectory);
     }
+
+    private void checkInputDoesNotContainsForbiddenCharacters(String inputName,String inputValue) throws ApiErrorException {
```
vlo-rte, line 684: Please correct the indentation in all the function.

vlo-rte, line 684: checkInputDoesNotContainsForbiddenCharacters => checkInputDoesNotContainForbiddenCharacters ?

vlo-rte (changes requested):

services/businessconfig/src/main/java/org/opfab/businessconfig/services/ProcessesService.java
freddidierRTE: Done

vlo-rte (approved):

PR #4987, opened 2023-09-11
Solve potential null pointer exception (Ref audit : CID:29174)

freddidierRTE:
Nothing in release note


quinarygio (changes requested):
Should add also RTE Copyright ?

freddidierRTE:
> Should add also RTE Copyright ?

done

PR #5006, opened 2023-09-12
Fix mail sending for cards sent to parent entity (#4962)

### LINKED ISSUE #4962: When using parent card , mail is not send
To reproduce : 
- configure mail for operator1_fr on state message
- disconnect from operator_1
- connect with operator2_fr , send a message to "French Control Center" 
- wait more than one minute , no mail received 

if you send  a message to Control Center FR North , the mail is received

quinarygio:
Fix #4962 

Nothing in release notes

### DIFF
--- a/services/users/src/main/java/org/opfab/users/services/CurrentUserWithPerimetersService.java
+++ b/services/users/src/main/java/org/opfab/users/services/CurrentUserWithPerimetersService.java
@@ -30,6 +30,16 @@ public class CurrentUserWithPerimetersService {
         this.entityRepository = entityRepository;
     }
 
+    public CurrentUserWithPerimeters fetchUserWithPerimeters(String login) {
+        CurrentUserWithPerimeters currentUserWithPerimetersData = new CurrentUserWithPerimetersData();
+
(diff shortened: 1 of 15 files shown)


services/users/src/main/java/org/opfab/users/services/CurrentUserWithPerimetersService.java
```diff
@@ -30,6 +30,20 @@ public CurrentUserWithPerimetersService(UsersService usersService, UserSettingsS
...
+
+        if (userResp.isPresent()) {
+            User user = userResp.get();
+            currentUserWithPerimetersData.setUserData(user);
+            handleUserSettings(user, currentUserWithPerimetersData);
+            handleGroups(user, currentUserWithPerimetersData);
+            handleEntities(user);
+        }
+        return currentUserWithPerimetersData;
+    }
```
freddidierRTE, line 45: to avoid duplicated code  , get the user object and if it exists call fetchCurrentUserWithPerimeters(user) 
Add a unit test

freddidierRTE (changes requested):

PR #5031, opened 2023-09-15
Cards-external-diffusion should not send mail when card has been read by user (#5014)

### LINKED ISSUE #5014: Should not send mail when card has been read by user
Even if user has read card , it received a mail 

This is due to the fact that the value that card.userReads never exist in : 

```
    private isCardUnreadForUser(card: any, user: any): boolean {
        return (
            card.publishDate < Date.now() - 1000 * this.secondsAfterPublicationToConsiderCardAsNotRead &&
            !card.userReads?.includes(user.login)
        );
    }
```

In fact it exists in the database but not when you get the card via API 

So modify the API to add it

quinarygio:
Fix #5014

Nothing in release notes

### DIFF
--- a/src/test/api/karate/cards/cardsUserRead.feature
+++ b/src/test/api/karate/cards/cardsUserRead.feature
@@ -94,6 +94,8 @@ Feature: CardsUserRead
     And match response.card.hasBeenRead == true
     And match response.card.uid == uid
 
+
+
 #get card with user operator2_fr and check hasBeenRead is set to false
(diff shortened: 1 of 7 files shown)


src/test/api/karate/cards/cardsUserRead.feature
```diff
@@ -115,9 +117,38 @@ Feature: CardsUserRead
     And header Authorization = 'Bearer ' + authToken
     When method get
     Then status 200
+    And print response
```
freddidierRTE, line 120: Remove this unecessary print

PR #5051, opened 2023-09-19
Modify the web-ui container to run unprivileged (#4199)

### LINKED ISSUE #4199: Modify the web-ui container to run unprivileged
It would be nice to modify or add an additional docker container for the web-ui that runs as an unprivileged user. 
This would increase security and compatibility with certain cloud-environments.

quinarygio:
Fix #4199

- In release note :
  -  In chapter :  Tasks
  -  Text : #4199 : Modify the web-ui container to run unprivileged


freddidierRTE:
too complex for production / do not do it now

PR #5077, opened 2023-09-21
Avoid losing users update in case of stop/start of services (#5076)

### LINKED ISSUE #5076: Avoid losing users update in case of stop/start of services
The listener in node services shall always be started to avoid loosing events in case of start/stop of the regular checks

quinarygio:
Fix #5076 

Nothing in release note

### DIFF
--- a/node-services/cards-external-diffusion/src/cardsExternalDiffusion.ts
+++ b/node-services/cards-external-diffusion/src/cardsExternalDiffusion.ts
@@ -83,7 +83,6 @@ app.get('/start', (req, res) => {
         if (!isAdmin) 
             res.status(403).send();
         else {
-            opfabServicesInterface.startListener();
             cardsExternalDiffusionService.start();
             res.send('Start service');
         }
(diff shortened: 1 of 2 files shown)


node-services/cards-external-diffusion/src/cardsExternalDiffusion.ts
```diff
@@ -132,9 +130,10 @@ app.listen(adminPort, () => {
...
 });
 
-logger.info('Application started');
+opfabServicesInterface.startListener();
 
 if (activeOnStartUp) {
-    opfabServicesInterface.startListener();
     cardsExternalDiffusionService.start();
+
+logger.info('Application started');
```
freddidierRTE, line 138: Log shall be written even if not active on startup

node-services/cards-reminder/src/cardsReminder.ts
```diff
@@ -124,10 +124,12 @@ app.listen(adminPort, () => {
...
 });
 
-logger.info('Application started');
+opfabServicesInterface.startListener();
 
 if (activeOnStartUp) {
     cardsReminderService.start();
-    opfabServicesInterface.startListener();
+
+logger.info('Application started');
```
freddidierRTE, line 132: Log shall be written even if not active on startup

freddidierRTE (changes requested):

PR #5154, opened 2023-09-29
Fix Automatic full screen mode for card detail in feed when filters visible (#5101)

### LINKED ISSUE #5101: Automatic full screen mode for card  detail in feed does not work well when fitlers visible
We should not end up with the following screenshot with the detail card too small . In this particular case card should be in full screen 

Capture

quinarygio:
Fix #5101

- In release note :
  -  In chapter :  Bugs 
  -  Text : #5101 : Fix Automatic full screen mode for card detail in feed when filters visible

### DIFF
--- a/ui/main/src/app/modules/card/components/card-body/card-body.component.ts
+++ b/ui/main/src/app/modules/card/components/card-body/card-body.component.ts
@@ -342,8 +342,14 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
         return this.card.lttd != null && this.card.lttd - new Date().getTime() <= 0;
     }
 
-    public isSmallscreen() {
-        return window.innerWidth < 1000;
+    public isThereEnoughSpaceToShowCard() {
+        const domElement = document.getElementsByTagName('of-card-body');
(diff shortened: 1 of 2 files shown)


ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -343,7 +343,13 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
     }
 
     public isSmallscreen() {
-        return window.innerWidth < 1000;
+        const domElement = document.getElementsByTagName('of-card-body');
+        const width =  domElement.item(0).getBoundingClientRect().width;
```
freddidierRTE, line 347: i suggest to name it cardWidth instead of width

ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -343,7 +343,13 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
     }
 
     public isSmallscreen() {
```
freddidierRTE, line 345: I suggest to name it isThereEnoughSpaceToShowCard

Issue #5108, opened 2023-09-26
User card preview  - user connected does not take into account access right and notification configuration filter


freddidierRTE:
If displayConnectionCirclesInPreview is set to true in web-ui.json , we see on the preview a green circle if an entity is connected 

But the fact that a user is connected to an entity does not necessary mean he will receive the card as it depends on his access rights and his notification filter .

So we shall see the green circle only if the connected user has the good perimeter and even the good configuration in notification filter.

To do that we may call an endpoint in cards-consultation service that give the information of the entity that will effectively receive the card

PR #5313, opened 2023-11-02
Preview of users connected takes notification configuration into account #5108

ClementBouvierN:
- In release note :
  -  In chapter :  Features 
  -  Text : #5108  Preview of users connected takes notification configuration into account

### DIFF
--- a/services/cards-publication/src/main/modeling/swagger.yaml
+++ b/services/cards-publication/src/main/modeling/swagger.yaml
@@ -997,6 +997,27 @@ paths:
           description: Authentication required
         '403':
           description: Forbidden - User doesn't have any group
+  /connectedRecipientsPreview:
+    post:
+      summary: get the connected Entity among the recipients in card preview
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: card
+          in: body
+          schema:
+              $ref: '#/definitions/Card'
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              type: string
+        '400':
+          description: Bad request
   '/cards/{id}':
     parameters:
       - in: path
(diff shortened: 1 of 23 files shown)


services/cards-consultation/src/main/java/org/opfab/cards/consultation/services/ConnectedEntityPreviewService.java
```diff
@@ -0,0 +1,51 @@
...
+        
+        List<String> cardRecipients = lightcard.getEntityRecipients(); 
+        List<String> cardRecipientsForInformation = lightcard.getEntityRecipientsForInformation();
+
+        List<String> cardTotalRecipients = new ArrayList<>(cardRecipients);
+        cardTotalRecipients.addAll(cardRecipientsForInformation);
+        Collection<CardSubscription> connections = this.cardSubscriptionService.getSubscriptions();
+        for (CardSubscription cardSubscription : connections) {
+            CurrentUserWithPerimeters userWithPerimeters = cardSubscription.getCurrentUserWithPerimeters();
+            if (CardRoutingUtilities.checkIfUserMustBeNotifiedForThisProcessState(lightcard.getProcess(), lightcard.getState(), userWithPerimeters)) {
```
freddidierRTE, line 40: The method checkIfUserMustBeNotifiedForThisProcessState does not test all the conditions for the user to be notify

You have to use checkIfUserMustReceiveTheCard()

But he method take as params a JSON Object so you need to "cut" this method in two to obtain a method with params : groupRecipients,entityRecipients,userRecipients & currentUserWithPerimeters

services/cards-publication/src/main/modeling/swagger.yaml
```diff
@@ -1080,6 +1080,16 @@ paths:
...
           description: Forbidden - User doesn't have any group
+  /connectedEntityPreview:
+    post:
+      summary: get the connected Entity among the recipients in card preview
+      consumes:
+        - application/json
+      responses:
+        '200':
+          description: OK
+        '400':
```
freddidierRTE, line 1091: You have to define here the parameters and the return type

freddidierRTE (changes requested):

Issue #5264, opened 2023-10-20
Add cypress tests for advanced task 


freddidierRTE:
in src/test/cypress/cypress/integration/BuiltInTemplates.spec.js  , add three new tests : 

**Test 1 :** 
Send a task  with option  daily freq and some day + some month selected + task description  + time + task duration + minutes before reminder 
Edit the card and check checkbox + input texts are well settled 

**Test 2 :** 
Send a task with monthly frequency + some months selected + Nth Day last day + a day of the month  + task description  + time + task duration + minutes before reminder 
Edit the card and check checkbox + input texts are well settled 

**Test 3:** 
Send a task with monthly frequency + some months selected + Nth Day weekday with occurrence number and weekday + task description  + time + task duration + minutes before reminder 
Edit the card and check checkbox + input texts are well settled

PR #5330, opened 2023-11-05
Add cypress tests for advanced task (#5264)

vlo-rte:
Nothing in release notes.

### DIFF
--- a/src/test/cypress/cypress/integration/TaskAdvanced.spec.js
+++ b/src/test/cypress/cypress/integration/TaskAdvanced.spec.js
@@ -1,0 +1,272 @@
+/* Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {UserCardCommands} from "../support/userCardCommands"
+import {OpfabGeneralCommands} from "../support/opfabGeneralCommands"
+import {FeedCommands} from "../support/feedCommands"
+import {ScriptCommands} from "../support/scriptCommands";
+
+describe('Task Advanced', function () {
+
+  const usercard = new UserCardCommands();
+  const opfab = new OpfabGeneralCommands();
+  const feed = new FeedCommands();
+  const script = new ScriptCommands();
+
+  before('Set up configuration', function () {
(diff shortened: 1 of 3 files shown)


src/test/cypress/cypress/integration/TaskAdvanced.spec.js
```diff
@@ -0,0 +1,300 @@
...
+
+      opfab.loginWithUser('operator2_fr');
+      feed.checkNumberOfDisplayedCardsIs(1);
+      feed.openFirstCard();
+      cy.get("#message").contains("Please confirm the issues in your area have been fixed");
+      cy.get("#comment").type("My answer")
+      cy.get("#opfab-card-details-btn-response").click();
+      cy.get("#childs-div").contains("My answer");
+    })
+  })
```
freddidierRTE, line 56: This test concerning question or message list is to be kept in BuildInTemplates.spec.js

src/test/cypress/cypress/integration/TaskAdvanced.spec.js
```diff
@@ -0,0 +1,300 @@
...
+
+    it('Send Task advanced user card for daily frequency from operator1_fr', () => {
+
+      opfab.loginWithUser('operator1_fr');
+      feed.checkNumberOfDisplayedCardsIs(0);
+      opfab.navigateToUserCard();
+      usercard.selectService('User card examples');
+      usercard.selectProcess('Task Advanced');
+      cy.get('#taskDescription').type('Test task for daily frequency');
+      cy.get('#Monday').click({force: true});
```
freddidierRTE, line 110: {force:true} is necessary here because the input html field (checkbox)  is not visible and replace with a custom checkbox so when selected the input cypress consider it as hidden and failed
 
To solve that pb , you need to click on  the label :

```
<label class="opfab-checkbox" style="padding-left:25px"> ${opfab.utils.getTranslation("shared.calendar.monday")} <input type="checkbox" id="Monday">   <span class="opfab-checkbox-checkmark"> </span>   </label>
```

To do that you can do for example : 

```
cy.get('.opfab-checkbox').contains('Monday').click();
```
code seen in src/test/cypress/cypress/support/activityAreaCommands.js

So you should be able to remove all force:true 

It is better to not use force:true to avoid having a passing test whereas checkbox are not visible

freddidierRTE (changes requested):

src/test/cypress/cypress/integration/TaskAdvanced.spec.js
vlo-rte: Done.

src/test/cypress/cypress/integration/TaskAdvanced.spec.js
vlo-rte: Done.

PR #5352, opened 2023-11-08
Implement ProcessesService with static methods

freddidierRTE:
Nothing in release note


ui/main/src/app/business/services/businessconfig/processes.service.ts
```diff
@@ -22,27 +21,28 @@ import {LightCard} from '@ofModel/light-card.model';
...
+    private static processServer: ProcessServer;
+    private static processesWithAllVersionsCache = new Map();
+    private static processesWithLatestVersionOnly: Process[];
+    private static processesWithAllVersions: Process[];
+    private static processGroups = new Map<string, {name: string; processes: string[]}>();
+
+    private static typeOfStatesPerProcessAndState: Map<string, TypeOfStateEnum>;
 
-    private typeOfStatesPerProcessAndState: Map<string, TypeOfStateEnum>;
+    constructor(private processServer: ProcessServer) {}
```
quinarygio, line 34: constructor should not be needed anymore

quinarygio (changes requested):

PR #5399, opened 2023-11-16
Refacto business data

freddidierRTE:
nothing in release note


ui/main/src/app/business/services/admin/crud-business-date.service.ts
quinarygio, line 1: rename to crud-business-data.service.ts

quinarygio (changes requested):

PR #5533, opened 2023-12-01
Use badge to show entity recipients for user card preview (#5448)

vlo-rte:
- In release note :
  -  In chapter : Features
  -  Text : #5448 : Use badge to show entity recipients for user card preview


vlo-rte:
> Add badge as well for recipients for information
> 
> image
> 
> Add some space between "RECIPIENTS:" and the first badge

Done.

Issue #5462, opened 2023-11-24
In question card build in template add date column for responses 


freddidierRTE:
![image](https://github.com/opfab/operatorfabric-core/assets/58251627/8b58a875-0bc7-4107-ae8b-9278a860b57a)

PR #5549, opened 2023-12-04
In question card build in template add date column for responses (#5462)

vlo-rte:
- In release note :
  -  In chapter : Features
  -  Text : #5462 : In question card build in template add date column for responses


freddidierRTE:
i suggest to have the date as the first column in the table

Issue #5483, opened 2023-11-27
Recurrent card : add a control in the code to forbid negative number for "minutes to remind before task"


vlo-rte:
Currently, setting a negative number in this field makes opfab crash.

PR #5554, opened 2023-12-05
Recurrent card : add a control in the code to forbid negative number for "minutes to remind before task" (#5483)

vlo-rte:
- In release note :
  -  In chapter : Bugs
  -  Text : #5483 : Recurrent card : add a control in the code to forbid negative number for "minutes to remind before task"


quinarygio:
Maybe we should add a karate test ?

vlo-rte:
> Maybe we should add a karate test ?

It's done.
Can you review again ? Thanks

quinarygio (approved):

PR #5558, opened 2023-12-05
Revert node upgrade for reminder service due to rrule incompatibility with node 21.3

freddidierRTE:
Nothing in release note , bug was not present in previous release


freddidierRTE:
cancel by #5560

Issue #5648, opened 2023-12-19
Integrate resizable columns for table using ag-grid


vlo-rte:
resize columns comes with update of ag-grid v31
https://github.com/opfab/operatorfabric-core/pull/5588

PR #5650, opened 2023-12-19
Integrate resizable columns for table using ag-grid (#5648)

vlo-rte:
Nothing in release notes.


freddidierRTE:
Should not be possible to adjust the last column in admin screen 

image

freddidierRTE:
After making a few test , i would prefer first to let the column resizing feature available only for admin screen and we will see later if we open the feature in other screens

vlo-rte:
> Should not be possible to adjust the last column in admin screen
> 
> image

Done.

vlo-rte:
> After making a few test , i would prefer first to let the column resizing feature available only for admin screen and we will see later if we open the feature in other screens

Done.

vlo-rte (commented):
PR validated by Frederic Didier.

PR #5651, opened 2023-12-20
Add actions field to Card. Handle PROPAGATE_READ_ACK_TO_PARENT_CARD action (#5461)

### LINKED ISSUE #5461: Enable the 'read' and 'acknowledgment' status of parent card depends of child card 'read' and 'acknowledgment' status
The objective is to introduce a feature enabling a parent card to have the card not read and not acknowledge if a  child card is published . To implement this functionality, a new field will be added to the card structure.

The proposed solution involves incorporating an 'action' field within the card, structured as an array of predefined actions. The initial predefined action to be introduced is 

'PROPAGATE_READ_ACK_TO_PARENT_CARD"

--> to be decided 

Anticipating future needs, additional predefined actions may be introduced, such as 'PROPAGATE_READ_ACK_TO_PARENT_CARD_FOR_PUBLISHER' and 'KEEP_READ_AND_ACK_OF_PREVIOUS_CARD_VERSION.'

Moreover, this new 'action' field could potentially replace the existing 'keepChildCards' field, transforming it into an action labeled 'KEEP_CHILD_CARDS_FROM_PREVIOUS_CARD' 

The use of this new field is to be done on the front code : 

 - when a card is read by user , it should set read in the back to the  card and child cards with value PARENT_CARD_READ_ACK_IF_CARD_READ_ACK
 -  when a card is ack by user , it should ack the card in the back  and the child cards with value PARENT_CARD_READ_ACK_IF_CARD_READ_ACK

On the front side : 
 - a card is consider ack if the card is ack and all the child card with PARENT_CARD_READ_ACK_IF_CARD_READ_ACK are ack
 - a card is consider read if the card is read and all the child card with PARENT_CARD_READ_ACK_IF_CARD_READ_ACK are read

Impact on lightCard and Card Components

quinarygio:
Fix #5461
- In release note :
  -  In chapter :  Features 
  -  Text : Add actions field to Card. Handle PROPAGATE_READ_ACK_TO_PARENT_CARD qction (#5461)

### DIFF
--- a/ui/main/src/app/business/store/lightcards/lightcards-store.ts
+++ b/ui/main/src/app/business/store/lightcards/lightcards-store.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2023, RTE (http://www.rte-france.com)
+/* Copyright (c) 2018-2024, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -7,7 +7,7 @@
  * This file is part of the OperatorFabric project.
  */
 
-import {LightCard} from '@ofModel/light-card.model';
+import {CardAction, LightCard} from '@ofModel/light-card.model';
 import {
     catchError,
     debounceTime,
@@ -28,6 +28,7 @@ import {CardOperationType} from '@ofModel/card-operation.model';
 import {LogOption, LoggerService as logger} from 'app/business/services/logs/logger.service';
 import {SelectedCardService} from 'app/business/services/card/selectedCard.service';
 import {AcknowledgeService} from '../../services/acknowledge.service';
+
 
 /**
  *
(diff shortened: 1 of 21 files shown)


ui/main/src/app/modules/card/components/card-ack/card-ack.component.ts
```diff
@@ -191,14 +202,22 @@ export class CardAckComponent implements OnInit, OnChanges, OnDestroy {
...
                 OpfabStore.getLightCardStore().setLightCardAcknowledgment(this.card.id, false);
+
+                const childCards = OpfabStore.getLightCardStore().getChildCards(this.card.id);
+                if (childCards) {
+                    childCards.forEach(child => {
+                        if (child.actions?.includes(CardAction.PROPAGATE_READ_ACK_TO_PARENT_CARD)) {
+                            AcknowledgeService.deleteUserAcknowledgement(child.uid).subscribe();
+                        }
+                    })
+                }
```
freddidierRTE, line 213: I think it is not necessary to cancel ack for child cards as if the parent ack is cancelled the card will be considered as unack whatever the status of the childs

ui/main/src/app/business/store/lightcards/lightcards-store.ts
```diff
@@ -246,6 +263,18 @@ export class LightCardsStore {
             } else {
                 this.childCards.set(card.parentCardId, [card]);
             }
+
+            const parent = this.lightCards.get(card.parentCardId);
```
freddidierRTE, line 267: parent could be null if child is received before parent cards (we have no guarantee of the receive order when loading cards) so you should check this case 
But if it arise , the processing of the ack/read should be done when receiving the card , it is already done for read card in your code line 239 but it has to be done as well for ack

freddidierRTE (changes requested):

ui/main/src/app/modules/card/components/card-ack/card-ack.component.ts
```diff
@@ -102,6 +104,38 @@ export class CardAckComponent implements OnInit, OnChanges, OnDestroy {
...
+            .pipe(
+                takeUntil(this.unsubscribe$),
+                map((lastCardDeleted) => {
+                    this.updateAcknowledgeButtonVisibilityIfNeeded(lastCardDeleted);
+                })
+            )
+            .subscribe();
+    }
+
+    private updateAcknowledgeButtonVisibilityIfNeeded(childCard) {
```
freddidierRTE, line 129: i would suggest naming it " updateAcknowledgeButtonVisibilityIfCardsIsChildOfCurrentCard"

ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -258,9 +276,19 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
...
+        if (this.childCards) {
+            this.childCards.forEach(child => {
+                if (child.actions?.includes(CardAction.PROPAGATE_READ_ACK_TO_PARENT_CARD) && !child.hasBeenRead) {
+                    CardService.postUserCardRead(child.uid).subscribe();
+                }
+            })
+        }
+        this.lastCardSetToReadButNotYetOnFeed = this.card;
+
     }
```
freddidierRTE, line 292: this.lastCardSetToReadButNotYetOnFeed = this.card shall only be done if read status change

So keep the old  this.lastCardSetToReadButNotYetOnFeed = this.card; 
and add a new after CardService.postUserCardRead(child.uid).subscribe();

ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -188,6 +195,16 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
...
+    private computeCardHasBeenRead() {
+        this.card = {...this.card,
+            hasBeenRead: OpfabStore.getLightCardStore().isLightCardHasBeenRead(this.card)
+        }
+    }
+
+    private computeCardHasBeenAcknowledged() {
+        this.card.hasBeenAcknowledged = OpfabStore.getLightCardStore().isLightCardHasBeenAcknowledged(this.card);
+    }
+
```
freddidierRTE, line 207: the ack status is not used directly in card-body.component , i suggest to avoid dealing with ack here but do the processing where it is needed i.e in setAcknowledgeButtonVisibility  method in card-ack.component.ts

freddidierRTE (changes requested):

ui/main/src/app/business/store/lightcards/lightcards-store.ts
```diff
@@ -233,19 +234,68 @@ export class LightCardsStore {
...
+        }
+        return hasBeenRead;
+    }
+
     private addChildCard(card: LightCard) {
         if (card.parentCardId) {
             const children = this.childCards.get(card.parentCardId);
             if (children) {
+                const childIndex = children.findIndex(child => child.id === card.id);
+                if (childIndex >= 0) children.splice(childIndex);
```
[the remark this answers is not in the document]
quinarygio: I saw that when a child card is updated we end up having the both version of the child card in the "children" collection. So before adding a child card I check if it already exists and eventually remove it before adding the updated one.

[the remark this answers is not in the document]
freddidierRTE: ok understood

src/test/resources/cards/defaultProcess/question.json
```diff
@@ -15,8 +15,9 @@
 		"summary" : {"key" : "message.summary"},
 		"title" : {"key" : "question.title"},
 		"data" : {},
-		"lttd" :  ${current_date_in_milliseconds_from_epoch_plus_3minutes},
+		"lttd" :  ${current_date_in_milliseconds_from_epoch_plus_4hours},
 		"timeSpans" : [
 			{"start" :  ${current_date_in_milliseconds_from_epoch_plus_4hours}}
-			]
+			],
+		"actions" : ["PROPAGATE_READ_ACK_TO_PARENT_CARD"]
```
freddidierRTE, line 22: I do not think these modifications are useful ?

PR #5706, opened 2024-01-09
Use entity role in usercards #5434

ClementBouvierN:
- In release note :
  -  In chapter :  Features 
  -  Text : #5434 Entities need the CARD_SENDER role to send a card and the CARD_RECEIVER role to receive a card


freddidierRTE:
Solve conflict and adjust documentation (https://opfab.github.io/documentation/current/reference_doc/#_entities_2)

PR #5779, opened 2024-01-22
Add a helper page to deal with delta quills format (#5777)

### LINKED ISSUE #5777: Add a helper html page to deal with delta quills format
the html page is accesible via a a specific url in opfab : /devtools/richtext

Two text area :  

On containing a quills text area 

Another containing the delta json format 

A button to convert from quill to json 

Like for example : https://www.base64decode.org/ or

quinarygio:
Fix #5777 

I wonder if we should add to release note as it is an hidden feature

### DIFF
--- a/ui/main/src/app/modules/devtools/richtext/richtext.component.ts
+++ b/ui/main/src/app/modules/devtools/richtext/richtext.component.ts
@@ -1,0 +1,62 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ *  See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {ChangeDetectionStrategy, Component, OnInit, Renderer2} from '@angular/core';
+
+import { Clipboard } from '@angular/cdk/clipboard';
+import {TranslationService} from 'app/business/services/translation/translation.service';
+
+declare const opfab;
+
+@Component({
+    selector: 'of-richtext-helper',
+    templateUrl: './richtext.component.html',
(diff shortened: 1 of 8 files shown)


ui/main/package.json
```diff
@@ -53,6 +53,7 @@
     "moment-timezone": "0.5.44",
     "ng-event-source": "1.0.14",
     "ngx-chips": "3.0.0",
+    "ngx-quill": "24.0.4",
     "ol": "8.2.0",
     "quill": "1.3.7",
     "rrule": "2.8.1",
```
freddidierRTE, line 59: instead of having quill import in two ways , i would prefer using the same solution as in build in template to manipulate 
quill component using :
` <opfab-richtext-editor id="myId"> </opfab-richtext-editor>`

The idea is to avoid having two solutions in the code base

Another advantage is that  we will be sure we share the same quill config (toolbarsOptions ...)

freddidierRTE (changes requested):

ui/main/src/app/modules/devtools/richtext/richtext.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+
+    convert() {
+        const quillEditor = document.getElementById('rich-editor') as any;
+        debugger;
+        this.delta = quillEditor.getContents();
+        
+    }
+
+    copyToClipboar(): void {
```
freddidierRTE, line 59: copyToClipboar**d**

ui/main/src/app/modules/devtools/richtext/richtext.component.ts
```diff
@@ -0,0 +1,63 @@
...
+        container.innerHTML = `
+        <label translate>${label}</label>
+        <opfab-richtext-editor id="rich-editor"></opfab-richtext-editor>
+        `
+    }
+
+
+    convert() {
+        const quillEditor = document.getElementById('rich-editor') as any;
+        debugger;
```
freddidierRTE, line 54: remove debug instruction

freddidierRTE (changes requested):

Issue #5844, opened 2024-02-01
Geographical map: visualize which Polygon representing a line on the geo-map belongs to which card in the card list


MonaRtei:
When the geographical map is enabled, points or polygons are drawn on the geo-maps for cards with geographical information in the feed.
At the moment, when the mouse hovers over a card in the feed, there is a highlight effect : 
- if it is a point, the radius gets bigger
- if it is a polygon, the fill color becomes more opaque

However, when Polygon is used to represent a line between 2 points, the highlight effect is not visible.

Possible solution:
highlight polygon by increasing the stroke width
https://github.com/opfab/operatorfabric-core/blob/develop/ui/main/src/app/modules/feed/components/map/map.component.ts#L420
![image](https://github.com/opfab/operatorfabric-core/assets/155954134/6d411fe3-e477-40d7-b2f3-1a978849f228)

```
private static polygonStyle(severity: Severity, highlight: boolean) {
        const fillOpacity = highlight ? 0.6 : 0.1;
        const strokeWidth = highlight ? 8 : 2; // Adding strokeWidth in the case the feature is highlighted
        return new Style({
            stroke: new Stroke({
                color: MapComponent.severityToColorMap(0.8)[severity],
                width: strokeWidth // Applying the new strokeWidth
            }),
            fill: new Fill({
                color: MapComponent.severityToColorMap(fillOpacity)[severity] // 'rgba(0, 0, 255, 0.1)'
            })
        });
    }
```

freddidierRTE:
To be able to adjust the stroke width , the value has to be a parameter in web-ui.json 

The parameter shall be named :  geomap.highlightPolygonStrokeWidth 

Default value is 2 

Add in documentation

PR #5864, opened 2024-02-05
Geographical map: visualize which Polygon representing a line on the geo-map belongs to which card in the card list (#5844)

vlo-rte:
- In release note :
  -  In chapter : Features
  -  Text : #5844 : Geographical map: visualize which Polygon representing a line on the geo-map belongs to which card in the card list


freddidierRTE:
Add in web-ui.json the value       "highlightPolygonStrokeWidth": 5 

The idea is that user can see in the config file that the param exist

vlo-rte:
> highlightPolygonStrokeWidth

Done.

PR #5910, opened 2024-02-08
Fix Ligthcard rRule serialization (#5892)

### LINKED ISSUE #5892: Task advanced disappearing from Calendar
Creating a new Task advanced card and going to Calendar page the card is shown for 1 second and than disappears.
Than editing and saving the card  is  correctly shown on calendar, but reloading the page it disappers againg

quinarygio:
Fix #5892 

Nothing in release note


services/cards-consultation/src/main/java/org/opfab/cards/consultation/model/LightCard.java
```diff
@@ -65,6 +67,7 @@ public class LightCard {
     private String wktProjection;
     private Integer secondsBeforeTimeSpanForReminder;
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonProperty("rRule")
```
freddidierRTE, line 70: As it is tricky , to avoid removing it in the future , add comment : `// if we don't use this annotation, the field will be serialized as "rrule"`

PR #5923, opened 2024-02-09
Add logLevel endpoint for node services (#5869)

### LINKED ISSUE #5869: Add logLevel endpoint for node services
For each node services , add an endpoint   /logLevel

When making GET /logLevel  it returns a string with the log level 
When making a POST /logLevel it set the log level set in the body of the request 

in src/test/resources/logs create script getLogsNodeService.sh and setLogsNodeService.sh 

add in documentation in https://opfab.github.io/documentation/current/deployment/#_runtime_configuration

quinarygio:
Fix #5869 

- In release note :
  -  In chapter :  Features 
  -  Text : #5869 : Add logLevel endpoint for node services

### DIFF
--- a/node-services/common/src/common/server-side/logger.ts
+++ b/node-services/common/src/common/server-side/logger.ts
@@ -11,54 +11,75 @@ import winston, {format} from 'winston';
 import DailyRotateFile from 'winston-daily-rotate-file';
 import config from 'config';
 
-const logFormat = winston.format.combine(
-    winston.format.colorize(),
-    winston.format.timestamp(),
-    winston.format.align(),
-    winston.format.printf((info) => `${info.timestamp} ${info.level}: ${info.message}`)
-);
+export default class Logger {
+    static logger: winston.Logger;
 
-const logConfiguration: any = config.get('operatorfabric.logConfig');
+    static readonly logFormat = winston.format.combine(
+        winston.format.colorize(),
+        winston.format.timestamp(),
+        winston.format.align(),
+        winston.format.printf((info) => `${info.timestamp} ${info.level}: ${info.message}`)
+    );
 
-let logger = winston.createLogger({
-    format: format.combine(
-        format((info: winston.LogEntry) => {
-            info.message = prefixNewLinesToAvoidLogInjection(info.message);
-            return info;
-        })(),
-        logFormat
-    ),
-    transports: [new winston.transports.Console({level: logConfiguration.logLevel})]
-});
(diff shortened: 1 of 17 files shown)


node-services/common/src/common/server-side/logger.ts
```diff
@@ -11,54 +11,74 @@ import winston, {format} from 'winston';
...
-        datePattern: 'YYYY-MM-DD',
-        zippedArchive: true,
-        maxSize: '20m',
-        maxFiles: '14d',
-        level: logConfiguration.logLevel
-    });
+    public static getLogger(): winston.Logger {
+        this.logger = winston.createLogger({
+            format: format.combine(
+                format((info) => {
```
freddidierRTE, line 29: I suggest to create the logger only if it does not exist (i.e only once)

Even if in the actual code , the service call only one time getLogger() , it may be use differently by a developer in the future . (for example getting logger in each class instead of passing the logger object) .

src/test/resources/logs/getLogsNodeService.sh
```diff
@@ -0,0 +1,28 @@
...
+if [[ -z $2 ]]
+then
+    echo "Usage : getLogsNodeService.sh host:port user opfab_url"
+else
+	serviceAddress=$1
+	user=$2
+	
+	source ../getToken.sh $url $user
+	
+	response=$( curl -s "http://$serviceAddress/logLevel" -H "Authorization: Bearer $token" -H "Content-type:application/json")
```
freddidierRTE, line 26: This script is based on existing script for java services .
I realized that it is not possible to set a password , could you add this possibility for the existing and for the new scripts 

`Usage : getLogsNodeService.sh host:port user passwd opfab_url`

With default value test

src/test/resources/logs/setLogsNodeService.sh
```diff
@@ -0,0 +1,29 @@
...
+cd "$(dirname "${BASH_SOURCE[0]}")"
+url=$4
+if [[ -z $url ]]
+then
+	url="http://localhost"
+fi
+if [[ -z $3 ]]
+then
+    echo "Usage : setLogsNodeService.sh host:port level user opfab_url"
+else
```
freddidierRTE, line 20: Same remark as the other script

freddidierRTE (changes requested):

freddidierRTE:
Solve the four sonar issues : https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5923&id=org.lfenergy.operatorfabric%3Aoperatorfabric-core

PR #5969, opened 2024-02-22
Refacto process monitoring - first step

vlo-rte:
Nothing in release notes.

### DIFF
--- a/ui/main/src/app/business/view/processmonitoring/processmonitoring.view.ts
+++ b/ui/main/src/app/business/view/processmonitoring/processmonitoring.view.ts
@@ -1,0 +1,97 @@
+/* Copyright (c) 2023-2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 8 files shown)


ui/main/src/app/business/view/processmonitoring/processmonitoring.view.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import {ProcessesService} from 'app/business/services/businessconfig/processes.service';
+import {ProcessGroups, ProcessToMonitor, StatePerProcessToMonitor} from './processmonitoringPage';
+import {UserService} from 'app/business/services/users/user.service';
+import {Utilities} from 'app/business/common/utilities';
+import {Process} from '@ofModel/processes.model';
+
+export class ProcessMonitoringView {
+    private processesToMonitor: ProcessToMonitor[] = null;
```
quinarygio, line 17: Is it really necessary to define processToMonitor as a class variable? why not define it as a local variable in getProcessList()  method and passing it as parameter to addProcessIfUserHasPermission() method ? It  would also avoid clonig it on response.

vlo-rte: After analyzing the code, we leave it as it is because putting a global variable prevents the function getProcessList() from redoing the calculation each time it is called. And this function is called several times, notably in the function getProcessesPerProcessGroups().

quinarygio (approved):

PR #6029, opened 2024-02-28
Ignore false positive sonar security rexgexp issue in edit user modal

freddidierRTE:
Sonar issue : 
image

Nothing in release note

### DIFF
--- a/ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
+++ b/ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
@@ -76,7 +76,7 @@ export class EditUserModalComponent implements OnInit {
             comment: new FormControl('', []),
             groups: new FormControl([]),
             entities: new FormControl([]),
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/admin/components/editmodal/users/edit-user-modal.component.ts
```diff
@@ -76,6 +76,8 @@ export class EditUserModalComponent implements OnInit {
             comment: new FormControl('', []),
             groups: new FormControl([]),
             entities: new FormControl([]),
+            //NOSONAR : The regex is used to validate the input of the authorizedIPAddresses field
+            // and is not a security issue as it is done on the client side
             authorizedIPAddresses: new FormControl('', [Validators.pattern(/^((\d+(\.\d+){3}),?\s?)+$/)])
```
ClementBouvierN, line 81: Change the regex pattern to 
^(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3},?)+$
The nested check ( {3} ) is what causes the sonar trigger.

freddidierRTE: You are right , done

Issue #6032, opened 2024-02-29
Add an export button for "user action logs" screen


freddidierRTE:
Same as for the admin screens

PR #6069, opened 2024-03-05
Add an export button for "user action logs" screen (#6032)

vlo-rte:
- In release note :
  -  In chapter :  Features
  -  Text : #6032 : Add an export button for "user action logs" screen

### DIFF
--- a/ui/main/src/app/business/view/useractionlogs/userActionLogs.view.ts
+++ b/ui/main/src/app/business/view/useractionlogs/userActionLogs.view.ts
@@ -23,6 +23,7 @@ import {map, Observable, of, switchMap} from 'rxjs';
 import {UserActionLogLine} from './userActionLogLine';
 import {UserActionLogsResult} from './userActionLogsResult';
 import {UserActionLogsPageDescription} from './userActionLogsPageDescription';
+import {ExcelExport} from '../../common/excel-export';
 
 export class UserActionLogsView {
     private userActionLogPage = new UserActionLogsPageDescription();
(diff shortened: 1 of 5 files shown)


ui/main/src/app/business/view/useractionlogs/userActionLogs.view.ts
```diff
@@ -97,12 +97,12 @@ export class UserActionLogsView {
         this.pageNumber = pageNumber;
     }
 
-    public search(): Observable<UserActionLogsResult> {
+    public search(isForExport: boolean): Observable<UserActionLogsResult> {
```
freddidierRTE, line 100: set default value to false , it will avoid to set the value in test for example

ui/main/src/app/modules/useractionlogs/useractionlogs.component.ts
```diff
@@ -176,6 +181,39 @@ export class UserActionLogsComponent implements OnInit {
...
+                    [actionColumnName]: line.action,
+                    [loginColumnName]: line.login,
+                    [entitiesColumnName]: line.entities,
+                    [cardUidColumnName]: line.cardUid,
+                    [commentColumnName]: line.comment
+                });
+            });
+            ExcelExport.exportJsonToExcelFile(exportUserActionLogsData, 'UserActionLogs');
+        });
+    }
```
freddidierRTE, line 215: Put this code which is not angular related in the userActionLogs.view.ts 
No unit test 

I do not think you need to call this.setViewParametersFromForm(0); because the export shall be based on what is visible on screen i.e the last search

freddidierRTE (changes requested):

ui/main/src/app/modules/useractionlogs/useractionlogs.component.ts
vlo-rte: Done.

ui/main/src/app/business/view/useractionlogs/userActionLogs.view.ts
vlo-rte: Done.

Issue #6067, opened 2024-03-05
Do not exclude current user connected on activity area screen 


freddidierRTE:
In the following example , it means having blue bubble for all activity area and the login of operator4_fr dispayed for all activity area

![image](https://github.com/opfab/operatorfabric-core/assets/58251627/a2e22751-8ff2-4062-9c3c-ea1300f8bdb4)

PR #6070, opened 2024-03-05
Do not exclude current user connected on activity area screen (#6067)

vlo-rte:
- In release note :
  -  In chapter :  Features
  -  Text : #6067 : Do not exclude current user connected on activity area screen

### DIFF
--- a/ui/main/src/app/business/view/activityarea/activityarea.view.spec.ts
+++ b/ui/main/src/app/business/view/activityarea/activityarea.view.spec.ts
@@ -225,7 +225,7 @@ describe('ActivityAreaView', () => {
         expect(activityAreaPage.activityAreaClusters[0].lines[1].connectedUsers).toEqual([]);
     });
 
-    it('GIVEN a user member of entity1 and entity2 WHEN user is currently connected to entity1 THEN entity1 line does not contains the current user login ', async () => {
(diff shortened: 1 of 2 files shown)


ui/main/src/app/business/view/activityarea/activityarea.view.spec.ts
```diff
@@ -238,7 +238,10 @@ describe('ActivityAreaView', () => {
         const activityAreaPage = await firstValueFrom(activityAreaView.getActivityAreaPage());
         expect(activityAreaPage.activityAreaClusters[0].lines).toHaveSize(2);
         expect(activityAreaPage.activityAreaClusters[0].lines[0].entityId).toEqual('ENTITY1');
-        expect(activityAreaPage.activityAreaClusters[0].lines[0].connectedUsers).toEqual(['anotherUser']);
+        expect(activityAreaPage.activityAreaClusters[0].lines[0].connectedUsers).toEqual([
+            'anotherUser',
+            'currentUser'
+        ]);
```
freddidierRTE, line 244: the current test is to be removed as it was testing the fact that we remove the current user 

`it('GIVEN a user member of entity1 and entity2 WHEN user is currently connected to entity1 THEN entity1 line does not contains the current user login ',`

vlo-rte: Test not removed but the title of the test has been updated for the new behavior.

PR #6083, opened 2024-03-06
Render quill delta in mail (#5878)

### LINKED ISSUE #5878: Render quill delta in mail
Use specific handlebar helper  + jsDom + quills

See what has been done in https://github.com/joelcolucci/node-quill-converter/tree/master

quinarygio:
Fix #5878 

- In release note :
  -  In chapter :  Features
  -  Text : #5878 : Render rich text in mail body


freddidierRTE:
It is working well 

I realize that it would be better to separate this helper in a specific helper class , the idea is to keep a common class with the front.

So could you put the code in a new Helper Service named mailHandlebarsHelpers.ts

And could you add a unit test like for the other helpers in the front code (mailHandlebarsHelpers.test.ts)


Add documentation in https://opfab.github.io/documentation/archives/SNAPSHOT/docs/single_page_doc.html#_email_notifications for this specific template (like documentation for card template in ui, https://opfab.github.io/documentation/archives/SNAPSHOT/docs/single_page_doc.html#_operatorfabric_specific_handlebars_helpers)

node-services/cards-external-diffusion/src/domain/server-side/handlebarsHelpers.ts
```diff
@@ -320,6 +322,38 @@ export class HandlebarsHelper {
...
+              document.execCommand = function (command, showUI, value) {
+                try {
+                    return document.execCommand(command, showUI, value);
+                } catch(e) {}
+                return false;
+              };
+            </script>`
+
+            const DOM = new JSDOM(TEMPLATE, { runScripts: 'dangerously', resources: 'usable' });
+            const quill = new DOM.window.Quill('#editor');
```
freddidierRTE, line 350: I would load all this once on startup and store quill as static variable .

So you just have to call the setContents in the helper method

node-services/cards-external-diffusion/src/domain/server-side/handlebarsHelpers.ts
```diff
@@ -320,6 +322,38 @@ export class HandlebarsHelper {
...
+ 
+            quill.setContents(JSON.parse(delta));            
+            const html = quill.root.innerHTML;
+            return html;
```
freddidierRTE, line 354: Add a try catch to deal with error in parsing

PR #6103, opened 2024-03-11
Refactor codebase to integrate ESLint/Prettier in cards-external-diffusion service

freddidierRTE:
- Integrate ESLint and Prettier packages to enforce consistent code style
- Resolve all ESLint and Prettier issues throughout the codebase, in particular
	- Refactor code to use explicit `null` checks instead of coercion (check undefined or null objects with explicit == null instead of !)
	- Replace `require` with `import` for module imports
	- Add explicit return types to all functions
	- Simplify Promise usage by :
		- removing unnecessary `Promise.resolve()`
		- replacing `Promise.resolve(myVar)` with `return myVar`
		- replacing `Promise.reject(new Error())` with `throw new Error()`
	- Introduce `Card` and `UserWithPerimeter` objects to minimize usage of `any` type
- Add ESLint disable comments where necessary due to complexity or lack of immediate solution
- Move test directory into `src` for better project structure and consistency with other node services
- Adjust some log debug

**Nothing in release note**


node-services/cards-external-diffusion/src/cardsExternalDiffusion.ts
```diff
@@ -23,38 +23,44 @@ import CardsExternalDiffusionDatabaseService from './domain/server-side/cardsExt
...
 
-// Token verification activated except for heathcheck request 
-const jwksUri : string =  config.get('operatorfabric.security.oauth2.resourceserver.jwt.jwk-set-uri');
+/* eslint-disable */
+// disable eslint as false positive , promise are authorized see
+// https://community.sonarsource.com/t/express-router-promise-returned-in-function-argument-where-a-void-return-was-expected/95772
+const jwksUri: string = config.get('operatorfabric.security.oauth2.resourceserver.jwt.jwk-set-uri');
 app.use(
-    /\/((?!healthcheck).)*/,
+    /\/((?!healthcheck).)*/, // Token verification activated except for heathcheck request
```
ClementBouvierN, line 35: except for hea**l**th check

PR #6115, opened 2024-03-12
Add path controls in PathUtils to mitigate path manipulation vulnerabilities

freddidierRTE:
- In release note :
  -  In chapter :  Tasks
  -  Text : #6115 Add path controls in PathUtils to mitigate path manipulation vulnerabilities

### DIFF
--- a/tools/generic/utilities/src/main/java/org/opfab/utilities/PathUtils.java
+++ b/tools/generic/utilities/src/main/java/org/opfab/utilities/PathUtils.java
@@ -7,10 +7,7 @@
  * This file is part of the OperatorFabric project.
  */
 
-
-
 package org.opfab.utilities;
-
 
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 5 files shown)


tools/generic/utilities/src/main/java/org/opfab/utilities/PathUtils.java
```diff
@@ -64,26 +67,33 @@ public static void moveDir(Path source, Path target) throws IOException {
...
     Files.walkFileTree(source, new CopyDir(source, target));
   }
 
+  private static void throwExceptionIfPathIsOutsideOfApplicationBasePath(Path path) throws IOException {
+    if (isPathOutsideOfApplicationBasePath(path))
+      throw new IOException("Path " + path.toString() + " is not in application base path " + applicationBasePath);
+  }
+
+  public static boolean isPathOutsideOfApplicationBasePath(Path file) {
+    return !file.toAbsolutePath().startsWith(applicationBasePath);
```
quinarygio, line 86: It could be better to call file.toRealPath​() instead of fle.toAbsolutePath() as it would remove any "redundant name element" as ".."

freddidierRTE: The issue with the toRealPath() method is that it throws an exception when the specified path does not exist. This becomes problematic when creating a bundle, as we need to create a directory that doesn't yet exist. Therefore, this method is not suitable for use in this context.

quinarygio (approved):

PR #6116, opened 2024-03-12
Add unit tests for angular translation service

freddidierRTE:
Nothing in release note


ui/main/src/app/services/angularTranslationService.spec.ts
```diff
@@ -0,0 +1,58 @@
+/* Copyright (c) 2023-2024, RTE (http://www.rte-france.com)
```
ClementBouvierN: /* Copyright (c) 2024, RTE (http://www.rte-france.com)

ClementBouvierN (approved):

PR #6171, opened 2024-03-19
Admin table handles strings instead of arrays

ClementBouvierN:
Nothing in release note


ui/main/src/app/business/services/admin/crudUtils.ts
```diff
@@ -11,6 +11,13 @@ import {Utilities} from 'app/business/common/utilities';
 import {EntitiesService} from '../users/entities.service';
 
 export class CrudUtilities {
+    static formatArrayToString(permissions: any) {
+        if (permissions != null) {
+            permissions.sort((a, b) => Utilities.compareObj(a, b));
+            return permissions.join(', ');
+        } else return '';
+    }
```
freddidierRTE: use arrayToConvert instead of permissions

freddidierRTE (approved):

PR #6215, opened 2024-03-25
Reduce Keycloak requests to improve configuration loading speed (#6176)

### LINKED ISSUE #6176: Reduce Keycloak Requests to Improve Configuration Loading Speed
The current implementation makes excessive requests to Keycloak during configuration loading, resulting in slower performance. To mitigate this, we propose modifying all scripts to use a shell variable,` opfab_token`, instead of requesting a token via getToken.sh.

The exceptions to this change are `loadTestConf.sh` and `send6TestCards.sh`, which should have a call to getToken added at the beginning of the script.

All scripts should also be updated to check if the _opfab_token_ environment variable is present. If it's not, the script should prompt an error message to the user.

quinarygio:
Fix #6176 

Nothing in release notes

### DIFF
--- a/src/test/resources/cardsForStressTests/send6xN_TestCards.sh
+++ b/src/test/resources/cardsForStressTests/send6xN_TestCards.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2021, RTE (http://www.rte-france.com)
+# Copyright (c) 2021-2024, RTE (http://www.rte-france.com)
 # See AUTHORS.txt
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -24,7 +24,7 @@ then
 
 else	
 (
-cd cards
+source ../getToken.sh $url "publisher_test"
 for i in `seq 1 $1`;
 do
 
--- a/src/test/resources/getToken.sh
+++ b/src/test/resources/getToken.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2021-2023, RTE (http://www.rte-france.com)
+# Copyright (c) 2021-2024, RTE (http://www.rte-france.com)
 # See AUTHORS.txt
 # This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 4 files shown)


freddidierRTE:
The current implementation functions correctly, but it requires at least on manual execution of source ./getToken.sh before running scripts such as ./loadTestConf.sh.

I suggest automatically sourcing the token at the beginning of loadTestConf.sh. You can achieve this by adding source ./getToken.sh $url at the start of the script.

The same enhancement should be applied to the following scripts for consistency and ease of use:

- send6TestCards.sh
- cardsForStressTests/send6xN_TestCards.sh

quinarygio:
> The current implementation functions correctly, but it requires at least on manual execution of source ./getToken.sh before running scripts such as ./loadTestConf.sh.
> 
> I suggest automatically sourcing the token at the beginning of loadTestConf.sh. You can achieve this by adding source ./getToken.sh $url at the start of the script.
> 
> The same enhancement should be applied to the following scripts for consistency and ease of use:
> 
> * send6TestCards.sh
> * cardsForStressTests/send6xN_TestCards.sh

added only to loadTestConf.sh because send6TestCards.sh and cardsForStressTests/send6xN_TestCards.sh are calling the sendCards.sh scripts where it is already sourced

freddidierRTE:
When it launch the send6xN_TestCards.sh , it gets the token all the time : 

```
./send6xN_TestCards.sh 2
Get token for user publisher_test on http://localhost
send card defaultProcess/message.json (url: http://localhost)
{"id":"defaultProcess.326861214831341","uid":"0094c878-3ff9-4874-9ba2-a0e3d7e15b74"}Get token for user publisher_test on http://localhost
send card defaultProcess/chart.json (url: http://localhost)
{"id":"defaultProcess.66862419818441","uid":"0146540e-f1a7-4f9c-bdbb-381864d8447c"}Get token for user publisher_test on http://localhost
send card defaultProcess/process.json (url: http://localhost)
{"id":"defaultProcess.1345434805541","uid":"9fc7708f-2e62-4760-a0b7-eb53ea86982c"}Get token for user publisher_test on http://localhost
```

I think it should be source at a higher level

freddidierRTE (approved):

PR #6259, opened 2024-04-04
Add an edit button to edit an external device configuration (#5950)

### LINKED ISSUE #5950: Add an edit button to edit an external device configuration
Add an edit button for each line that open a popup for edition (similar to other admin screen)

image

vlo-rte:
The issue mentioned by sonarcloud is not related to the work done for this issue.

- In release note :
  -  In chapter : Features
  -  Text : #5950 : Add an edit button to edit an external device configuration


src/test/api/karate/externaldevices/createDeviceConfig.feature
```diff
@@ -31,7 +31,7 @@ Feature: Device Configuration Management (Create)
     And header Authorization = 'Bearer ' + authToken
     And request configuration
     When method post
-    Then status 400
+    Then status 201
```
quinarygio, line 34: The Scenario description should be "Update an existing configuration" ?
And as it is an update I would expect a 200 status code

services/external-devices/src/main/java/org/opfab/externaldevices/controllers/ConfigurationsController.java
```diff
@@ -71,7 +71,7 @@ public Void createDeviceConfiguration(HttpServletRequest request, HttpServletRes
             @Valid @RequestBody DeviceConfiguration deviceConfiguration) {
 
         String id = deviceConfiguration.id;
-        configService.insertDeviceConfiguration(deviceConfiguration);
+        configService.saveDeviceConfiguration(deviceConfiguration);
```
quinarygio, line 74: When it is an update it should return a 200 status code

quinarygio (changes requested):

quinarygio:
When editing a "disabled" device, the modal has always the checkbox "is enabled" checked

freddidierRTE:
@quinarygio  :  You've made a valid point about the return code.  However, I suggest we maintain the current code for these reasons:

- This API is not intended for third-party use.
- The feature in question is minor.
- It doesn't affect reliability or security.
- Addressing this would require time, and there are other higher-priority tasks to focus on.

vlo-rte:
> When editing a "disabled" device, the modal has always the checkbox "is enabled" checked

Problem solved. Please check again

src/test/api/karate/externaldevices/createDeviceConfig.feature
vlo-rte: You're right, I've renamed the scenario description.

quinarygio (approved):

PR #6275, opened 2024-04-08
Add geographical view for monitoring process screen (#6242)

### LINKED ISSUE #6242: Add geographical view for monitoring process screen
We aim to enhance the user experience on the monitoring process screen by introducing a "View Results on a Geographical Map" feature. This feature will allow users to visualize search results on a geographical map. Here are the specifications for this feature:

- The feature should be represented as a checkbox labeled "View Results on a Geographical Map".
- Upon selection, this option should display a geographical map similar to the existing one in the feed.
- The availability of this feature should be contingent on the feed.geomap.enableMap being set to true in the web-ui.json configuration file.
- The configuration for this feature should use the settings defined under feed.geomap in the web-ui.json file.

The differences between this feature and the feed version are as follows:

- Clicking on a card on the map should open the card details in a popup.
- The map should display the results of all cards returned from the backend, instead of just the cards currently in memory.
- The map should not update in real-time; it should only update when the user clicks on the search button.

Please note that no Cypress tests are required for this feature. 

The feature should be documented in the Process Monitoring section of the OperatorFabric documentation (https://opfab.github.io/documentation/current/reference_doc/#_process_monitoring).

quinarygio:
Fix #6242 

- In release note :
  -  In chapter :  Features
  -  Text : #6242 : Add geographical view for monitoring process screen

### DIFF
--- a/src/docs/asciidoc/reference_doc/monitoring.adoc
+++ b/src/docs/asciidoc/reference_doc/monitoring.adoc
@@ -48,6 +48,7 @@ This screen provides the following features :
 in the `web-ui.json` configuration file. This object is a list of the fields you want to see in the array.
 - depending on the permissions of the groups to which the user is attached, he can also see the cards he is not the
 recipient of.
+- if geographical map view is enabled ('feed.geomap.enableMap' set to 'true' in web-ui.json) it is possible to view the cards on the map by checking the "View Results on a Geographical Map" checkbox.
 
 Here is an example of processMonitoring configuration :
 ```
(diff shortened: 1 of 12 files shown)


freddidierRTE:
- Minimize the margin around the geomap to make it closer to the screen border.  or maybe no margin ? to test ? 
- The 'View Result on Geomap' should be closely aligned with the 'Search Header'.
- Unlike the current monitoring table which uses pagination, the geomap should retrieve all cards, similar to the export function.
- Adjust the initial zoom setting. The current setting zooms in too much, making the blue circle in Arnhem invisible.
- Bug Fix: After navigating from 'Process Monitoring' to 'Show Result on Geomap' and then back to the feed, the geomap is no longer visible on the feed

freddidierRTE:
the common abstract map class shall be in ui/main/src/app/modules/share/map

freddidierRTE:
Last tuning  : could you avoid vertical scroll bar that always appear ?

image

freddidierRTE (approved):

PR #6281, opened 2024-04-09
Added the lastUpdate field to the card object (#6248)

### LINKED ISSUE #6248: Implement lastUpdate Field in Cards for Enhanced Connection Recovery Mechanism
At present, when a user loses connection, the system retrieves missed cards by querying the database based on their publish date. This method, however, has two significant limitations:

It does not consider acknowledgements (ack/unack) made on older cards.
The forthcoming change in issue #6183 will enable card updates without altering the publish date, potentially causing missed updates during connection recovery.
To overcome these challenges, we propose the introduction of a lastUpdate field in the card documents within the database. This field will record the date of the most recent modification and will be updated to the current date each time the card undergoes any form of update, including changes to the card itself, acknowledgements, and read status.

The proposed changes for this issue include:

- Addition of a lastUpdate field to the card documents in the backend.
- Setting lastUpdate to the current date upon card creation.
- Updating lastUpdate to the current date when a card is modified.
- Updating lastUpdate to the current date when a card's acknowledgement or read status changes.

The CardRepositoryImpl.java file appears to be the most suitable location for implementing these changes. Here's a sample of how to set the lastUpdate field:

```
mongoTemplate.updateFirst(query(where("_id").is(cardId)), 
    new Update().pull(USERS_READS, userName).set("lastUpdate", new Date()), Card.class);
```

The lastUpdate field in MongoDB should be stored as an epoch, in line with other date fields.

To test this feature, we should employ Karate tests instead of unit tests. This approach is more suitable as the logic is implemented directly in the MongoDB repository.

 Five tests should be conducted:

- Card update
- Add read
- Unread
- Ack
- Unack

Each test should:

- Send a card.
- Retrieve the card and verify that lastUpdate is present and its value is within the range of (now - 10s) and now.
- Apply modifications to the card and check that lastUpdate has been updated to a value greater than its previous value and less than or equal to now.
 

Please note that due to the precision of timestamps, checking the exact value of lastUpdate would be challenging. Hence, we suggest checking a timeframe instead of a precise value.

The UI-side changes to utilize the lastUpdate field for connection recovery will be addressed in a separate issue.

ClementBouvierN:
- In release note :
  -  In chapter :  Features
  -  Text : #6248 The date of the most recent modification to a card is now saved in the "lastUpdate" field

### DIFF
--- a/services/cards-consultation/src/main/java/org/opfab/cards/consultation/model/Card.java
+++ b/services/cards-consultation/src/main/java/org/opfab/cards/consultation/model/Card.java
@@ -51,8 +51,9 @@ public class Card {
     private String titleTranslated;
     private String summaryTranslated;
 
-    @CreatedDate
     private Instant publishDate;
+
+    private Instant lastUpdate;
 
     private Instant lttd;
     private Instant startDate;
--- a/services/cards-publication/src/main/java/org/opfab/cards/publication/model/Card.java
+++ b/services/cards-publication/src/main/java/org/opfab/cards/publication/model/Card.java
@@ -70,6 +70,10 @@ public class Card  {
     @Indexed
     private Instant publishDate;
 
+    @CreatedDate
+    @Indexed
+    private Instant lastUpdate;
+
     private Instant lttd;
     
     @Indexed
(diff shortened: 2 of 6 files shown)


services/cards-consultation/src/main/java/org/opfab/cards/consultation/model/Card.java
```diff
@@ -54,6 +54,9 @@ public class Card {
     @CreatedDate
     private Instant publishDate;
 
+    @CreatedDate
```
freddidierRTE: Annotation to remove

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6289, opened 2024-04-10
Fix child cards dates when updating parent card and keeping child cards (#6202)

### LINKED ISSUE #6202: Mismatch between Parent and Child Card Dates when Updating Parent Card with keepChildCard set to True
**Problem :** 

Currently, when loading cards on the feed or monitoring, we filter by publishDate, startDate, and endDate based on the selected period. The request does not explicitly load the child card associated with each card. The child cards are retrieved because they are expected to match the date filter of the parent card. This assumes that the period defined in the child card is identical to that of the parent card.

However, an issue arises when the parent card is updated with the keepChildCard flag set to true. If the start or end date of the parent card is updated, it results in a mismatch between the periods of the child and parent cards. This can lead to a situation where the parent card is loaded, but not the child cards, causing errors in the feed.

**Solution :**

To resolve this, we propose updating the child card's startDate and endDate when the parent card is updated. Specifically, we suggest setting the startDate as the earlier date between the parent's publishDate and startDate, and setting the endDate to the parent's endDate.

The rationale for using the earlier date can be illustrated with the following example:

Consider an initial parent card with a publishDate of 10/01, a startDate of 15/10, and an endDate of 15/11. The corresponding child card has a publishDate of 10/02, a startDate of 15/10, and an endDate of 15/11. Now, let's say the parent card is updated with a new publishDate of 15/03, while the startDate and endDate remain the same.

If we were to view the cards on 16/03, the child card would not be visible. This is because its publishDate is not within the selected period, its startDate is not within the period, and the selected period does not overlap with the startDate to endDate period of the child card. To rectify this, we can set the startDate of the child card to the new publishDate of the parent card (15/03). This adjustment ensures that the child card is correctly loaded within the selected period.

It's important to note that this issue doesn't occur when the card isn't updated, as the child card's publishDate is typically set after the parent card's publishDate. Also, when navigating to past entries in the feed, we load intermediate cards, which mitigates the problem.

However, this is a complex issue and the current behavior might not hold true if we alter the way we navigate to past entries in the future. Therefore, to ensure robustness, it's safer to set the startDate of the child card as the minimum of the parent card's startDate and publishDate during the creation of the child card. This approach provides a safeguard against potential future changes in navigation behavior.

Another scenario to consider is when the parent card does not have an endDate. In such cases, to ensure proper card updates, we should set the endDate of the child card to the later date between the parent card's startDate and publishDate. This adjustment ensures that the child card remains correctly associated with the parent card, even when updates are made.

**Proposed Changes:**

- Enforce in the back the creation of child cards with startDate = min(parent.startDate, parent.publishDate) and endDate = parent.endDate.
- Update child cards when receiving a parentCard update with the same rule and if parent endDate is null  set the endDate of the child card to the later date between the parent card's startDate and publishDate

This approach ensures that the child card is loaded whenever the parent card is loaded.

An alternative solution would be to include the child card inside the parent card. While this seems interesting, it would have significant implications on the code and would require a substantial data migration for existing production databases. Therefore, we propose sticking to the initial solution and considering this alternative for the future after a more in-depth impact analysis.

quinarygio:
Fix #6202

- In release note :
  -  In chapter :  Bugs
  -  Text : #6202 : Fix child cards dates when updating parent card and keeping child cards

### DIFF
--- a/services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
+++ b/services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
@@ -52,6 +52,7 @@ import jakarta.validation.ConstraintViolationException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
```diff
@@ -412,6 +421,71 @@ void GIVEN_an_invalid_card_WHEN_sending_card_THEN_card_is_rejected() {
...
+                                .startDate(Instant.now())
+                                .state("state1")
+                                .build();
+
+                Assertions.assertThatCode(
+                                () -> cardProcessingService.processUserCard(childCard, currentUserWithPerimeters,
+                                                token))
+                                .doesNotThrowAnyException();
+                Assertions.assertThat(checkChildCardsDates(childCard, card)).isTrue();
+        }
```
freddidierRTE: instead of using method checkChildCardsDates , i suggest having a precise assert 

Like : 

```
assertThat(chidcard.startDate).equals(card.publishDate)
assertThat(chidcard.endDate).equals(card.publishDate)
```

The idea is to ease the understanding of the test

services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
```diff
@@ -412,6 +421,71 @@ void GIVEN_an_invalid_card_WHEN_sending_card_THEN_card_is_rejected() {
...
+                Assertions.assertThat(checkCardCount(2)).isTrue();
+                Assertions.assertThat(checkChildCardsDates(childCard, card)).isTrue();
+
+                card.setActions(List.of(CardActionEnum.KEEP_CHILD_CARDS));
+                card.setStartDate(Instant.now().plus(1, ChronoUnit.DAYS));
+                card.setEndDate(Instant.now().plus(5, ChronoUnit.DAYS));
+                cardProcessingService.processCard(card);
+
+                Assertions.assertThat(checkCardCount(2)).isTrue();
+                Assertions.assertThat(checkChildCardsDates(childCard, card)).isTrue();
```
freddidierRTE: Same remark as previous test

services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
```diff
@@ -412,6 +421,71 @@ void GIVEN_an_invalid_card_WHEN_sending_card_THEN_card_is_rejected() {
...
+                                .title(new I18n("title",null))
+                                .summary(new I18n("summary",null))
+                                .startDate(Instant.now())
+                                .state("state1")
+                                .build();
+
+                Assertions.assertThatCode(
+                                () -> cardProcessingService.processUserCard(childCard, currentUserWithPerimeters,
+                                                token))
+                                .doesNotThrowAnyException();
```
freddidierRTE: Would it be possible to create nested test and factorize in a method the beginning of the tests which are similar ?

services/cards-publication/src/test/java/org/opfab/cards/publication/services/CardProcessServiceShould.java
freddidierRTE: Add a test with parent startDate < publishDate

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6337, opened 2024-04-19
Use new modal service for app-loaded-in-another-tab component (#6157)

### LINKED ISSUE #6157: Use new modal service for app-loaded-in-another-tab component
See https://github.com/opfab/operatorfabric-core/pull/6144 for an example of code

vlo-rte:
Nothing in release notes.

### DIFF
--- a/ui/main/src/app/business/services/modal.service.ts
+++ b/ui/main/src/app/business/services/modal.service.ts
@@ -9,30 +9,22 @@
 
 import {I18n} from '@ofModel/i18n.model';
 import {ModalServer} from '../server/modal.server';
-import {TranslationService} from './translation/translation.service';
 import {ModalConfig} from '@ofModel/modal-config.model';
 
 export class ModalService {
     private static modalServer: ModalServer;
-    private static translationService: TranslationService;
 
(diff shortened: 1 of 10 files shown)


ui/main/src/app/business/services/modal.service.ts
```diff
@@ -52,19 +52,16 @@ export class ModalService {
...
                 {
-                    label: this.translationService.getTranslation('button.cancel'),
+                    label: new I18n('button.cancel'),
                     id: 'cancel'
                 },
                 {
-                    label: this.translationService.getTranslation('button.ok'),
+                    label: new I18n('button.ok'),
                     id: 'ok',
                     isSelected: true
```
freddidierRTE, line 51: Could you do the same for the other methods (using I18n)
And remove all translation feature from the class (and getTranslatedValue method)

freddidierRTE (changes requested):

ui/main/src/app/business/services/modal.service.ts
vlo-rte: Done.

freddidierRTE (approved):

PR #6400, opened 2024-04-29
Add security authorization logs in node services (#6301)

### LINKED ISSUE #6301: Add security authorization logs in node services
For each node services , if a request is refuse to a user not having the admin roles add a security logs with level warn containing the following line : 

"SECURITY : user $user try to access resource  $resourcename without the required authorization"

quinarygio:
Fix #6301 

- In release note :
  -  In chapter : Features
  -  Text : #6301 : Add security authorization logs in node services

### DIFF
--- a/node-services/common/src/common/server-side/jwtTokenUtils.ts
+++ b/node-services/common/src/common/server-side/jwtTokenUtils.ts
@@ -29,7 +29,6 @@ export default class JwtTokenUtils {
     public validateToken(token: string, margin: number): boolean {
         if (token?.length > 0) {
             const jwt = this.decodeToken(token);
-
             if (jwt != null) {
                 const expirationDate = jwt.exp;
                 if (new Date().valueOf() < expirationDate * 1000 - margin) {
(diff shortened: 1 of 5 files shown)


node-services/supervisor/src/supervisor.ts
```diff
@@ -91,8 +92,10 @@ app.get('/start', (req, res) => {
     authorizationService
         .isAdminUser(req)
         .then((isAdmin) => {
-            if (!isAdmin) res.status(403).send();
-            else {
+            if (!isAdmin) {
+                authorizationService.handleUnauthorizedAccess(req, res);
+                res.status(403).send();
```
freddidierRTE: You do not need to send the 403 as it is sent in handleUnauthorizedAccess

Same remark in other parts of the code

node-services/common/src/common/server-side/jwtTokenUtils.ts
```diff
@@ -40,6 +39,17 @@ export default class JwtTokenUtils {
         return false;
     }
 
+    public getUser(token: string | null): string | undefined {
+        let user;
+        if (token != null && token?.length > 0) {
+            const jwt = this.decodeToken(token);
+            if (jwt != null) {
+                user = jwt.sub;
```
freddidierRTE: The default value for the user login is "sub" but it is possible in opfab to use another value define in common.yml  with the field "security.jwt.login-claim".
You need to take that into account

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6404, opened 2024-04-30
Wrong display of pagination in user action logs screen (#6381)

### LINKED ISSUE #6381: Wrong display of pagination in user action logs screen
The pagination should be on one line : 
image

vlo-rte:
- In release note :
  -  In chapter : Bugs
  -  Text : #6381 : Wrong display of pagination in user action logs screen

### DIFF
--- a/ui/main/src/app/modules/useractionlogs/useractionlogs.component.html
+++ b/ui/main/src/app/modules/useractionlogs/useractionlogs.component.html
@@ -87,7 +87,7 @@ font-size: 20px;font-weight: bold;border-bottom: solid 1px;">
                 </table>
             </div>
             <div class="opfab-pagination" style="padding-bottom: 5px; margin-top: 5px">
-                <div style="margin-top: 12px">
+                <div style="margin-top: 12px;white-space: nowrap">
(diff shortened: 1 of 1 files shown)


ui/main/src/app/modules/useractionlogs/useractionlogs.component.html
```diff
@@ -87,7 +87,7 @@
                 </table>
             </div>
             <div class="opfab-pagination" style="padding-bottom: 5px; margin-top: 5px">
-                <div style="margin-top: 12px">
+                <div style="margin-top: 12px;white-space: nowrap">
```
ClementBouvierN, line 90: Combine the two margin-top like such : 
`<div class="opfab-pagination" style="padding-bottom: 5px">
      <div style="margin-top: 17px;white-space: nowrap">`
to be consistent with the other pagination styles

vlo-rte: This is not code introduced by my PR.
But anyway what you propose is wrong because with your suggestion, the display is not good. It is not vertically aligned anymore.

ClementBouvierN (approved):

vlo-rte:
@ClementBouvierN  Add the bug in the relase notes.

PR #6422, opened 2024-05-03
Do not mock opfab API anymore as it erases method that can be used in future unit test

freddidierRTE:


ui/main/src/app/business/buildInTemplates/message-or-question-list/card/message-or-question-listCardTemplateView.spec.ts
```diff
@@ -7,89 +7,78 @@
...
-        opfab.currentCard.getCard = function () {
-            return {data: {title: 'My title <script>'}};
-        };
+        OpfabAPIService.currentCard.card = {data: {title: 'My title <script>'}};
         expect(view.getTitle()).toEqual('My title &lt;script&gt;');
     });
 
     it('Given a card WHEN user is not allowed to response THEN response input is hidden', () => {
-        opfab.currentCard.isUserAllowedToRespond = () => false;
+        OpfabAPIService.currentCard.isUserAllowedToRespond = true;
```
quinarygio: should be isUserAllowedToRespond = false; ?
Anyway the test always pass, so it does not work well.
I think you must also set: 
 OpfabAPIService.currentCard.card = {data: {question: true}};

quinarygio (changes requested):

ui/main/src/app/business/buildInTemplates/message-or-question-list/card/message-or-question-listCardTemplateView.spec.ts
freddidierRTE: Yes, you are right

quinarygio (approved):

PR #6437, opened 2024-05-08
Add filters LESSTHAN & GREATERTHAN to FilterMatchTypeEnum (#6319)

### LINKED ISSUE #6319: Add filters LESS_THAN & GREATER_THAN to org.opfab.cards.model.FilterMatchTypeEnum class
In the case of a search for cards whose **publishDate** is less than a specific date for example. Would it be possible to add the `LESS_THAN` & `GREATER_THAN` filters in order to be comparable ?

quinarygio:
Fix #6319 

- In release note :
  -  In chapter :  Features
  -  Text : #6319 : Add filters LESSTHAN & GREATERTHAN to FilterMatchTypeEnum

### DIFF
--- a/client/cards/src/main/java/org/opfab/cards/model/FilterMatchTypeEnum.java
+++ b/client/cards/src/main/java/org/opfab/cards/model/FilterMatchTypeEnum.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022, RTE (http://www.rte-france.com)
+/* Copyright (c) 2022-2024, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -17,13 +17,15 @@ package org.opfab.cards.model;
  *
  */
 public enum FilterMatchTypeEnum {
-    EQUALS,
-    NOTEQUAL,
-    STARTSWITH,
-    ENDSWITH,
-    CONTAINS,
-    NOTCONTAINS,
     BLANK,
+    CONTAINS,
+    ENDSWITH,
+    EQUALS,
+    GREATERTHAN,
+    IN,
+    LESSTHAN,
     NOTBLANK,
-    IN
+    NOTCONTAINS,
+    NOTEQUAL,
+    STARTSWITH    
 }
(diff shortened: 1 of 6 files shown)


client/cards/src/main/java/org/opfab/cards/model/FilterMatchTypeEnum.java
```diff
@@ -25,5 +25,7 @@ public enum FilterMatchTypeEnum {
     NOTCONTAINS,
     BLANK,
     NOTBLANK,
-    IN
+    IN,
+    LESSTHAN,
+    GREATERTHAN
```
freddidierRTE: I propose to set the order as alphabetical to ease reading

services/cards-consultation/src/main/java/org/opfab/cards/consultation/model/FilterMatchTypeEnum.java
```diff
@@ -21,5 +21,7 @@ public enum FilterMatchTypeEnum {
     NOTCONTAINS,
     BLANK,
     NOTBLANK,
-    IN
+    IN,
+    LESSTHAN,
+    GREATERTHAN
 }
```
freddidierRTE: I propose to set the order as alphabetical to ease reading

services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/UserUtilitiesCommonToCardRepository.java
```diff
@@ -286,13 +284,17 @@ private List<Criteria> getFilterCriteria(CardsFilter filter) {
         filter.filters().forEach(columnFilter -> {
             if (!SPECIAL_PARAMETERS.contains(columnFilter.columnName()) && columnFilter.operation() == null) {
                 // Multiple conditions operations are not supported yet
-                criteria.add(getMatchingCriteria(columnFilter));
+                try {
+                    criteria.add(getMatchingCriteria(columnFilter));
+                } catch (NumberFormatException | NoSuchFieldException | SecurityException e) {
+                    // Ignore criteria on wrong columns
+                }
```
freddidierRTE, line 291: I suggest handling exceptions within each case of the getMatchingCriteria switch statement and adding a meaningful `log.warn`.

 This would make it easier to identify and resolve potential problems.

freddidierRTE (changes requested):

services/cards-consultation/src/main/java/org/opfab/cards/consultation/repositories/UserUtilitiesCommonToCardRepository.java
quinarygio: After discussion it was decided to keep it as it is, because the behaviour is similar to the current implementation: if a wrong column is specified in the condition than the condition is ignored.

freddidierRTE (approved):

PR #6443, opened 2024-05-10
Add karate end to end test for reminder feature (#6385)

### LINKED ISSUE #6385: End to end test for reminder feature
Karate test waiting for reminder ...

quinarygio:
Fix #6385 

- In release note :
  -  In chapter :  Tasks
  -  Text : #6385 : Add karate end to end test for reminder feature

### DIFF
--- a/src/test/api/karate/cards/cardsReminder.feature
+++ b/src/test/api/karate/cards/cardsReminder.feature
@@ -32,6 +32,104 @@ Feature: CardsReminder
     ]
     """
 
+    
+  * def entity1Array =
+      """
+      [   "ENTITY1_FR"
+      ]
+      """
+
+            * def entity2Array =
+      """
+      [   "ENTITY2_FR"
+      ]
+      """
+
+Scenario: Send card and receive reminder
+
+
+#Create new perimeter
+* call read('../common/createPerimeter.feature') {perimeter: '#(perimeter)', token: '#(authTokenAdmin)'}
+
+#Attach perimeter to group
+    Given url opfabUrl + 'users/groups/ReadOnly/perimeters'
+    And header Authorization = 'Bearer ' + authTokenAdmin
+    And request perimeterArray
+    When method patch
+    Then status 200
+
+
+    * def currentDate = new Date()
+    * def now = new Date().valueOf()
+    * def yesterday = new java.math.BigDecimal(now - 8640000)
+    * def tomorrow = new java.math.BigDecimal(now + 8640000)
+
+    * def datePlus2Min = new Date(now + 2*60*1000).getTime();
+    * def hours = new java.lang.Integer(new Date(datePlus2Min).getHours())
+    * def minutes = new java.lang.Integer(new Date(datePlus2Min).getMinutes())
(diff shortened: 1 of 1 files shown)


src/test/api/karate/cards/cardsReminder.feature
```diff
@@ -32,46 +32,122 @@ Feature: CardsReminder
...
+    Then status 200
+
+
+    * def currentDate = new Date()
+    * def now = new Date().valueOf()
+    * def yesterday = new java.math.BigDecimal(now - 8640000)
+    * def tomorrow = new java.math.BigDecimal(now + 8640000)
+    * def hours = new java.lang.Integer(new Date().getHours())
+    * def minutes = new java.lang.Integer(new Date().getMinutes() + 2)
+    * def seconds = new java.lang.Integer(new Date().getSeconds())
```
freddidierRTE: I think it s gone to create problems particularly when the current time is close to the end of the hour (e.g., 10:59). In such cases, adding 2 minutes would result in a time with 61 minutes, which is not valid.

I suggest to first create a date + 2 min ( new Date().valueOf() + `2*60*1000` )
and then extract hours/minutes/seconds from this new object

src/test/api/karate/cards/cardsReminder.feature
```diff
@@ -32,46 +32,122 @@ Feature: CardsReminder
...
+      """
+      [   "ENTITY2_FR"
+      ]
+      """
+
+Scenario: Send card and receive reminder
+
+
+#Create new perimeter
+* callonce read('../common/createPerimeter.feature') {perimeter: '#(perimeter)', token: '#(authTokenAdmin)'}
```
freddidierRTE: i realize that we are using every where `callonce` , but we can use `call` instead in a lot of place (callonce just cache the result , so it is interesting when getting token)

freddidierRTE (changes requested):

src/test/api/karate/cards/cardsReminder.feature
```diff
@@ -32,46 +32,123 @@ Feature: CardsReminder
...
+
+
+    * def currentDate = new Date()
+    * def now = new Date().valueOf()
+    * def yesterday = new java.math.BigDecimal(now - 8640000)
+    * def tomorrow = new java.math.BigDecimal(now + 8640000)
+
+    * def hours = new java.lang.Integer(new Date().getHours())
+    * def minutes = new java.lang.Integer(new Date(now + 2*60*1000).getMinutes())
+    * def seconds = new java.lang.Integer(new Date().getSeconds())
```
freddidierRTE: As for minutes it has to be the same for hours otherwise 9:59 will ends up to 9:01 


```suggestion
    * def datePlus2Min = new java.lang.Integer(new Date(now + 2*60*1000))
    * def hours = new java.lang.Integer(datePlus2Min.getHours())
    * def minutes = new java.lang.Integer(datePlus2Min.getMinutes())
    * def seconds = new java.lang.Integer(datePlus2Min.getSeconds())
 
```

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6482, opened 2024-05-21
Link the period selection for the Dashboard, Card Feed and Monitoring screens (#6463)

### LINKED ISSUE #6463: Link the period selection for the Dashboard, Card Feed and Monitoring screens
Hello OperatorFabric's colleagues,

We noticed that currently, the period selection of the different screens having the timeline (Dashboard, Card Feed and Monitoring) is not linked.

We expected that this period selection is the same for the 3 screens so once I change the period in one screen I keep it by navigating to the other(s) screen(s).

I remain available to discuss it further.

Kind regards,

quinarygio:
Fix #6463

- In release note :
  -  In chapter :  Features 
  -  Text : #6463 : Link the period selection for the Dashboard, Card Feed and Monitoring screens

### DIFF
--- a/src/test/cypress/cypress/integration/Translation.spec.js
+++ b/src/test/cypress/cypress/integration/Translation.spec.js
@@ -436,8 +436,7 @@ describe('Test translations', function () {
 
     it('Check Business period translations', function () {
         const currentDate = new Date(2020, 11, 31, 23, 46); // Date must be in the past to avoid session close with token expiration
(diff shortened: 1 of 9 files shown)


freddidierRTE:
- On monitoring screen using << and >> does not refresh results 
- When loading or reloading, lock is always activated

freddidierRTE (approved):

PR #6494, opened 2024-05-22
Allowed to delete admin group (#6490)

### LINKED ISSUE #6490: Permit to delete admin group
In previous version of opfab it was necessary to be in the group admin to be admin , now you just need to be in a group having admin permissions 

A feature had been develop to avoid removing admin group , this feature is not needed anymore as we do not necessarily need an admin group .

ClementBouvierN:
- In release note :
  -  In chapter :  Features
  -  Text : #6490 Possibility to delete Admin group

### DIFF
--- a/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
+++ b/ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
@@ -91,6 +91,7 @@ export abstract class AdminTableDirective implements OnDestroy {
 
     protected i18NPrefix = 'admin.input.';
     protected crudService: CrudService;
+    private currentUserLogin;
 
     constructor(
         protected translateService: TranslateService,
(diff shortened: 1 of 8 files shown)


ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -201,18 +201,9 @@ export abstract class AdminTableDirective implements OnDestroy {
             columnDefs[index] = columnDef;
         });
 
-        const deleteActionCellClassRules = {
-            'action-cell-delete-admin': (params) =>
```
quinarygio, line 205: It is possible to remove the css class "action-cell-delete-admin" from styles.scss as it not used anymore

quinarygio (changes requested):

ui/main/src/app/modules/admin/components/table/admin-table.directive.ts
```diff
@@ -201,12 +203,11 @@ export abstract class AdminTableDirective implements OnDestroy {
             columnDefs[index] = columnDef;
         });
 
+        console.log('CURRENT USER LOGIN ', this.currentUserLogin);
```
quinarygio: remove this debug console log

quinarygio:
In admin users screen when logged as operator5_fr, who has admin rights, there is still the delete button visible for operator5_fr user 
image

quinarygio (changes requested):

quinarygio (approved):

PR #6500, opened 2024-05-23
Fix logout procedure in IMPLICIT mode (#6480)

### LINKED ISSUE #6480: Logout functionnality in implicit mode is not working properly
In Implicit mode, the logout process is handled by the Angular OIDC library. This approach conflicts with the standard OperatorFabric logout procedure, where a logout URL is provided for logout and redirection (as set in web-ui.json under security.logout-url).

The attached screenshot illustrates that the logout URL is invoked but subsequently cancelled by the standard OperatorFabric logout procedure:

Issue Screenshot

The issue at hand is to correctly configure the logout process in Implicit mode using the Angular OIDC library. 

It appears feasible to provide the library with a redirect URL. The proposed solution is to utilize security.logout-url as the redirect URL in Implicit mode."

quinarygio:
Fix #6480

- In release note :
  -  In chapter :  Bugs 
  -  Text : #6480 : Fix logout procedure in IMPLICIT mode

### DIFF
--- a/src/docs/asciidoc/deployment/configuration/security_configuration.adoc
+++ b/src/docs/asciidoc/deployment/configuration/security_configuration.adoc
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2023 RTE (http://www.rte-france.com)
+// Copyright (c) 2018-2024 RTE (http://www.rte-france.com)
 // See AUTHORS.txt
 // This document is subject to the terms of the Creative Commons Attribution 4.0 International license.
 // If a copy of the license was not distributed with this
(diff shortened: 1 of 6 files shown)


src/docs/asciidoc/deployment/configuration/security_configuration.adoc
```diff
@@ -143,6 +143,9 @@ To enable IMPLICIT Flow authentication the following properties need to be set i
 - `security.oauth2.flow.mode` to `IMPLICIT`;
 - `security.oauth2.flow.delegate-url` with the URL of the OAuth2 leading to the `.well-known/openid-configuration` end-point used for authentication configuration.
 
+It is possible to configure the URL to redirect the browser after logout by settin the `security.implicit-mode-post-logout-url` parameter in web-ui.json.
```
freddidierRTE: by settin**g**

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6540, opened 2024-05-31
Upgrade to springboot 3.3.0

### LINKED ISSUE #6535: Update spring boot to v3.3.0
Build is not passing with springboot 3.3.0 
See https://github.com/opfab/operatorfabric-core/pull/6515

quinarygio:
After upgrade to springboot 3.3.0 the jackson serialization of records is failing, it is due to the specific objectMapper setting in CardObjectMapper class:  
`this.objectMapper = JsonMapper.builder().enable(MapperFeature.REQUIRE_SETTERS_FOR_GETTERS).build()`

This setting is needed to exclude some avro specific fields from JSON serialization


services/cards-publication/src/main/java/org/opfab/cards/publication/kafka/CardObjectMapper.java
```diff
@@ -31,8 +33,16 @@ public class CardObjectMapper {
...
+        this.objectMapper = JsonMapper.builder().build();
 
+        objectMapper.setAnnotationIntrospector(new JacksonAnnotationIntrospector() {
+            @Override
+            public boolean hasIgnoreMarker(final AnnotatedMember m) {
+                List<String> exclusions = Arrays.asList("getSchema", "getSpecificData");
+                return exclusions.contains(m.getName()) || super.hasIgnoreMarker(m);
+            }
+        });
+
```
freddidierRTE: Move the exclusions list to a constant to avoid creating it every time hasIgnoreMarker is called  -->

`private static final List<String> EXCLUSIONS = Arrays.asList("getSchema", "getSpecificData");`

freddidierRTE: Add a comment to explain why we do that

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6601, opened 2024-06-10
Refactor sound and system notification rules (#6532)

### LINKED ISSUE #6532: Usercard: when sending "Confirmation" user card a sound notification is played while it should not
To reproduce:
 Activate sound for "ACTION" severity in user settings.
Send a new Usercard examples- Confirmation usercard.
A sound notification is played because when sending Confirmation usercard, a child response card is also sent automatically.  The child card triggers an update of the parent card (to set hasChildCardFromCurrentUserEntity = true) on lightCardsStore, so the sound notification service receives the parent card twice. the first time the sound is not played because card id equals lastSentCardId, the second time instead the sound is played.

quinarygio:
Fix #6532

- In release note :
  -  In chapter :  Bugs
  -  Text : #6532 : Fix bug to avoid sound notification is played when it should not

### DIFF
--- a/ui/main/src/app/business/services/notifications/notification-decision.spec.ts
+++ b/ui/main/src/app/business/services/notifications/notification-decision.spec.ts
@@ -1,0 +1,177 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {Severity} from '@ofModel/light-card.model';
+import {NotificationDecision} from './notification-decision';
+import {getOneLightCard} from '@tests/helpers';
+
+describe('Sound decisions', () => {
+    beforeEach(() => {
+        jasmine.clock().install();
+        jasmine.clock().mockDate(new Date(100));
+        NotificationDecision.init();
+    });
+
+    afterEach(() => {
+        jasmine.clock().uninstall();
+    });
+
+    describe('Check sound and system notification', () => {
+        it('Sound and system notification for new card ', (done) => {
+            const publishDate = new Date().getTime();
+            const card = getOneLightCard({publishDate: publishDate});
+            NotificationDecision.setLastUserAction(publishDate - NotificationDecision.ERROR_MARGIN);
+            expect(NotificationDecision.isSoundToBePlayedForCard(card)).toBeTrue();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card)).toBeTrue();
+            done();
+        });
+
+        it('No sound and system notification for cards recently sent by the user', (done) => {
+            const publishDate = new Date().getTime();
+            const card1 = getOneLightCard({publishDate: publishDate});
+            const card2 = getOneLightCard({publishDate: publishDate});
+            NotificationDecision.setLastUserAction(publishDate - NotificationDecision.ERROR_MARGIN);
+            NotificationDecision.addSentCard(card1.id);
+            NotificationDecision.addSentCard(card2.id);
+            expect(NotificationDecision.isSoundToBePlayedForCard(card1)).toBeFalse();
+            expect(NotificationDecision.isSoundToBePlayedForCard(card2)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card1)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card2)).toBeFalse();
+            done();
+        });
+
+        it('No sound and no system notification if card has been read', (done) => {
+            const publishDate = new Date().getTime();
+            const card = getOneLightCard({
+                hasBeenRead: true,
+                publishDate: publishDate
+            });
+            NotificationDecision.setLastUserAction(publishDate - NotificationDecision.ERROR_MARGIN);
+
+            expect(NotificationDecision.isSoundToBePlayedForCard(card)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card)).toBeFalse();
+
+            done();
+        });
+
+        it('No sound and no system notification if card is too old', (done) => {
+            const publishDate = new Date().getTime();
+            const card = getOneLightCard({
+                hasBeenRead: false,
+                publishDate: publishDate
+            });
+            jasmine.clock().tick(NotificationDecision.RECENT_THRESHOLD + NotificationDecision.ERROR_MARGIN);
+            expect(NotificationDecision.isSoundToBePlayedForCard(card)).toBeFalse();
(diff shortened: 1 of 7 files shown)


ui/main/src/app/business/services/notifications/notification-decision.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static checkCardIsRecent(card: LightCard): boolean {
+        return new Date().getTime() - card.publishDate <= NotificationDecision.RECENT_THRESHOLD;
+    }
+
+    private static cleanSentCards() {
+        const now = new Date().valueOf();
+        const toRemove = [];
+        console.log('cleanSentCards size ' + NotificationDecision.lastSentCards.size);
```
freddidierRTE: Log to be removed

ui/main/src/app/business/services/notifications/notification-decision.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.ALARM)).toBeTrue();
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.ACTION)).toBeTrue();
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.COMPLIANT)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.INFORMATION)).toBeFalse();
+            NotificationDecision.setSystemNotificationEnabledForSeverity(Severity.ACTION, false);
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.ACTION)).toBeFalse();
+            done();
+        });
+
+        it('Play sound for session end', (done) => {
```
freddidierRTE: ' .. for session end if at least one severity is activated'

ui/main/src/app/business/services/notifications/notification-decision.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        });
+
+        it('System notification and NO sound for card sent before last user action', (done) => {
+            const publishDate = new Date().getTime() - TIME_MARGIN;
+            const card = getOneLightCard({publishDate: publishDate});
+            NotificationDecision.setLastUserAction(new Date().getTime());
+            expect(NotificationDecision.isSoundToBePlayedForCard(card)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card)).toBeTrue();
+            done();
+        });
```
freddidierRTE, line 82: Add a test , no sound or notification if card is too old

ui/main/src/app/business/services/notifications/notification-decision.ts
```diff
@@ -0,0 +1,139 @@
...
+        NotificationDecision.lastSentCards.forEach((v, k) => {
```
freddidierRTE: Suggestion: instead of using v,k , add more meaningfull name , for example : (timestamp,key)

ui/main/src/app/business/services/notifications/notification-decision.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+            NotificationDecision.setSystemNotificationEnabledForSeverity(Severity.ACTION, false);
+            expect(NotificationDecision.isSystemNotificationEnabledForSeverity(Severity.ACTION)).toBeFalse();
+            done();
+        });
+
+        it('Play sound for session end', (done) => {
+            expect(NotificationDecision.isPlaySoundWhenSessionEndEnabled()).toBeFalse();
+            NotificationDecision.setSoundEnabledForSeverity(Severity.INFORMATION, true);
+            expect(NotificationDecision.isPlaySoundWhenSessionEndEnabled()).toBeTrue();
+            done();
```
freddidierRTE, line 164: Coverage reports (/ui/main/reports/coverage/src/app/business/services/notifications/notification-decision.ts.html) shows a few line not tested : 

you should add tests : 
-  for methods isAtLeastOneSoundActivated and isAtLeastOneSystemNotificationSeverityEnabled() 
-  for cleanSendCard : mock time and check that after a while the card is sounded even if user has send the card (this scenario can arise if the card is updated by another user).

ui/main/src/app/business/services/notifications/notification-decision.ts
```diff
@@ -0,0 +1,139 @@
...
+        if (this.lastSentCards.get(card.id))
+            return false; // no sound as the card was sent by the current user
+        else {
+            return (
+                !card.hasBeenRead &&
+                this.checkCardHasBeenPublishAfterLastUserAction(card) &&
+                this.checkCardIsRecent(card)
+            );
+        }
+    }
```
freddidierRTE: The actual implmentation does not solve this use case :

The user send a card and does not read it 
After a while a user of the same entity send a reponse , the sound will be triggered 

To avoid that we need to keep last send card in memory for at least RECENT_THRESHOLD 

And when we received the card : 
- if card is read --> No sound 
- if card is too old --> No Sound 
- if card is not in last send --> Sound 
- if card is in last send and publish Date is close to last send date --> No sound 

I proposed to use ERROR_MARGIN to set it is close to

freddidierRTE (changes requested):

ui/main/src/app/business/services/notifications/notification-decision.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+        it('System notification and NO sound for card sent before last user action', (done) => {
+            const publishDate = new Date().getTime() - NotificationDecision.ERROR_MARGIN;
+            const card = getOneLightCard({publishDate: publishDate});
+            NotificationDecision.setLastUserAction(new Date().getTime());
+            expect(NotificationDecision.isSoundToBePlayedForCard(card)).toBeFalse();
+            expect(NotificationDecision.isSystemNotificationToBeShownForCard(card)).toBeTrue();
+            done();
+        });
+
+        it('Sound and system notification if card is in lastSentCards and publish date is after sent timestamp plus error margin', (done) => {
```
freddidierRTE, line 84: Add a comment to explain : 

"This use case arises when a different user updates a card originally created by the current user.
We detect this scenario by verifying that the card's publish date is later than the date when the current user sent the card."

ui/main/src/app/business/services/notifications/notification-decision.ts
```diff
@@ -0,0 +1,141 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {LightCard, Severity} from '@ofModel/light-card.model';
+
+export class NotificationDecision {
+    static readonly RECENT_THRESHOLD = 18000000; // in milliseconds , 30 minutes
```
freddidierRTE, line 16: Add a comment : 
"We allow a 30-minute window for a card to be considered 'recent'. This accounts for potential network issues 
that might delay the card's delivery. For instance, a notification might sound 20 minutes after the card's publication.

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6646, opened 2024-06-14
Opfab-cli : Add command to get/set logs level for java service (#6528)

### LINKED ISSUE #6528: Opfab-cli  : Add command to get/set  logs level for java service
Create command similar to 
src/test/resources/logs/setLogs.sh & src/test/resources/logs/getLogs.sh

command to get logs :

`opfab service servicename get-log-level`

command to set logs: 

`opfab service servicename set-log-level level`
 

- If the service name is not provided user shall choose in a pre-defined list 
- if the level is not provided user shall choose in a pre-defined list 

Add help command 

Requests should be done via nginx (port 2002)

vlo-rte:
Nothing in release notes.

### DIFF
--- a/cli/serviceCommands.js
+++ b/cli/serviceCommands.js
@@ -1,0 +1,199 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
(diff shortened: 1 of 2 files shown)


cli/serviceCommands.js
```diff
@@ -0,0 +1,179 @@
...
+        }
+        return serviceName;
+    },
+
+    async printHelp() {
+        console.log(`Usage:
+opfab service get-log-level <serviceName>
+or
+opfab service set-log-level <serviceName> <level>
+        `);
```
freddidierRTE: - To be consistent with other services , the content shall be similar to printHelp() of cli/cardCommands.js
- Add the list of possible serviceName  and possible log level 

This gives something like : 

```
Comand list : 
        get-log-level ....

serviceName : card-consultation , card ... 

level : TACE, DEBUG ....


```

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6703, opened 2024-06-25
Add new endpoints to set a process / state as notified/not notified for all user (#6660)

### LINKED ISSUE #6660: Add new endpoints to set a process / state as notified/not notified for all user
We need to be able to create a new process/state having no user receiving the process/state by default 

To do that the idea is to provide an endpoint to set ProcessStatesNotNotified for all users with one request 

The end point will be from the service point 
 - to remove notification for all users : DELETE /notificationConfiguration/processStateNotified
 - to add notification for all users : POST /notificiationConfiguration/processStateNotified

The body shall contains the processId and the state to modify

The modification will address only the existing user settings , it a user has no settings, no settings is to be created

Should be tested via karate tests

Add a command line in cli to access this endpoint , the endpoint shall be access via nginx (port 2002)

```
opfab users notif-config  set-notified <process> <state>
opfab users notif-config  set-not-notified <process> <state>
```

The same is to be done for ProcessesStatesNotifiedByEmail

End point : /notificationConfiguration/processStateNotifiedByMail
cli  :
opfab users notif-config  set-not-notified-mail ... 
opfab users notif-config  set-notified-mail

quinarygio:
Fix #6660

- In release note :
  -  In chapter :  Features
  -  Text : #6660 : Add new endpoints to set a process / state as notified/not notified for all user

### DIFF
--- a/cli/src/usersCommands.js
+++ b/cli/src/usersCommands.js
@@ -1,0 +1,124 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+const config = require('./configCommands.js');
+const prompts = require('prompts');
+
+const usersCommands = {
+    async processUsersCommand(args) {
+        let action = args[0];
+        if (!action) {
+            action = (
+                await prompts({
+                    type: 'select',
+                    name: 'value',
+                    message: 'Users action',
+                    choices: [
+                        {title: 'Set notified', value: 'set-notified'},
+                        {title: 'Set not notified', value: 'set-not-notified'},
+                        {title: 'Set notified by mail', value: 'set-notified-mail'},
+                        {title: 'Set not notified by mail', value: 'set-not-notified-mail'}
+                    ]
+                })
+            ).value;
+            if (!action) {
+                console.log('Users action is required');
+                return;
+            }
+        }
+
+        switch (action) {
+            case 'set-notified':
+                await this.configureNotification(args[1], args[2], 'POST', 'processstatenotified');
+                break;
+            case 'set-not-notified':
+                await this.configureNotification(args[1], args[2], 'DELETE', 'processstatenotified');
+                break;
+            case 'set-notified-mail':
+                await this.configureNotification(args[1], args[2], 'POST', 'processstatenotifiedbymail');
+                break;
+            case 'set-not-notified-mail':
+                await this.configureNotification(args[1], args[2], 'DELETE', 'processstatenotifiedbymail');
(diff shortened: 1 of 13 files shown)


cli/src/usersCommands.js
```diff
@@ -0,0 +1,116 @@
...
+        if (response.ok) {
+            console.error('Success');
+        } else {
+            console.error('Failed to configure process/state notification for ' + process + '/' +state);
+            console.error('Response:', response);
+        }
+    },
+
+    async printHelp() {
+        console.log(`Usage: opfab users set-notif <process> <state>`);
```
freddidierRTE: Explain all possible set-* values :

``` 
Usage opfab users <command> <process> <state>

Commands list : 
        set-notified                set a process and state nofified for all users 
        set ...

```

You can have an example with "opfab bundle help"

services/users/src/main/java/org/opfab/users/services/UserSettingsService.java
```diff
@@ -176,4 +176,71 @@ public OperationResult<UserSettings> updateUserSettings(String login, UserSettin
...
+            }
+        });
+    }
+
+    public void unsetProcessStateNotified(String process, String state) {
+        log.info("unsetProcessStateNotified for {} {}", process, state);
+        List<UserSettings> settings = userSettingsRepository.findAll();
+
+        settings.forEach(userSettings -> {
+            if (userSettings.getProcessesStatesNotNotified() != null && userSettings.getProcessesStatesNotNotified().get(process) != null) {
```
freddidierRTE: here you add an entry in the repository so even if getProcessStatsNotNotified() is null , you need to add the state as not notified 

The same if userSettings.getProcessesStatesNotNotified().get(process) is null , you need to add it anyway

services/users/src/main/java/org/opfab/users/services/UserSettingsService.java
```diff
@@ -176,4 +176,71 @@ public OperationResult<UserSettings> updateUserSettings(String login, UserSettin
...
+            }
+        });
+    }
+
+    public void setProcessStateNotifiedByEmail(String process, String state) {
+        log.info("setProcessStateNotifiedByEmail for {} {}", process, state);
+        List<UserSettings> settings = userSettingsRepository.findAll();
+
+        settings.forEach(userSettings -> {
+            if (userSettings.getProcessesStatesNotifiedByEmail() != null && userSettings.getProcessesStatesNotifiedByEmail().get(process) != null) {
```
freddidierRTE: Same remarks as the one on unsetProcessStateNotified

freddidierRTE (changes requested):

freddidierRTE:
In documentation (https://opfab.github.io/documentation/current/deployment/#_users) add a chapter "Notification configuration" with a small text explaining that we can set notification configuration for all user at once using the cli , do not explain cli usage just refer to "opfab users help" command.

freddidierRTE:
a last remark , this is two sonar issues to solve : https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=6703&id=org.lfenergy.operatorfabric%3Aoperatorfabric-core

freddidierRTE (approved):

PR #6755, opened 2024-07-09
In built-in question usercard add option to keep response history (#6670)

### LINKED ISSUE #6670: In built-in question keep response history
Each time user is responding append the response message to the existing one 

So on the screen , you will see response as follow 

```
Entity 1    Date1 - response text1 
            Date2 - response text2 
```
This shall be an option defined in the parameter of the built-in template  with value "keepResponseHistoryInCard" 

Add a cypress test

quinarygio:
Fix #6670

- In release note :
  -  In chapter :  Features
  -  Text : #6670 : In built-in question usercard add option to keep response history

### DIFF
--- a/ui/main/src/app/business/builtInTemplates/question/card/questionCardTemplate.ts
+++ b/ui/main/src/app/business/builtInTemplates/question/card/questionCardTemplate.ts
@@ -44,8 +44,10 @@ export class QuestionCardTemplate extends HTMLElement {
     }
 
     private setFunctionToGetUserResponseInput() {
+        const keepResponseHistoryInCard = this.getAttribute('keepResponseHistoryInCard');
         this.questionCardTemplateView.setFunctionToGetResponseInput(
-            () => (<HTMLInputElement>document.getElementById('template_response_input')).value
+            () => (<HTMLInputElement>document.getElementById('template_response_input')).value,
+            keepResponseHistoryInCard?.toLocaleLowerCase() === 'true'
         );
     }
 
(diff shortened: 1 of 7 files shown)


ui/main/src/app/business/builtInTemplates/question/card/questionCardTemplate.ts
```diff
@@ -55,15 +58,27 @@ export class QuestionCardTemplate extends HTMLElement {
...
+                    html += `<table width="100%" style="border-bottom:none">`;
+                    response.responses.forEach((resp) => {
+                        html +=
+                            ' <tr style="background-color: initial;"><td>' +
+                            moment(resp.responseDate).format('HH:mm DD/MM/yyyy') +
+                            '</td>';
+                        html += '<td>' + resp.response + '</td> </tr>';
+                    });
+                    html += `</table></td>`;
+                    html += ' </tr>';
```
freddidierRTE: It would be better to have the logic as much as possible in the view so put in the response object the already formatted field : entityName, responseDate 

And add a test in questionCardTemplateView.spec.ts to check responseDate and entityDate

ui/main/src/app/business/builtInTemplates/question/card/questionCardTemplate.ts
```diff
@@ -55,15 +58,27 @@ export class QuestionCardTemplate extends HTMLElement {
...
+                        opfab.utils.escapeHtml(opfab.users.entities.getEntityName(response.entityId)) +
+                        '</td>';
+                    html += ' <td>';
+                    html += `<table width="100%" style="border-bottom:none">`;
+                    response.responses.forEach((resp) => {
+                        html +=
+                            ' <tr style="background-color: initial;"><td>' +
+                            moment(resp.responseDate).format('HH:mm DD/MM/yyyy') +
+                            '</td>';
+                        html += '<td>' + resp.response + '</td> </tr>';
```
freddidierRTE: I think the resp.response field need to be escape (as it could contains malicious code)

freddidierRTE (changes requested):

freddidierRTE:
image

If we have space or not in the text response we may have a "tab" inserted

quinarygio:
> image
> 
> If we have space or not in the text response we may have a "tab" inserted

I was not able to reproduce it

freddidierRTE:
i still reproduce non alignment in the table , as you do not reproduce it i try to solve it on my computer .

Adding `style="width:150px"`  it the "td" of the response date solve the problem  : 

image

freddidierRTE:
Trying to solve the alignment pb , i saw that the date was saved as formatted text in the data of the child card . Could you store it as an epoch date . (This is to be consistent with all storage of dates) 

Sorry for this late review remark

freddidierRTE (approved):

PR #6763, opened 2024-07-10
Set default settings in mongoDB (#6715)

### LINKED ISSUE #6715: Set default settings in mongoDB
When fetchUserSettings and settings does not exist in database , insert an empty settings and return it 

This issue has two objectives : 
 -  avoid 404 error in the console 
 - permit to modify notification notification even if user has not modify his settings when calling new endpoint  /notificationConfiguration/

quinarygio:
Fix #6715 

- In release note :
  -  In chapter :  Features
  -  Text : #6715 : Set default settings in mongoDB


services/users/src/test/java/org/opfab/users/stubs/UserSettingsRepositoryStub.java
```diff
@@ -31,8 +31,11 @@ public UserSettings save(UserSettings userSettings) {
     @Override
     public Optional<UserSettings> findById(String id) {
         UserSettings settings = userSettingsStorage.get(id);
-        if (settings == null)
-            return Optional.empty();
+        if (settings == null) {
+            settings = new UserSettings();
+            settings.setLogin(id);
+            userSettingsStorage.put(id, settings);
+        }
```
freddidierRTE: The provided stub does not behave as the findBy shall behave (do not created an entry if not existing) 

I propose you the following code 
```
  if (settings==null) return Optional.ofNullable(null);
```

services/users/src/main/java/org/opfab/users/services/UserSettingsService.java
```diff
@@ -64,6 +64,7 @@ public OperationResult<UserSettings> patchUserSettings(String login, UserSetting
 
         UserSettings newSettings = oldSettings.patch(userSettingsPatch);
         userSettingsRepository.save(newSettings);
+        log.info("Pathed user settings " + login + " " + newSettings.getEntitiesDisconnected());
```
freddidierRTE: I guess this logs where needed for debugging , i do not think there are needed  ?

src/test/api/karate/users/fetchUserSettings.feature
```diff
@@ -34,8 +34,8 @@ Feature: fetch user settings
...
 
   Scenario: fetch non-existent user
-    #Fetching details about the admin user that doesn't exist, expected response 404
+    #Fetching details about the admin user that doesn't exist, expected response 200
     Given url opfabUrl + 'users/users/fakeuser'+'/settings'
     And header Authorization = 'Bearer ' + authToken
     When method get
-    Then status 404
\ No newline at end of file
+    Then status 200
```
freddidierRTE: Having a look to the test , i realize that we need to distinguish two use cases  : 

- the user exist but has no settings  --> we create the settings and it return 200 
- the user does not exist -->  we should return 404 

The second use case implies a modification in the code to check user existence before creating new settings 
and a new test in services/users/src/test/java/org/opfab/users/services/UserSettingsServiceShould.java

For the karate test it is enough to have only the test with no existing user as the test with existing user with no settings is in unit test

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #6795, opened 2024-07-16
opfab-cli : permit to load more than one bundle using wildcards (#6695)

### LINKED ISSUE #6695: opfab-cli : permit to load more than one bundle using wildcards
Have a look to  cli/src/perimeterCommands.js for a code example

quinarygio:
Fix #6695

Nothing in release notes

### DIFF
--- a/cli/src/bundleCommands.js
+++ b/cli/src/bundleCommands.js
@@ -46,9 +46,9 @@ const bundleCommand = {
     },
 
     async loadBundle(args) {
-        let bundleDirectory = args[0];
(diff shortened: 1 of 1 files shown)


freddidierRTE:
1) When using prompt it does not work 

To reproduce : 

opfab bundle 
Choose load option 

--> the program finishes without asking for directory name 

image


2) Add in help three dots to show that we can input more than one file 
```
load     Load bundle from directory: opfab bundle load <bundleDirectory>  ...
```

freddidierRTE (approved):

PR #6898, opened 2024-07-24
Add a karate test for card deletion with kafka (#6782)

### LINKED ISSUE #6782: Add a karate test for card deletion with kafka

quinarygio:
Fix #6782 

Nothing in release note


tools/kafka-test/src/main/java/org/opfab/kafka/OpfabKafkaProducer.java
```diff
@@ -81,6 +105,28 @@ private List<String> getListFromCvsString(String input) {
...
+            case "ACTION":
+                severityType = SeverityType.ACTION;
+                break;
+            case "ALARM":
+                severityType = SeverityType.ALARM;
+                break;
+            default:
+                severityType = SeverityType.INFORMATION;
+                break;
+        }
```
ClementBouvierN: In this switch you can remove case INFORMATION since it's the same as default case

ClementBouvierN (approved):

PR #6958, opened 2024-07-31
Use HTML in tooltip component (#6826)

### LINKED ISSUE #6826: Use HTML in tooltip component
According to old issue [Add a tooltip component for template or external apps #4518](https://github.com/opfab/operatorfabric-core/issues/4518).

We want to use HTML (using of _Handlebars_ of course) in order to have specific content within tooltip in order to apply it in this example 
image

**The deadline for this feature is early to mid-September.**

ClementBouvierN:
- In release note :
  -  In chapter :  Features
  -  Text : #6826 Use HTML in tooltip component

### DIFF
--- a/src/docs/asciidoc/reference_doc/template_description.adoc
+++ b/src/docs/asciidoc/reference_doc/template_description.adoc
@@ -601,12 +601,16 @@ where the tooltip will be displayed using `left`, `top` or `bottom` (by default,
 example :
 
 ....
-  <span opfab-tooltip-text="Here is an example of tooltip" class="opfab-tooltip">Some tooltip text</span>
+<div class="opfab-tooltip">Some tooltip text
+    <div class="opfab-tooltip-content">Here is an example of tooltip</div>
+</div>
 ....
 
-or for a tooltip displyed on the left :
+or for a tooltip displayed on the left :
 ....
-  <span opfab-tooltip-text="Here is an example of tooltip" class="opfab-tooltip left">Some tooltip text</span>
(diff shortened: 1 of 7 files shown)


vlo-rte (changes requested):
It doesn't work. There must be an arrow between the text and the bubble.

image

vlo-rte:
It doesn't work here either, I don't see any tooltip.

image

vlo-rte:
After a second code review, I still don't see any tooltip : 

image

vlo-rte:
In day mode, it doesn't work : 

image

vlo-rte:
Tooltip on top is not good : 

image

vlo-rte:
Tooltip on bottom is not good : 

image

vlo-rte (commented):
There are still problems.

vlo-rte (changes requested):
There are still problems.

vlo-rte (changes requested):
This is not good. You've changed the colors of opfab.

image

vlo-rte (changes requested):
In day mode, the tooltip is no longer the same color as before.

vlo-rte (changes requested):
1. Still doesn't work : 

image


2. Write a note in the migration guide to tell users that now the positioning is on opfab-tooltip-content and not opfab-tooltip any more.

vlo-rte (changes requested):
I had asked to write in the migration guide that the positioning (left/top/bottom) is not on the same class than previously : 
image

vlo-rte (approved):

PR #7012, opened 2024-08-09
Add card publisher entity in email body (#6926)

### LINKED ISSUE #6926: Add entity publisher in email
In email notification add the entity publisher name if the publisher is an entity.

This is not to be done for daily mail 

The mail content will be for instance : 

``` You received a card in opfab : [Electricity consumption forecast - Message received](http://localhost:2002/#/feed/cards/defaultProcess.process5)

The card has been sent by French Control Center 

This email has been sent by Opfab, there is no need to reply. 
``` 

To do that we need to have access to entity names in the cards-diffusion services . 

This is to be done via a request to the user service (using a cache mechanism to avoid accessing the users service for each card to send) 

**To be reviewed by freddidierRTE**

quinarygio:
Fix #6926

- In release note :
  -  In chapter :  Features
  -  Text : #6926: Add card publisher entity in email body

### DIFF
--- a/config/docker/cards-external-diffusion.yml
+++ b/config/docker/cards-external-diffusion.yml
@@ -11,6 +11,7 @@ operatorfabric:
       subjectPrefix: 'Opfab card received '
       bodyPrefix: 'You received a card in opfab : '
       bodyPostfix: 'This email has been sent by Opfab, there is no need to reply.'
+      publisherEntityPrefix: 'The card has been sent by '
       opfabUrlInMailContent: http://localhost:2002
 #  logConfig:
 #    logLevel: debug
--- a/node-services/cards-external-diffusion/config/default-docker.yml
+++ b/node-services/cards-external-diffusion/config/default-docker.yml
@@ -25,6 +25,7 @@ operatorfabric:
       subjectPrefix: 'Opfab card received '
       bodyPrefix: 'You received a card in opfab : '
       bodyPostfix: 'This email has been sent by Opfab, there is no need to reply.'
+      publisherEntityPrefix: 'The card has been sent by '
       dailyEmailTitle: 'Cards received during the day'
       hourToSendDailyEmail: 7
       minuteToSendDailyEmail: 30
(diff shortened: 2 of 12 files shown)


node-services/cards-external-diffusion/src/domain/server-side/BusinessConfigOpfabServicesInterface.ts
```diff
@@ -35,13 +37,20 @@ export default class BusinessConfigOpfabServicesInterface extends OpfabServicesI
         if (businessconfigMessage === 'BUSINESS_CONFIG_CHANGE') {
             this.clearBusinessConfigCache();
         }
+        if (businessconfigMessage === 'USER_CONFIG_CHANGE') {
+            this.clearBusinessConfigCache();
```
freddidierRTE: should be this.clearEntitiesCache()

node-services/cards-external-diffusion/src/domain/server-side/BusinessConfigOpfabServicesInterface.ts
```diff
@@ -20,6 +20,8 @@ export default class BusinessConfigOpfabServicesInterface extends OpfabServicesI
     // eslint-disable-next-line @typescript-eslint/ban-types
     private static templateCompilerCache = new Map<string, Map<string, Function> | undefined>();
 
+    private static readonly entitiesCache = new Map<string, any>();
```
freddidierRTE: The name of this class (BusinessConfigOpfabServicesInterface)  is confusing as this class is not only dealing with businessconfig but also with user information with this PR . So i propose to rename it cardsExternalDiffusionOpfabServicesInterface.ts (like it is named for reminder service)

freddidierRTE (changes requested):

node-services/cards-external-diffusion/src/domain/server-side/BusinessConfigOpfabServicesInterface.ts
quinarygio: As there is already a CardsExternalDiffusionOpfabServicesInterface class we could just move the functionalities of BusinessConfigOpfabServicesInterface into it ?

freddidierRTE (approved):

PR #7255, opened 2024-09-23
Refacto : transform selected card service to a store object

freddidierRTE:
This class is to share an internal state: the current selected card. So it is more a store than a service 

Nothing in release note


freddidierRTE:
@coderabbitai review

quinarygio (approved):

PR #7370, opened 2024-10-14
CLI : Add a command line to add or remove a supervised entity (#7331)

### LINKED ISSUE #7331: Opfab cli : Add a command line to add or remove a supervised entity
opfab supervisor add-entity  ... 

opfab supervisor delete-entity ...

ClementBouvierN:
- In release notes :
  -  In chapter : Features
  -  Text : #7331 Opfab Cli : Add a command line to add or remove a supervised entity

### DIFF
--- a/cli/src/supervisorCommands.js
+++ b/cli/src/supervisorCommands.js
@@ -1,0 +1,121 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+const prompts = require('prompts');
+const utils = require('./utils.js');
+
+const supervisorCommands = {
+    async processSupervisorCommand(args) {
+        let command = args[0];
+        if (!command) {
+            command = (
+                await prompts({
+                    type: 'select',
+                    name: 'value',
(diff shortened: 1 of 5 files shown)


cli/src/supervisorCommands.js
```diff
@@ -0,0 +1,119 @@
...
+            default:
+                console.log(`Unknown supervisor command : ${command}
+                `);
+                await this.printHelp();
+                break;
+        }
+    },
+
+    async addSupervisorEntity(supervisedEntity, supervisingEntity) {
+        supervisedEntity = await utils.missingTextPrompt('Supervised entity name', supervisedEntity);
```
quinarygio, line 50: Maybe it is better to check if supervisedEntity was set and if was not set then exit without asking for supervisor.

cli/src/supervisorCommands.js
```diff
@@ -0,0 +1,119 @@
...
+        const entityToSuperviseResponse = await utils.sendRequest(
+            `supervisor/supervisedEntities`,
+            'GET',
+            undefined,
+            '',
+            `Failed to fetch entities to supervise`,
+            ``
+        );
+        if (!entityToSuperviseResponse.ok) {
+            return;
```
quinarygio, line 68: Log an error message if entityToSuperviseResponse not ok

quinarygio (changes requested):

cli/src/supervisorCommands.js
ClementBouvierN: Done

cli/src/supervisorCommands.js
ClementBouvierN: It is already logged by the utils.sendRequest() method calling it. The error message is specified in the call of the method

quinarygio (approved):

PR #7374, opened 2024-10-14
Opfab cli : rename commands for log (#7337)

### LINKED ISSUE #7337: Opfab cli : rename commands for log
Instead of :  

`opfab service get-log-level <serviceName> `

do  

`opfab log get-level  <serviceName>`

Instead of :  

`opfab service set-log-level <serviceName> <level>`

do  

`opfab log set-level  <serviceName> <level>`

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #7337 : Opfab cli : rename commands for log

### DIFF
--- a/cli/src/commands.js
+++ b/cli/src/commands.js
@@ -18,7 +18,7 @@ const perimeters = require('./perimeterCommands.js');
 const processGroups = require('./processGroupCommands.js');
 const realtimescreen = require('./realtimescreenCommands.js');
 const businessData = require('./businessDataCommands.js');
-const service = require('./serviceCommands.js');
+const log = require('./logCommands.js');
 const bundleCommand = require('./bundleCommands.js');
 const monitoringConfig = require('./monitoringConfigCommands.js');
 const processMonitoring = require('./processMonitoringCommands.js');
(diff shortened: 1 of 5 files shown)


cli/src/commands.js
```diff
@@ -18,7 +18,7 @@ const perimeters = require('./perimetersCommands.js');
 const processGroups = require('./processGroupsCommands.js');
 const realtimescreen = require('./realtimescreenCommands.js');
 const businessData = require('./businessDataCommands.js');
-const service = require('./serviceCommands.js');
+const service = require('./logCommands.js');
```
quinarygio: Rename "const service" to "const log"

cli/src/commands.js
quinarygio, line 1: In printHelp function is still referenced the "service" command (row 204 and 246) :

        service                 Get or set log level for services

...

                case 'service':
                    service.printHelp();
                    break;

quinarygio (changes requested):

cli/src/commands.js
vlo-rte: Done.

cli/src/commands.js
vlo-rte: Done.

quinarygio (approved):

PR #7474, opened 2024-11-04
Replace toNotify field with STORE_ONLY_IN_ARCHIVES card Action (#7447)

### LINKED ISSUE #7447: Move toNotify Field to STORE_ONLY_IN_ARCHIVES Action in Card Actions
Today, we have a field toNotify in the card.

The default value is true. When it is set to false, the card is only stored in the archives, so there is no notification on the feed.

The issue is to move this option to an action, i.e., a new possible action in the actions field of the card.

The name of this new action is STORE_ONLY_IN_ARCHIVES.

If the card is an update of an existing card that is in the feed:

- It should first delete the existing card.
- If the action is combined with KEEP_EXISTING_PUBLISH_DATE, it should keep the existing publish date for the new card.
- If the action is combined with PROPAGATE_READ_ACK_TO_PARENT_CARD or KEEP_EXISTING_ACKS_AND_READS, these two actions shall be ignored.

The old field should be kept for backward compatibility but should be stated as deprecated in the documentation, and a log should be put in the cards-publication service with a warning if the field is used.

Add a Karate test similar to [src/test/api/karate/cards/postCardWithToNotify.feature](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html).

Update documentation

Add in migration guide

quinarygio:
Fix #7447 

- In release notes :
  -  In chapter : Features B
  -  Text : #7447 : Replace toNotify field with STORE_ONLY_IN_ARCHIVES card Action


src/test/api/karate/cards/postCardWithStoreOnlyInArchives.feature
```diff
@@ -0,0 +1,126 @@
...
+	"processInstanceId" : "process_postCardWithoutStoreOnlyInArchives",
+	"state": "messageState",
+	"groupRecipients": ["Dispatcher"],
+	"severity" : "ALARM",
+	"startDate" : 1583943924000,
+	"endDate" : 1584943924000,
+	"summary" : {"key" : "defaultProcess.summary"},
+	"title" : {"key" : "defaultProcess.title"},
+	"data" : {"message":"a message without STORE_ONLY_IN_ARCHIVES action"}
+}
```
freddidierRTE: I think you can only keep the test of the card with the action STORE_ONLY_IN_ARCHIVES as the case where the action is not set is already tested in other karate tests

src/docs/asciidoc/resources/migration_guide_to_4.6.adoc
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024 RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// This document is subject to the terms of the Creative Commons Attribution 4.0 International license.
+// If a copy of the license was not distributed with this
+// file, You can obtain one at https://creativecommons.org/licenses/by/4.0/.
+// SPDX-License-Identifier: CC-BY-4.0
+
+= Migration Guide from release 4.5.X to release 4.6.0
+
+== Deprecated `toNotify` field
```
freddidierRTE, line 10: The migration guide has to be referenced in src/docs/asciidoc/resources/index.adoc

freddidierRTE (changes requested):
Action should be add in the front card model , see src/app/business/model/light-card.model.ts

freddidierRTE (approved):

PR #7528, opened 2024-11-13
Solve issue :  Exception in console when editing cards (#7526)

### LINKED ISSUE #7526: Exception in console when editing cards
Sometimes when editing a card we have the following exception : 

image

It does not seems to have any impact as the recipients for information are still present

freddidierRTE:
- In release notes :
  -  In chapter :  Bugs
  -  Text : `#7526 Exception in console when editing cards`

### DIFF
--- a/ui/main/src/app/modules/share/multi-select/multi-select.component.ts
+++ b/ui/main/src/app/modules/share/multi-select/multi-select.component.ts
@@ -31,7 +31,10 @@ export class MultiSelectComponent implements AfterViewInit, OnDestroy, OnChanges
     @Output() selectionChange: EventEmitter<string[]> = new EventEmitter<string[]>();
 
     private oldOptions: Array<MultiSelectOption>;
-    private oldSelectedOptions: any;
+    // It is important to initialize it with an empty array an not undefined to avoid
+    // to trigger the change event when no option is selected on the first load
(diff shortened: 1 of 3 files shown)


ui/main/src/app/modules/share/multi-select/multi-select.component.ts
```diff
@@ -31,7 +31,10 @@ export class MultiSelectComponent implements AfterViewInit, OnDestroy, OnChanges
     @Output() selectionChange: EventEmitter<string[]> = new EventEmitter<string[]>();
 
     private oldOptions: Array<MultiSelectOption>;
-    private oldSelectedOptions: any;
+    // It is important to initialize it with an empty array an not undefined to avoid
+    // to trigger the change event when no option is selected on the first load
+    // See Issue #7526
+    private oldSelectedOptions: any = new Array<string>();
```
ClementBouvierN: There is a little typo in the comment, I suggest:
```
    // It is important to initialize it with an empty array and not undefined to avoid
    // triggering the change event when no option is selected on the first load
    // See Issue #7526
```
And I don't think it's necessary to keep the ": any" on the declaration of oldSelectedOptions

ClementBouvierN (approved):

PR #7640, opened 2024-12-05
Add the list of child entities in recipient and recipient for information multiselect  (#7588)

### LINKED ISSUE #7588: Add the list of child entities in recipient and recipient for information multiselect
image

The list of child entities shall be in grey and italic or 
see if it is possible to show it as a tooltip instead   (https://opfab.github.io/documentation/current/reference_doc/#_tooltip)

To do that , you can use a specific render fonction for multiselect , see : https://sa-si-dev.github.io/virtual-select/#/examples?id=add-imageicon

This render fonction shall be pass a input in multi-select.component.ts 

The result of processing of building the child entities per entities shall be somehow cached/memorized to avoid unecessary processing . It is to be done in entities.service and unit test

quinarygio:
Fix #7588

- In release notes :
  -  In chapter : Features 
  -  Text : #7588 : Add the list of child entities in recipient and recipient for information multiselect

### DIFF
--- a/ui/main/src/app/business/services/users/entities.service.ts
+++ b/ui/main/src/app/business/services/users/entities.service.ts
@@ -18,7 +18,8 @@ import {ErrorService} from '../error-service';
 import {RolesEnum} from '@ofModel/roles.model';
 
 export class EntitiesService {
-    protected static _entities: Entity[];
+    private static _entities: Entity[];
+    private static readonly _childEntities: Map<string, Entity[]> = new Map<string, Entity[]>();
     private static readonly ngUnsubscribe$ = new Subject<void>();
     private static entitiesServer: EntitiesServer;
 
@@ -40,6 +41,16 @@ export class EntitiesService {
 
     private static deleteFromCachedEntities(id: string): void {
         EntitiesService._entities = EntitiesService._entities.filter((entity) => entity.id !== id);
+        EntitiesService._childEntities.delete(id);
+        EntitiesService.deleteFromCachedChildEntities(id);
+    }
+
+    private static deleteFromCachedChildEntities(id: string) {
+        const parentEntitiesIds = [...EntitiesService._childEntities.entries()]
+            .filter(([, childs]) => childs.findIndex((child) => child.id === id) >= 0)
+            .map((entry) => entry[0]);
+
+        parentEntitiesIds.forEach((parentId) => EntitiesService._childEntities.delete(parentId));
     }
 
     public static queryAllEntities(): Observable<Entity[]> {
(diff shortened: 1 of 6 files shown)


ui/main/src/app/business/services/users/entities.service.ts
```diff
@@ -19,6 +19,7 @@ import {RolesEnum} from '@ofModel/roles.model';
 
 export class EntitiesService {
     protected static _entities: Entity[];
+    protected static _childEntities: Map<String, Entity[]> = new Map<String, Entity[]>();
```
freddidierRTE: When calling loadAllEntitiesData() , you need to clear the child entities cache (loadAllEntitiesData is called when entities changes on the server side)

ui/main/src/app/business/services/users/entities.service.ts
```diff
@@ -197,12 +218,18 @@ export class EntitiesService {
...
+        let childEntities = EntitiesService._childEntities.get(parentId);
+        if (!childEntities) {
+            const resolved = new Set<Entity>();
+            const parent = EntitiesService._entities.find((e) => e.id === parentId);
+            if (parent) {
+                EntitiesService.findChildEntities(parent).forEach((cc) => resolved.add(cc));
+            }
+            childEntities = Array.from(resolved);
+            childEntities.sort((a, b) => a.description.localeCompare(b.description));
+            if (childEntities.length > 0) EntitiesService._childEntities.set(parentId, childEntities);
```
freddidierRTE: i would store the array in the cache even if empty as it would avoid to check every time every entities with no child

ui/main/src/app/modules/usercard/usercard.component.ts
```diff
@@ -176,6 +178,24 @@ export class UserCardComponent implements OnDestroy, UserCardUIControl, AfterVie
...
         );
     }
+
+    public entityRecipientLabelRenderer(data) {
+        const childEntities = EntitiesService.resolveChildEntities(data.value);
+        let label = data.label;
+        if (childEntities.length) {
+            label =
+                label +
+                ' <i style="color: var(--opfab-lightcard-detail-textcolor);text-overflow: ellipsis">(' +
```
freddidierRTE: It should be grey for day mode as well

image

ui/main/src/app/business/services/users/entities.service.ts
```diff
@@ -19,6 +19,7 @@ import {RolesEnum} from '@ofModel/roles.model';
 
 export class EntitiesService {
     protected static _entities: Entity[];
+    protected static _childEntities: Map<string, Entity[]> = new Map<string, Entity[]>();
```
freddidierRTE: The cache shall be clean when calling loadAllEntitiesData (this method is called by application-update-service to relaod entities when it changes on the server side

freddidierRTE: _entities and _childEntities shall be private variables (i guess it was necessary protected before for admin screens features, but some refactoring has been done and it is not necessary anymore)

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #7666, opened 2024-12-11
Usercard preview : show child entities as tooltip (#7651)

### LINKED ISSUE #7651: Usercard preview : show child entities as tool tip
On the card preview, when the user hovers over the recipient or recipient for information labels ,if  there is child entities associated with the entity ,the child entities shall be displayed in a tooltip

Use https://opfab.github.io/documentation/current/reference_doc/#_tooltip to do the tooltip 

Modify an existing cypress test to check the behavior

quinarygio:
Fix #7651

- In release notes :
  -  In chapter : Features
  -  Text : #7651 : Usercard preview : show child entities as tooltip


freddidierRTE (changes requested):
Should be done for recipients for information as well 

Set a max size for tooltip with a new line if too long 

image

freddidierRTE (approved):

PR #7690, opened 2024-12-16
Add the possibility to have a button to edit card (#7525)

### LINKED ISSUE #7525: Edit card link : add it as a button
In a card created from a user card, we need to add a link to edit the card. Currently, the user can edit the card by clicking on the pencil (EDIT) icon. Here, the recipient who needs to edit the card is not the publisher. We think it would be more user-friendly to have the possiblity to add a button within the card to edit the card.

image

Thanks.

quinarygio:
- In release notes :
  -  In chapter : Features
  -  Text : #7525 : Add the possibility to have a button to edit card

### DIFF
--- a/ui/main/src/app/modules/card/components/card-body/card-body.component.ts
+++ b/ui/main/src/app/modules/card/components/card-body/card-body.component.ts
@@ -51,6 +51,7 @@ import {OpfabTitleCasePipe} from '../../../share/pipes/opfab-title-case.pipe';
 import {CardBodyView} from 'app/business/view/card/card-body.view';
 import {ConfigService} from 'app/services/config/ConfigService';
 import {CardTemplateGateway} from 'app/business/templateGateway/cardTemplateGateway';
+import {PermissionEnum} from '@ofModel/permission.model';
 
 @Component({
     selector: 'of-card-body',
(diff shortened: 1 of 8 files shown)


ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -233,6 +236,20 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
         }
     }
 
+    private computeUserAllowedToEdit() {
+        this.isUserAllowedToEdit =
+            !UserService.hasCurrentUserAnyPermission([PermissionEnum.READONLY]) &&
```
freddidierRTE: you do not need to check the readonly anymore as it is checked in doesTheUserHavePermissionToEditCard

The editCardEnabledOnUserInterface shall not be checked anymore as even if the edit link is not visible on screen it should be possible to edit the card via js code from the template

freddidierRTE (changes requested):

ui/main/src/app/modules/card/components/card-body/card-body.component.ts
```diff
@@ -233,6 +236,18 @@ export class CardBodyComponent implements OnChanges, OnInit, OnDestroy {
         }
     }
 
+    private computeUserAllowedToEdit() {
+        this.isUserAllowedToEdit =
+            this.cardState.editCardEnabledOnUserInterface && this.doesTheUserHavePermissionToEditCard();
```
freddidierRTE: The editCardEnabledOnUserInterface shall not be checked anymore as even if the edit link is not visible on screen it should be possible to edit the card via js code from the template

freddidierRTE (approved):

PR #7700, opened 2024-12-17
Add entity ID in entity selection for admin screens (#7662)

### LINKED ISSUE #7662: Add entity ID in entity selection for admin screens
For certain use cases , it is possible to have two entities with same name 

In this case when we need to choose an entity in admin screen we need to provide more information than the name , the idea is to provide the name + the ID inside the enitites mutli-selects

This concerns : 
- the screen to create and edit users  (User entities)
- the screen to create and edit entities  (Parent entities)

Use the same feature as for usercard recipients to show the ID when selecting inside the mutliselect but do not show the ID in the labels

quinarygio:
- In release notes :
  -  In chapter : Features
  -  Text : #7662 : Add entity ID in entity selection for admin screens

### DIFF
--- a/ui/main/src/app/business/services/admin/crudUtils.spec.ts
+++ b/ui/main/src/app/business/services/admin/crudUtils.spec.ts
@@ -1,0 +1,56 @@
+/* Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {EntitiesService} from 'app/business/services/users/entities.service';
+import {EntitiesServerMock} from '@tests/mocks/entitiesServer.mock';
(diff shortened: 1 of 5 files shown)


ui/main/src/app/business/services/admin/crudUtils.ts
```diff
@@ -28,4 +29,24 @@ export class CrudUtilities {
...
+        return data.label;
+    }
+
+    private static isDuplicatedEntityName(label: string, id: string): boolean {
+        return (
+            label &&
+            EntitiesService.getEntities().filter((entity) => entity.name === label.trim() && entity.id !== id).length >
+                0
+        );
+    }
```
freddidierRTE, line 51: Add unit tests

ui/main/src/app/business/services/admin/crudUtils.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import {EntitiesService} from 'app/business/services/users/entities.service';
+import {EntitiesServerMock} from '@tests/mocks/entitiesServer.mock';
+import {Entity} from '@ofModel/entity.model';
+import {firstValueFrom} from 'rxjs';
+import {CrudUtilities} from './crudUtils';
+import {MultiSelectOption} from '@ofModel/multiselect.model';
+
+describe('User action logs view ', () => {
```
freddidierRTE: Test name to change

freddidierRTE (approved):

PR #7839, opened 2025-01-14
Move events services in new ui services directory

freddidierRTE:
Nothing in release note


vlo-rte (changes requested):
One issue reported by sonar. Please check.
And two security hotspots.

vlo-rte (approved):

PR #7853, opened 2025-01-16
Improve admin supervised entity service error management

freddidierRTE:
Nothing in release note

### DIFF
--- a/ui/main/src/app/services/admin/SupervisedEntitiesService.ts
+++ b/ui/main/src/app/services/admin/SupervisedEntitiesService.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2023-2024, RTE (http://www.rte-france.com)
+/* Copyright (c) 2023-2025, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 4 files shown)


ui/main/src/app/services/admin/SupervisedEntitiesService.ts
```diff
@@ -30,7 +31,12 @@ export class SupervisedEntitiesService {
                 if (entitiesResponse.status === ServerResponseStatus.OK) {
                     SupervisedEntitiesService.deleteFromCachedEntities(id);
                 } else {
-                    ErrorService.handleServerResponseError(entitiesResponse);
+                    logger.error(`Error while deleting entity :  ${entitiesResponse.statusMessage}`);
```
vlo-rte: Maybe we should add in the log the id of the entity we're trying to delete ?

ui/main/src/app/services/admin/SupervisedEntitiesService.ts
```diff
@@ -63,7 +74,12 @@ export class SupervisedEntitiesService {
 
                     return responseEntities.data;
                 } else {
-                    ErrorService.handleServerResponseError(responseEntities);
+                    logger.error(`Error while updating entity :  ${responseEntities.statusMessage}`);
```
vlo-rte: Maybe we should add in the log the id of the entity we're trying to update ?

vlo-rte (changes requested):

ui/main/src/app/services/admin/SupervisedEntitiesService.ts
freddidierRTE: ok

ui/main/src/app/services/admin/SupervisedEntitiesService.ts
freddidierRTE: ok

vlo-rte (approved):

PR #7865, opened 2025-01-20
Add possibility to hide publisher multiselect in usercard (#7850)

### LINKED ISSUE #7850: Add possibility to hide publisher mutliselect in usercard
Add a new param "publisherVisible" 

image

Use it in "Message from user" use case

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #7850 : Add possibility to hide publisher multiselect in usercard


freddidierRTE (changes requested):
In https://opfab.github.io/documentation/archives/SNAPSHOT/docs/single_page_doc.html#_send_a_card_on_behalf_of_the_user

add  the advice to set the field publisherVisible to false when sending a card on behaf of user to avoid the user seeing the publisher entity mutliselect

freddidierRTE (approved):

PR #7899, opened 2025-01-28
Add an option to notify the user only when the card is reminded (#7883)

### LINKED ISSUE #7883: Add an option to notify the user only when the card is reminded
Add a new  action NOT_NOTIFIED : 

- [x] when this action si set, the card is not visible in the card feed except for the publisher but it is visible in all the other screen : monitoring screen / agenda  / timeline 

- [x] It is visible in timeline only if "Do not apply filter to timeline" is set 

- [x] But when a card it remind the option not_notified shall be removed by the reminder when launching the remind card : so this should be implemented in the deleteAckAndRead endpoint

- [x] This action shall be used in the task usercard

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #7883 : Add an option to notify the user only when the cars is reminded


freddidierRTE (changes requested):
- You need to stop sound as well , this is to be done in NotificationDecision.ts 

Add a new test in NotificationDecision.spec.ts : no sound if action is NOT_NOTIFIED and user entities does not contains card publisher entity 

So the only case where it sounds is when the user is member of the publihser entity and is not the sender of the card

- Add a test in cardReminder.feature : send a card with action NOT_NODIFIED and call resetReadAndAcks endpoint 

- Add in documentation : https://opfab.github.io/documentation/current/reference_doc/#_actions_actions

vlo-rte:
> * You need to stop sound as well , this is to be done in NotificationDecision.ts
> 
> 
> Add a new test in NotificationDecision.spec.ts : no sound if action is NOT_NOTIFIED and user entities does not contains card publisher entity
> 
> So the only case where it sounds is when the user is member of the publihser entity and is not the sender of the card
> 
>     * Add a test in cardReminder.feature : send a card with action NOT_NODIFIED and call resetReadAndAcks endpoint
> 
>     * Add in documentation : https://opfab.github.io/documentation/current/reference_doc/#_actions_actions

Done.

src/test/api/karate/cards/cardsReminder.feature
```diff
@@ -178,6 +196,28 @@ Scenario: ResetCardsReadsAndAcks
...
+    Then status 200
+    And match response.card.hasBeenRead == false
+    And match response.card.hasBeenAcknowledged == false
+    And def uid = response.card.uid
+
+    Given url opfabUrl + 'cards-publication/cards/resetReadAndAcks/' + uid
+    And header Authorization = 'Bearer ' + authTokenInternal
+    And request ''
+    When method post
+    Then status 200
```
freddidierRTE, line 220: Get card and check response.card.actions does not contains  'NOT_NOTIFIED'

src/test/api/karate/cards/cardsReminder.feature
```diff
@@ -178,6 +196,28 @@ Scenario: ResetCardsReadsAndAcks
...
+    And request cardWithNotNotifiedAction
+    When method post
+    Then status 201
+
+    Given url opfabUrl + 'cards-consultation/cards/api_test.cardWithNotNotifiedAction'
+    And header Authorization = 'Bearer ' + authToken
+    When method get
+    Then status 200
+    And match response.card.hasBeenRead == false
+    And match response.card.hasBeenAcknowledged == false
```
freddidierRTE, line 212: And match response.card.actions == ['NOT_NOTIFIED']

ui/main/src/app/services/notifications/NotificationDecision.ts
```diff
@@ -150,4 +152,11 @@ export class NotificationDecision {
             this.cleanSentCards();
         }, this.CLEAN_CARDS_PERIOD);
     }
+
+    private static shouldCardBeNotified(card: Card): boolean {
+        return (
+            !card.actions?.includes(CardAction.NOT_NOTIFIED) ||
+            UsersService.getCurrentUserWithPerimeters().userData.entities.includes(card.publisher)
```
freddidierRTE: i change my mind : just test if NOT NOTIFIED

freddidierRTE (changes requested):

ui/main/src/app/services/notifications/NotificationDecision.ts
vlo-rte: Done.

src/test/api/karate/cards/cardsReminder.feature
vlo-rte: Done.

src/test/api/karate/cards/cardsReminder.feature
vlo-rte: Done.

freddidierRTE (approved):

PR #8042, opened 2025-02-26
Add loading spinner in custom card list screen (#7904)

### LINKED ISSUE #7904: Add loading spinner in custom card list screen
Do like in monitoring screen 

See monitoring.component.html

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #7904 : Add loading spinner in custom card list screen

### DIFF
--- a/ui/main/src/app/components/customCardList/CustomCardListComponent.html
+++ b/ui/main/src/app/components/customCardList/CustomCardListComponent.html
@@ -75,7 +75,7 @@
         <div *ngIf="responseFromMyEntitiesFilterVisible">
             <label
                 id="opfab-show-cards-with-response"
-                style="font-weight: bold; margin-left: 5px; margin-right: 30px; margin-bottom: 0px"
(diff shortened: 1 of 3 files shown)


ui/main/src/scss/styles.scss
```diff
@@ -720,3 +720,13 @@ tag-input {
...
+
+.opfab-loading-spinner {
+    float: right;
+    margin-right: 5%;
+    font-size: 30px;
+    border: solid 1px var(--opfab-bgcolor);
+    text-align: center;
+    background-color: var(--opfab-bgcolor);
+    color: #aaa;
+}
```
freddidierRTE: I would prefer keeping loading spinner in css component files (event if duplication with monitoring) as there is other opfab-loading-spinner class in other components (filters, spinner and archives components) with differents settings.

And i think it is conflicting with loading spinner that appear on application loading : 

image

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #8104, opened 2025-03-14
Custom card list screen : stop updating screen when user is editing (#7999)

### LINKED ISSUE #7999: Custom card list screen : stop updating screen when user is editing
When user select a card for reponse stop updating the table of cards (this is to avoid overriding user input and selection if for example a new card arise) 

Update again if all checkbox are unset (i.e no card is selected for response)

ClementBouvierN:
Nothing in release notes

### DIFF
--- a/ui/main/src/app/components/customCardList/CustomCardListComponent.ts
+++ b/ui/main/src/app/components/customCardList/CustomCardListComponent.ts
@@ -16,7 +16,13 @@ import {SelectedCardService} from '@ofServices/selectedCard/SelectedCardService'
 import {TranslationService} from '@ofServices/translation/TranslationService';
 import {CardComponent} from 'app/components/card/card.component';
 import {AgGridAngular} from 'ag-grid-angular';
-import {AllCommunityModule, ModuleRegistry, provideGlobalGridOptions, RowSelectionOptions} from 'ag-grid-community';
+import {
+    AllCommunityModule,
+    ModuleRegistry,
+    provideGlobalGridOptions,
+    RowSelectionOptions,
+    SelectionChangedEvent
+} from 'ag-grid-community';
 import {DateRangePickerConfig} from 'app/utils/DateRangePickerConfig';
 import {ExcelExport} from 'app/utils/ExcelExport';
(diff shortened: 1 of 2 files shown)


ui/main/src/app/components/customCardList/CustomCardListComponent.ts
```diff
@@ -331,6 +345,27 @@ export class CustomCardListComponent implements OnInit, OnDestroy {
...
+        this.customCardListView
+            .getResults()
+            .pipe(takeUntil(this.inputMode$))
+            .subscribe((results) => {
+                this.rowData = results;
+                this.rowDataSubject.next(this.rowData);
+            });
+    }
+
+    onInputCellsAvailable() {
```
freddidierRTE: i would name it stopListeningToResults

ui/main/src/app/components/customCardList/CustomCardListComponent.ts
```diff
@@ -323,6 +327,16 @@ export class CustomCardListComponent implements OnInit, OnDestroy {
...
     onGridReady(params: any) {
         this.gridApi = params.api;
+
+        this.gridApi.addEventListener('selectionChanged', (event: SelectionChangedEvent) => {
+            const userHasSelectedRows = this.gridApi.getSelectedRows().length > 0;
+
+            if (userHasSelectedRows) {
+                this.onInputCellsAvailable();
+            } else {
+                this.onInputCellsDone();
```
freddidierRTE: Directly call method startListeningToResults

freddidierRTE: Add a comment to explain we stop updating the screen as soon as user has something selected to avoid moving line when user si filling input fields

ui/main/src/app/components/customCardList/CustomCardListComponent.ts
freddidierRTE: And you should call this method on component destroy

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #8108, opened 2025-03-17
Refacto ack in the front

freddidierRTE:
Transfert the logic from the components to the service

Nothing in release note


ui/main/src/app/services/acknowlegment/AcknowledgeService.spec.ts
```diff
@@ -1,4 +1,5 @@
-/* Copyright (c) 2025, RTE (http://www.rte-france.com)
+/* Copyright (c) 2025, RTE (http://www.rte-france.com):console.warn();
```
ClementBouvierN: Typo in copyright

ClementBouvierN (approved):

PR #8126, opened 2025-03-21
Custom Card List refactoring

freddidierRTE:
Modularize the code by splitting classes:

- Extract logic from ResultTable into two new classes:
       - CardFilter: filters cards based on user filtering values.
       - TableRowBuilder: creates the row base and card content and column definition.
       
- Add a ButtonAction class to deal with user actions (respond or acknowledge).

- Remove the classes Acknowledgments and Responses as there are no longer needed; the code has been distributed into the classes defined above.

- Split tests into more files to avoid overly long files.

- Add a FilterValues object to transmit filter values and avoid calling multiple methods.

- Simplify the code regarding acknowledge and read filters.

- Add a helper function to mock translations in tests.

- Make some general cleanup in code and test code.

**Nothing in release note**


ui/main/src/app/components/customCardList/CustomCardListComponent.ts
```diff
@@ -316,7 +316,10 @@ export class CustomCardListComponent implements OnInit, OnDestroy {
                 isRowSelectable: (node) => {
                     return (
                         (this.responseButtons.length > 0 && node.data.isResponsePossible) ||
-                        (this.customCardListView.isAcknowledgmentButtonVisible() && node.data.isAcknowledgmentPossible)
+                        false ||
+                        (this.customCardListView.isAcknowledgmentButtonVisible() &&
+                            node.data.isAcknowledgmentPossible) ||
+                        false
```
ClementBouvierN: Adding "|| false" to prevent undefined is not the most readable. There should be a better way to do this check

ui/main/src/app/views/customCardList/FilterValues.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+export class FilterValues {
+    public startDate: number;
+    public endDate: number;
+    public processes: string[];
+    public typesOfStateFilter: string[];
+    public readAndAckFilter: string[];
+    public responseFromMyEntitiesChoice: boolean;
+    public responseFromAllEntitiesChoice: boolean;
```
ClementBouvierN: Potentially rename those variables for clarity

ClementBouvierN (dismissed):

ClementBouvierN (approved):

PR #8183, opened 2025-03-31
Adjust task advanced cypress test to avoid flaky tests

freddidierRTE:
Nothing in release note

### DIFF
--- a/src/test/cypress/cypress/integration/TaskAdvanced.spec.js
+++ b/src/test/cypress/cypress/integration/TaskAdvanced.spec.js
@@ -1,4 +1,4 @@
-/* Copyright (c) 2023-2024, RTE (http://www.rte-france.com)
+/* Copyright (c) 2023-2025, RTE (http://www.rte-france.com)
  * See AUTHORS.txt
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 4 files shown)


src/test/cypress/cypress/integration/TaskAdvanced.spec.js
```diff
@@ -7,286 +7,306 @@
...
+            opfab.navigateToUserCard();
+            usercard.checkEmitterSelectDoesNotExist();
+            usercard.selectService('User card examples');
+            usercard.selectProcess('Task Advanced');
+            cy.waitDefaultTime();
+            cy.get('#time').type('00:00');
+            usercard.previewThenSendCard();
+
+            // It should appears 7 circles on the timeline but when current day is near the day-savings time change,
+            // it can be 8 circles as the timeline in Week mode is shift from one hour hour.
```
ClementBouvierN: The word hour is duplicated here

src/test/cypress/cypress/integration/TaskAdvanced.spec.js
```diff
@@ -7,286 +7,306 @@
...
+            checkHaveAtLeastCircle(7);
+
+            opfab.navigateToCalendar();
+
+            // The number of cards is not always the same especially when the test is run on the end of the month
```
ClementBouvierN: The correct sentence here would be 
The number of cards is not always the same especially when the test is ran at the end of the month

ClementBouvierN (approved):

PR #8303, opened 2025-04-24
Usercard Payload Size Limitation (#8247)

### LINKED ISSUE #8247: Usercard Payload Size Limitation - Error Handling and Configuration
**Problem:** 
Users encounter errors when attempting to send usercards containing large images or text payloads. This is caused by a size limitation on request payloads, currently configured within the Nginx server.

**Root Cause**
- The default client_max_body_size in the Nginx configuration is set to 1.5MB.
- While this value can be increased, there's still a risk of exceeding the limit with excessively large usercard content.
- Storing very large usercard data in the database can negatively impact performance.

**Proposed Solution**

- Add a user-friendly error message when the usercard payload exceeds the configured limit : This message should clearly indicate the size restriction and advise users to reduce the content size.

- Maintain Default Limit: Retain the client_max_body_size at 1.5MB to prevent excessive data storage and potential performance issues.

- Document the 1.5MB limit, and the reason for that limit.

ClementBouvierN:
- In release notes :
  -  In chapter : Features
  -  Text : #8247 Usercard Payload Size Limitation

### DIFF
--- a/ui/main/src/assets/i18n/en.json
+++ b/ui/main/src/assets/i18n/en.json
@@ -446,6 +446,7 @@
             "notAvailable": "You are not allowed to send cards.",
             "templateError": "Technical error : invalid template",
             "impossibleToSendCard": "Impossible to send card for technical reasons",
+            "cardContentTooLarge": "Impossible to send card, the content of the card is too large (max 1.5 Mb)",
             "lttdBeforeStartDate": "Limit date for response must be after start date and after the current date and time.",
             "lttdAfterEndDate": "Limit date for response must be before end date.",
             "expirationDateBeforeStartDate": "Expiration date must be after start date."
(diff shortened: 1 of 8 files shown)


ui/main/src/assets/i18n/en.json
```diff
@@ -446,6 +446,7 @@
             "notAvailable": "You are not allowed to send cards.",
             "templateError": "Technical error : invalid template",
             "impossibleToSendCard": "Impossible to send card for technical reasons",
+            "cardContentTooLarge": "The content of the card is too large (max 1.5 Mb)",
```
freddidierRTE: Impossible to send card , the content of the card ...

freddidierRTE (changes requested):
Add a unit test for this new error

freddidierRTE (approved):

PR #8410, opened 2025-05-20
Change the logic of activation of mail details (#8367)

### LINKED ISSUE #8367: Change the logic of activation of mail details
Hi OpFab colleagues,

Would it be possible to change the logic of this setting:

Image

The idea would be to have "Enable card content in emails" and have it disabled by default.

We remain available for further discussion.

Kind regards,

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #8367 : Change the logic of activation of mail details


freddidierRTE:
not needed anymore

PR #8414, opened 2025-05-21
Use ModalService for ActivityAreaComponent (#8385)

### LINKED ISSUE #8385: Use ModalService for ActivityAreaComponent
Replace confirmationPopup modal with a call to ModalService

Example : 

```
    public openDeleteConfirmationModal() {
        ModalService.openConfirmationModal(
            new I18n('userCard.deleteCard.title'),
            new I18n('userCard.deleteCard.doYouReallyWant')
        ).then((confirm) => {
            if (confirm) {
                this.deleteCard();
            }
        });
    }
```

vlo-rte:
Nothing in release notes.

### DIFF
--- a/ui/main/src/app/components/activityarea/ActivityAreaComponent.ts
+++ b/ui/main/src/app/components/activityarea/ActivityAreaComponent.ts
@@ -8,8 +8,6 @@
  */
 
 import {Component, ElementRef, EventEmitter, Input, OnDestroy, OnInit, Output, ViewChild} from '@angular/core';
-import {NgbModal} from '@ng-bootstrap/ng-bootstrap';
-import {NgbModalRef} from '@ng-bootstrap/ng-bootstrap/modal/modal-ref';
 import {FormControl, FormGroup, FormsModule, ReactiveFormsModule} from '@angular/forms';
 import {ActivityAreaView} from 'app/views/activityarea/ActivityAreaView';
 import {ActivityAreaPage} from 'app/views/activityarea/ActivityAreaPage';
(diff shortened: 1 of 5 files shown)


ui/main/src/app/components/activityarea/ActivityAreaComponent.ts
```diff
@@ -119,20 +120,16 @@ export class ActivityAreaComponent implements OnInit, OnDestroy {
...
+                new I18n('shared.popup.title'),
+                new I18n('shared.popup.areYouSure')
+            ).then((confirm) => {
+                if (confirm) {
+                    this.confirmSaveSettings();
+                }
+            });
         } else {
             this.confirmSaveSettings();
         }
```
freddidierRTE, line 122: - the variable confirmationPopup should be removed as it is not used anymore
- the inject NgbModal in the constructor is not needed as well

freddidierRTE (changes requested):
I think the following css from ActivityAreaComponent.scss is not needed anymore, to be tested 

```
.modal-body {
    background-color: var(--opfab-bgcolor);
    color: var(--opfab-textcolor);
}

.modal-header {
    background-color: var(--opfab-bgcolor);
    color: var(--opfab-textcolor);
}
```

vlo-rte:
> ActivityAreaComponent

Done.

ui/main/src/app/components/activityarea/ActivityAreaComponent.ts
vlo-rte: Done.

freddidierRTE (approved):

PR #8467, opened 2025-06-04
Custom screens : Sorting takes into account emoticones whereas it should not (#8239)

### LINKED ISSUE #8239: Custom screens : Sorting takes into account emoticones whereas it should not
Image

vlo-rte:
- In release notes :
  -  In chapter : Bugs
  -  Text : #8239 : Custom screens : Sorting takes into account emoticones whereas it should not


freddidierRTE (changes requested):
Add sorting without emoticones for html column type as well

vlo-rte:
> Add sorting without emoticones for html column type as well

Done.

freddidierRTE (approved):

PR #8488, opened 2025-06-06
Export of "keywords field" in custom screen (#8476)

### LINKED ISSUE #8476: Keyword export in custom screen
When exporting data in custom screen 1, the keyword field isn't being exported. There is a discrepancy between the fields that the export function is looking for and the fields defined in the customscreenexample file for this column

vlo-rte:
- In release notes :
  -  In chapter : Bugs
  -  Text : #8476 : Export of "keywords field" in custom screen


freddidierRTE:
Add unit test  in `ResultTable-Export.specs.ts`  similar to     `'Should contain numerical value for colored circle'
`

vlo-rte:
> Add unit test in `ResultTable-Export.specs.ts` similar to `'Should contain numerical value for colored circle' `

Done.

freddidierRTE (approved):

PR #8585, opened 2025-07-03
Store last user connection in database (#8331)

### LINKED ISSUE #8331: Store last user connection in database
Today we store user action so we know when user has last connected or has made certain actions 

The problem is that this collection for confidential reasons is time limited . We want to see if an account is still in use , for example if it has been used in the current year but the action log store only three months data

The idea is to store last user action in new collection : lastUserAction that should contains two fields 

userLogin : string 
lastAction : epoch date 

The collection shall be updated each time a user action is traced , should be added in the current code dealing with user action 

Add access to the information via cli  : opfab user last-action 

Add karate test 

Add in documentation

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #8331 : Store last user connection in database

### DIFF
--- a/services/users/src/main/modeling/swagger.yaml
+++ b/services/users/src/main/modeling/swagger.yaml
@@ -442,6 +442,20 @@ definitions:
         type: number
         description: >-
           Page number
+  LastUserAction:
+    type: object
+    properties:
+      login:
+        type: string
+      lastActionDate:
+        type: number
+        description: User last action as epoch date
+    required:
+      - login
+      - lastActionDate
+    example:
+      login: jcleese
+      lastActionDate: 1719911567000
 paths:
   /users/users:
     get:
@@ -635,7 +649,7 @@ paths:
       tags:
         - users
       summary: Update existing user settings
-      description: Update existing user settiogs
+      description: Update existing user settings
       operationId: updateUserSettings
       consumes:
         - application/json
@@ -1866,6 +1880,47 @@ paths:
           schema:
             type: object
             $ref: '#/definitions/UserActionLogPage'
+  /users/lastUserAction:
+    get:
+      tags:
+        - last user action
+      summary: Get last user action
+      description: 'Get the list the user actions.'
+      operationId: getLastUserAction
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/LastUserAction'
+  /users/lastUserAction/{login}:
+    get:
+      tags:
+        - last user action
+      summary: Fetch a last user action, for a specific user
+      description: Fetch a last user action, for a specific user
+      operationId: fetchLastUserAction
+      produces:
+        - application/json
+      parameters:
+        - in: path
+          name: login
+          description: user login
+          type: string
+          required: true
+      responses:
+        '200':
+          description: OK
+          schema:
+            $ref: '#/definitions/LastUserAction'
+        '401':
+          description: Authentication required
+        '403':
+          description: >-
+            Authenticated users who are not admins can only access the data
   /users/notificationconfiguration/processstatenotified/{process}/{state}:
       post:
         tags:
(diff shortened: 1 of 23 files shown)


services/users/src/main/modeling/swagger.yaml
```diff
@@ -442,6 +442,19 @@ definitions:
         type: number
         description: >-
           Page number
+  LastUserAction:
+    type: object
+    properties:
+      login:
+        type: string
+      lastActionDate:
+        type: number
```
freddidierRTE, line 451: Add the description yml field to explain that the date is epoch date

```
properties : 
    type: number
    description : User last action as epoch date 
```

src/test/api/karate/users/getLastUserAction.feature
```diff
@@ -0,0 +1,47 @@
...
+    * def signIn = callonce read('../common/getToken.feature') { username: 'admin'}
+    * def authToken = signIn.authToken
+    * def signInAsTSO = callonce read('../common/getToken.feature') { username: 'operator1_fr'}
+    * def authTokenAsTSO = signInAsTSO.authToken
+    * def signInAsSupervisor = callonce read('../common/getToken.feature') { username: 'itsupervisor1'}
+    * def authTokenAsSupervisor = signInAsSupervisor.authToken
+
+
+  Scenario: Get last user action as admin
+    # get all 'last user action' as admin
```
freddidierRTE: Instead of adding comment , just modifty the scenario name :  "Get all user last action as admin"

src/test/api/karate/users/getLastUserAction.feature
```diff
@@ -0,0 +1,47 @@
...
+    Given url opfabUrl + 'users/lastUserAction'
+    And header Authorization = 'Bearer ' + authToken
+    When method get
+    Then status 200
+
+  Scenario: Get last user action as admin
+    # get all 'last user action' as admin
```
freddidierRTE: Instead of adding comment , just modifty the scenario name "Get operator1_fr last action"

src/test/api/karate/users/getLastUserAction.feature
```diff
@@ -0,0 +1,47 @@
...
+    Given url opfabUrl + 'users/lastUserAction/operator1_fr'
+    And header Authorization = 'Bearer ' + authToken
+    When method get
+    Then status 200
```
freddidierRTE, line 29: i would test that we have in the response the good json : a field name login with the good login and a field name lastActionDate 
To be sure we have an action , it is better to call first  for example /cards-consultation/cardSubscription with operator1_fr (in case we launch this test manually just after restart with a fresh install)

tools/user-action-tracing/src/main/java/org/opfab/useractiontracing/repositories/LastUserActionRepository.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public LastUserAction save(LastUserAction userActionLog);
+
+    public Optional<LastUserAction> findById(String id);
+
+    public void deleteExpiredLastUserAction(Instant expirationDate);
+
+    public void delete(LastUserAction entity);
+
+    public void deleteAll();
```
freddidierRTE: The delete methods are never used , to be removed

freddidierRTE (changes requested):

services/users/src/main/modeling/swagger.yaml
vlo-rte: Done.

src/test/api/karate/users/getLastUserAction.feature
vlo-rte: Done.

src/test/api/karate/users/getLastUserAction.feature
vlo-rte: Done.

src/test/api/karate/users/getLastUserAction.feature
vlo-rte: Done.

tools/user-action-tracing/src/main/java/org/opfab/useractiontracing/repositories/LastUserActionRepository.java
vlo-rte: Done.

freddidierRTE (approved):

PR #8612, opened 2025-07-09
Change the default period of the calendar date picker for custom screens (#8607)

### LINKED ISSUE #8607: Change the default period of the calendar date picker for custom screens
Hi OperatorFabric colleagues,

Would it be possible to change the default period of the new calendar date picker ?

Ideally, we would like that by default the selected period is from today to the end of the current year.

Have a nice rest of the day.

### LINKED ISSUE #8655: Custom screen : add option to set initial business period
Add a custom screen parameter  "initialBusinessPeriod" , this parameter shall set the business period when the custom screen is displayed to the user 

Possible values : 

-  FEED : period based on feed period
-  FROM_TODAY_TO_YEAR_END : from now to the end of the year 
 
 
 Default value is FEED
 

 Add in documentation

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #8607 : Change the default period of the calendar date picker for custom screens

### DIFF
--- a/ui/main/src/app/views/customCardList/CustomCardListView.ts
+++ b/ui/main/src/app/views/customCardList/CustomCardListView.ts
@@ -45,8 +45,17 @@ export class CustomCardListView {
         this.resultTable = new ResultTable(this.customScreenDefinition);
         this.buttonActions = new ButtonActions(this.customScreenDefinition);
         const filterValues = new FilterValues();
-        filterValues.startDate = RealTimeDomainService.getCurrentDomain()?.startDate;
-        filterValues.endDate = RealTimeDomainService.getCurrentDomain()?.endDate;
+
+        if (this.customScreenDefinition?.initialBusinessPeriod === 'FROM_TODAY_TO_YEAR_END') {
+            const now = new Date();
+            const currentYear = now.getFullYear();
+            filterValues.startDate = new Date(currentYear, now.getMonth(), now.getDate()).getTime();
+            filterValues.endDate = new Date(currentYear, 11, 31, 23, 59, 59, 999).getTime();
+        } else {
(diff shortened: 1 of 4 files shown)


ui/main/src/app/views/customCardList/CustomCardListView.ts
```diff
@@ -92,7 +92,16 @@ export class CustomCardListView {
...
-        return RealTimeDomainService.getCurrentDomain();
+        if (this.customScreenDefinition.initialBusinessPeriod === 'FROM_TODAY_TO_YEAR_END') {
+            const now = new Date();
+            const currentYear = now.getFullYear();
+            return {
+                startDate: new Date(currentYear, now.getMonth(), now.getDate()).getTime(),
+                endDate: new Date(currentYear, 11, 31, 23, 59, 59, 999).getTime()
+            };
+        } else {
+            return RealTimeDomainService.getCurrentDomain();
```
freddidierRTE: When you do that , you get the business period from today to year end but you do not set the Realtime Domain which means you have the good period on the screen but the domain is stil the one from the feed . 

So you need to call `RealTimeDomainService.setStartAndEndPeriod` method with the computed period FROM_TODAY_TO_YEAR_END


But doing that inside the current get method is not good practice (get method should not modify internal state) 

So i propose to set the domain in the constructor of the CustomCardListView when we have FROM_TODAY_TO_YEAR_END 

And when we will call getBusinessPeriod , we just have to call ` RealTimeDomainService.getCurrentDomain();`

vlo-rte: Done.

freddidierRTE (approved):

PR #8616, opened 2025-07-10
Ag Grid headers disappearing with scroll (#8477)

### LINKED ISSUE #8477: Headers in ag grid disappearing with scroll
The ag grid in custom screen keeps the column names visible even when scrolling down. This should be the case for the tables in admin screen and logging

HubertDuchatelle:
nothing in release note


freddidierRTE (changes requested):
Avoid vertical scroll in logging screen

freddidierRTE (approved):

PR #8618, opened 2025-07-10
Export in Excel of a select type value from a custom screen (#8606)

### LINKED ISSUE #8606: Export in Excel of a select type value from a custom screen
Hi OperatorFabric colleagues,

I have the feeling that there is a bug when exporting to Excel a value/column containing a value from a "select" value type.

As a matter of fact, in the Excel file, the column is empty.

Have a nice rest of the day.

vlo-rte:
- In release notes :
  -  In chapter : Bugs
  -  Text : #8606 : Export in Excel of a select type value from a custom screen

### DIFF
--- a/ui/main/src/app/components/customCardList/cellRenderers/SelectCellRendererComponent.ts
+++ b/ui/main/src/app/components/customCardList/cellRenderers/SelectCellRendererComponent.ts
@@ -55,11 +55,9 @@ export class SelectCellRendererComponent implements ICellRendererAngularComp {
         if (this.fieldValue && !this.selectOptions.some((option) => option.value === this.fieldValue)) {
             this.selectOptions.push({value: this.fieldValue, label: this.fieldValue});
         }
-        this.fieldLabel =
-            params.getValue()?.possibleValues?.find((value: any) => value.value === this.fieldValue)?.label ??
-            this.fieldValue;
+        this.fieldLabel = params.getValue().text;
 
-        // Initialize control with current value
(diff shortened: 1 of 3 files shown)


ui/main/src/app/views/customCardList/resultTable/ResultTableExport.ts
```diff
@@ -67,6 +67,10 @@ export class ResultTableExport {
                             cellValue = cellValue?.rowValue;
                             row[column.headerName] = cellValue;
                             break;
+                        case 'select':
+                            cellValue = cellValue?.value;
+                            row[column.headerName] = cellValue;
+                            break;
```
freddidierRTE: We need to have the label if possible , so the idea is to iterate throught possible values to get label if exist .

The logic is already implemented in SelectCellRenderComponent : 


```
        this.fieldLabel =
            params.getValue()?.possibleValues?.find((value: any) => value.value === this.fieldValue)?.label ??
            this.fieldValue;
```
This solution will duplicate the logic , i guess it would be better when generating the result table to add a field text. So  the logic would be to implement it in TableRowBuilder (Line 128) 

So as it is will be compute when building the result table : 
-  it will not be needed to be compute in the SelectCellRenderComponent
-  for the export it will just be getting cellValue?.text as for default cell

vlo-rte: Done.

freddidierRTE (approved):

PR #8627, opened 2025-07-11
Numerical filtering/sorting in custom screens (#8603)

### LINKED ISSUE #8603: Numerical filtering/sorting in custom screens
Hi OperatorFabric colleagues,

Would it be possible to check if it is possible to make possible to filter/sort the columns when the column type is a numerical value?

Currently it looks like being only an alphabetic ordering.

Have a nice rest of the day.

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #8603 : Numerical filtering/sorting in custom screens

### DIFF
--- a/config/web-ui/externalJs/CustomScreenExample.js
+++ b/config/web-ui/externalJs/CustomScreenExample.js
@@ -285,6 +285,12 @@
                     multiLinesInCell: true
                 },
                 {
+                    field: 'startDate',
+                    headerName: 'START DATE AS EPOCH',
+                    cardField: 'startDate',
+                    fieldType: 'NUMBER'
+                },
+                {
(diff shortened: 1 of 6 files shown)


config/web-ui/externalJs/CustomScreenExample.js
```diff
@@ -284,6 +284,15 @@
...
                     multiLinesInCell: true
                 },
+                {
+                    field: 'startDate',
+                    headerName: 'START DATE',
+                    cardField: 'startDate',
+                    fieldType: 'NUMBER',
+                    getValue: (card) => {
+                        return getDateTime(card.startDate);
+                    },
```
freddidierRTE: remove the getValue method to use directly epochdate 
and change header name to "START DATE AS EPOCH"

freddidierRTE (approved):

PR #8787, opened 2025-08-27
Changed timeline navigation behavior (#8620)

### LINKED ISSUE #8620: Improve timeline navigation for past data
Currently, viewing past cards in "day mode" requires users to click the '<<' button day by day, which is inefficient for navigating deep into the past. We need a more intuitive and quicker method to access historical daily data.

**Current Behavior**

When a user switches between different time periods (e.g., from month to day view), the system resets the displayed time to the current date.

Example:

User selects "month" view.

User navigates back two months.

User clicks "day" view.

The system displays the current day, losing the previously selected historical context.

**Proposed Solution: Maintain Historical Context on Period Change**

We can enhance the user experience by not resetting the time to the current date when a user changes the period, allowing them to retain their historical context.

**Example of Proposed Behavior:**

User selects "month" view.

User navigates back two months.

User clicks "day" view.

The system displays the first day of the previously selected historical month, making it easy to then navigate to a specific day within that month.

This change would allow users to quickly navigate to a day far in the past by first selecting a broader historical period (like a month or year) and then switching to day mode.

**Detailed Implementation Notes:**

Longer Period to Shorter Period: When switching from a longer period (e.g., month) to a shorter period (e.g., day), the shorter period should start at the beginning of the longer period's selected timeframe.

Example: Switching from "month" to "day" should display the first day of the currently viewed month.

Shorter Period to Longer Period: When switching from a shorter period to a longer period, the system should select the longer period that encompasses the previously selected shorter period.

- **Special Cases:**

Switch from "7 Days" to "Week": The system should select the week that contains the start date of the previously viewed "7 Days" period.

Switch from "Week" to "7 Days": No change in displayed time is needed, as the "7 Days" view will naturally show the selected week's duration.

- **Exception: Real-time View**

This new behavior should not apply when switching to a "real-time" view. In this specific case, the system should continue to reset and display the current time, maintaining the expected "live data" experience.

- **Additional Note:**

Clicking the Same Period: If a user clicks on the period that is already selected (e.g., clicking "Day" when already in "Day" view), the system should reset the view to the current time.

ClementBouvierN:
- In release notes :
  -  In chapter : Features
  -  Text : #8620 Improve timeline navigation for past data

### DIFF
--- a/ui/main/src/app/services/realTimeDomain/RealTimeDomainService.ts
+++ b/ui/main/src/app/services/realTimeDomain/RealTimeDomainService.ts
@@ -34,9 +34,13 @@ export class RealTimeDomainService {
         if (UserPreferencesService.getPreference('opfab.timeLine.domain') === 'TR') {
             UserPreferencesService.setPreference('opfab.timeLine.domain', 'RT');
         }
+        // Needed for compatibility with versions prior to 4.10.0
+        if (UserPreferencesService.getPreference('opfab.timeLine.domain') === 'J') {
+            UserPreferencesService.setPreference('opfab.timeLine.domain', 'D');
+        }
         RealTimeDomainService.domains = ConfigService.getConfigValue('feed.timeline.domains', [
             'RT',
-            'J',
+            'D',
             '7D',
             'W',
             'M',
@@ -45,7 +49,7 @@ export class RealTimeDomainService {
 
         RealTimeDomainService.currentDomainId =
             UserPreferencesService.getPreference('opfab.timeLine.domain') ?? RealTimeDomainService.getDefaultDomainId();
-        RealTimeDomainService.setDefaultStartAndEndDomain();
+        RealTimeDomainService.computeStartAndEndDomainDates(RealTimeDomainService.currentDomainId, true);
         RealTimeDomainService.followClockTick = true;
     }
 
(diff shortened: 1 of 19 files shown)


ui/main/src/app/services/realTimeDomain/RealTimeDomainService.ts
```diff
@@ -58,9 +58,14 @@ export class RealTimeDomainService {
...
+        if (!isCurrentDomain) {
+            RealTimeDomainService.currentDomainId = domainId;
+        }
+        RealTimeDomainService.computeStartAndEndDomainDates(domainId, isCurrentDomain);
+
         if (!RealTimeDomainService.currentDomain || reset) {
-            RealTimeDomainService.setDefaultStartAndEndDomain();
+            RealTimeDomainService.computeStartAndEndDomainDates(RealTimeDomainService.currentDomainId, true);
         } else {
             RealTimeDomainService.updateCardFilter();
```
freddidierRTE: Line 67 : RealTimeDomainService.currentDomain cannot be undefined or null as it has been set the line before
Line 70 : i do not think we need to call updateCardFilter as computeStartAndEndDomainDates already calls it via setStartAndEndDomain

So i suggest a simplification : 

```
const anchorToNow  = reset ||  RealTimeDomainService.currentDomainId === domainId;
 RealTimeDomainService.currentDomainId = domainId;
 RealTimeDomainService.computeStartAndEndDomainDates(domainId, anchorToNow);
```


```

freddidierRTE (changes requested):
Add in migration guide the change to make in web-ui.json : J to change in D

freddidierRTE (approved):

PR #8827, opened 2025-09-08
Implement set filter in process monitoring (#8707) (with request to back)

### LINKED ISSUE #8707: Implement set filter in process monitoring
For each column in the process monitoring screen it should be possible to define a fiter with a set of possible value . The filter shall be accessible via the column header as for other filters.

The user shall be able to choose multiple value.

Set filter is available via ag-grid but in entreprise edition which is not compatible with our open source license : https://www.ag-grid.com/javascript-data-grid/filter-set/

But we can implement a similar filter using a custom component filter :

https://www.ag-grid.com/javascript-data-grid/component-filter/

For the configuration , the idea is to define a type of field "set" in the fields definition and add another field "possibleValues" containing an array ot possible values as string. 

Example : 

```
    "fields": [
      { "field": "startDate", "type": "date", "colName": "Start Date", "size": 200},
      { "field": "titleTranslated", "colName": "Title", "size": 250},
      { "field": "mySet" ,"colName":"mySet","type":"set","possibleValues":["val1","val2","val3"]}
]
````

For testing ,you can  use the defaultProcess/process card and configure a list of possible stateName 

Add in documentation 
Add a cypress test

### PR DESCRIPTION
**Pull request by vlo-rte**
- In release notes :
  -  In chapter : Features
  -  Text : #8707 : Implement set filter in process monitoring

### DIFF
--- a/ui/main/src/app/components/processmonitoring/components/processmonitoring-table/CustomFilterComponent.ts
+++ b/ui/main/src/app/components/processmonitoring/components/processmonitoring-table/CustomFilterComponent.ts
@@ -1,0 +1,65 @@
+/* Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {Component} from '@angular/core';
+import {FormsModule} from '@angular/forms';
+
+import type {IFilterDisplayAngularComp} from 'ag-grid-angular';
+import type {FilterDisplayParams} from 'ag-grid-community';
+import {NgForOf} from '@angular/common';
+
+@Component({
+    standalone: true,
+    imports: [FormsModule, NgForOf],
+    template: `
+        <div style="margin: 10px">
+            <label class="opfab-checkbox" *ngFor="let cb of checkboxes">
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: ui/main/src/app/components/processmonitoring/components/processmonitoring-table/CustomFilterComponent.ts [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+        <div style="margin: 10px">
+            <label class="opfab-checkbox" *ngFor="let cb of checkboxes">
+                <input type="checkbox" [(ngModel)]="cb.checked" (ngModelChange)="onCheckboxChange()" />
+                {{ cb.label }}
+                <span class="opfab-checkbox-checkmark"></span>
+            </label>
+        </div>
+    `
+})
+export class CustomFilterComponent implements IFilterDisplayAngularComp<any, any, string> {
```
freddidierRTE: I suggest to rename it "SetFilterComponent"

ui/main/src/app/components/processmonitoring/components/processmonitoring-table/CustomFilterComponent.ts
```diff
@@ -0,0 +1,65 @@
...
+    protected checkboxes: {label: string; checked: boolean}[] = [];
+    private possibleValues: string[] = [];
+
+    private filterParams: FilterDisplayParams<any, any, string>;
```
freddidierRTE: By settings `FilterDisplayParams<any,any,any>`

You can send  in `this.filterParams.onModelChange' the object you want and you do not need to use a string with separators 

So i propose to send on object like : 

```
{
type: 'set',
filter: ['value1','value2' ..] 
}
```

freddidierRTE (changes requested):

freddidierRTE (approved):

PR #8832, opened 2025-09-09
Make unchecked by default checkbox 'cards with response from my entities' for a custom screen (#8821)

### LINKED ISSUE #8821: Make configurable the by default state of a filter in a custom screen
Hi OpFab colleagues,

Would it be possible to make configurable the by default state of list filter of a custom screen?

As a matter of fact, we implemented this button which is enabled by default, and we would like to set it unabled by default:

Image

We remain available.

Have a nice day.

vlo-rte:
- In release notes :
  -  In chapter : Features
  -  Text : #8821 : Make unchecked by default checkbox 'cards with response from my entities' for a custom screen

### DIFF
--- a/ui/main/src/app/components/customCardList/CustomCardListComponent.ts
+++ b/ui/main/src/app/components/customCardList/CustomCardListComponent.ts
@@ -76,7 +76,7 @@ export class CustomCardListComponent implements OnInit, OnDestroy {
         processes: new FormControl([]),
         typeOfState: new FormControl([]),
         readAndAck: new FormControl([]),
-        responseFromMyEntities: new FormControl(true),
+        responseFromMyEntities: new FormControl(false),
         responseFromAllEntities: new FormControl(true)
     });
 
--- a/ui/main/src/app/components/customCardList/view/CustomCardListView.ts
+++ b/ui/main/src/app/components/customCardList/view/CustomCardListView.ts
@@ -56,8 +56,11 @@ export class CustomCardListView {
             filterValues.endDate = RealTimeDomainService.getCurrentDomain()?.endDate;
         }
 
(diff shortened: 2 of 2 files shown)


ui/main/src/app/components/customCardList/CustomCardListComponent.ts
```diff
@@ -76,7 +76,7 @@ export class CustomCardListComponent implements OnInit, OnDestroy {
         processes: new FormControl([]),
         typeOfState: new FormControl([]),
         readAndAck: new FormControl([]),
-        responseFromMyEntities: new FormControl(true),
+        responseFromMyEntities: new FormControl(false),
```
freddidierRTE, line 79: It sets the checkbox on the screen but it does not take it into account when going on the screen first as the request is done by default on init with value set to true .

To reproduce : 

- Remove from custom screen 1 definition the following line : 
```
-        // Possible values: `INPROGRESS`, `FINISHED`, `CANCELED`
-        defaultSelectedTypeOfState: [
-            'INPROGRESS',
-            'FINISHED'
-        ],
-        // Possible values: 'READ', 'NOT_READ', 'ACKNOWLEDGED', 'NOT_ACKNOWLEDGED'
-        defaultSelectedReadAndAck: [
-            'NOT_ACKNOWLEDGED',
-            'NOT_READ'
-        ],
```
- Send a usercard question and respond to it

The card are visible whereas it should not be : 

image

Solution proposal , in CustomCardListView init value to false if filter visible : 

```
        if (this.customScreenDefinition.headerFilters?.includes(HeaderFilter.RESPONSE_FROM_MY_ENTITIES))
            filterValues.includeCardsWithResponseFromMyEntities = false;
        else filterValues.includeCardsWithResponseFromMyEntities = true;
```

freddidierRTE (approved):

PR #8908, opened 2025-09-19
Use scroll bar in rich text editor for build in template message (#8709)

### LINKED ISSUE #8709: Use scroll bar in rich  text editor for build in template message
Add scrolling inside rich text editor instead of scrolling globally

See if we can use a scrolling for all rich text editor with a default height value that can be overrided locally

Image

vlo-rte:

### DIFF
--- a/ui/main/src/scss/styles.scss
+++ b/ui/main/src/scss/styles.scss
@@ -571,6 +571,7 @@ ngb-pagination {
 // QUILLJS STYLE
 .ql-snow .ql-editor {
     min-height: 100px;
+    max-height: 200px;
 }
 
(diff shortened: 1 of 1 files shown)


freddidierRTE:
Add documentation in migration guide to advice user to check their templates using rich text editor as we set a default fixed height

Add in documentation an explaination of how to extend the default height of the rich text editor (in https://opfab.github.io/documentation/current/reference_doc/#_rich_text_editor)

vlo-rte:
> advice user to check their templates using rich text editor as we set a default fixed height

Done.

freddidierRTE (approved):

PR #9416, opened 2025-12-24
Hide card reponse header if no entity allowed to respond (#9367)

### LINKED ISSUE #9367: Hide card reponse header if no entity allowed to respond

vlo-rte:
- In release notes :
  -  In chapter : Features 
  -  Text : #9367 : Hide card reponse header if no entity allowed to respond


frontend/src/app/components/card/components/card-header/CardHeaderComponent.html
```diff
@@ -47,6 +48,7 @@
         </span>
         }
     </div>
+    }
 </div>
```
freddidierRTE: Image

In case no state type and no entity allowed to respond , there is still a small line visible

freddidierRTE (changes requested):

frontend/src/app/components/card/components/card-header/CardHeaderComponent.html
vlo-rte: Done.

freddidierRTE (approved):

PR #9525, opened 2026-01-26
cards-external-diffusion : access cards directly via database (#9459)

### LINKED ISSUE #9459: cards-external-diffusion : access cards directly via database
In realTimeCardsDiffusion.ts

see line :           `  const cardContentResponse = await this.cardsExternalDiffusionOpfabServicesInterface.getCard(cardId);`

vlo-rte:
Nothing in release notes.

### DIFF
--- a/backend/cards-external-diffusion/src/domain/application/realTimeCardsDiffusionControl.ts
+++ b/backend/cards-external-diffusion/src/domain/application/realTimeCardsDiffusionControl.ts
@@ -142,7 +142,7 @@ export default class RealTimeCardsDiffusionControl extends CardsDiffusionControl
             const alreadySent = await this.wasCardsAlreadySentToUser(lightCard.uid, userEmail);
             if (alreadySent == null || !alreadySent) {
                 if (this.isSendingAllowed(userEmail)) {
-                    await this.sendMail(lightCard.id, userEmail, emailToPlainText, timezone);
+                    await this.sendMail(lightCard.id, lightCard.uid, userEmail, emailToPlainText, timezone);
                 } else {
                     this.logger.warn(
                         `Send rate limit reached for ${userEmail}, not sending mail for card ${lightCard.uid}`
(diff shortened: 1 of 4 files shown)


backend/cards-external-diffusion/src/domain/application/realTimeCardsDiffusionControl.ts
```diff
@@ -175,49 +175,51 @@ export default class RealTimeCardsDiffusionControl extends CardsDiffusionControl
...
         let cardConfig;
         let stateName;
 
         try {
-            const cardContentResponse = await this.cardsExternalDiffusionOpfabServicesInterface.getCard(cardId);
-
-            if (cardContentResponse.isValid()) {
-                const card = cardContentResponse.getData();
-                this.logger.info('Send Mail to ' + to + ' for card ' + card.uid);
+            const card = await this.cardsExternalDiffusionDatabaseService.getCard(cardUid);
```
freddidierRTE, line 192: Add a check if card is defined if not return and put a logger.info with the uid of the card (Example : "impossible to load card {cardUid} from database")

backend/cards-external-diffusion/src/domain/server-side/cardsExternaDiffusionDatabaseService.ts
```diff
@@ -121,4 +121,40 @@ export default class CardsExternalDiffusionDatabaseService {
         }
         return cards;
     }
+
+    public async getCard(cardUid: string): Promise<any> {
+        let card: any = {};
```
freddidierRTE: Set to undefined , so if there is a problem it will return undefined .

This will be process by the calling method (See previous comment)

freddidierRTE (changes requested):
You do not need anymore to set opfabServicesInterfaceStub.card in all tests as cards are only taken from database  , see this example of correction : https://github.com/opfab/operatorfabric-core/commit/b5466038d04f565b4a1e3bf9b7cd87a9be134702

backend/cards-external-diffusion/src/domain/server-side/cardsExternaDiffusionDatabaseService.ts
vlo-rte: Done.

backend/cards-external-diffusion/src/domain/application/realTimeCardsDiffusionControl.ts
vlo-rte: Done.

vlo-rte:
> opfabServicesInterfaceStub.card

Done.

freddidierRTE (approved):

PR #9545, opened 2026-01-29
Email to card : send attachment as txt in card data (#9516)

### LINKED ISSUE #9516: Email to card : send attachment as txt in card data
Add in example , the attachement content in field data.attachment

It should be possible to have multiple attachment

Adjust karate test

vlo-rte:
Nothing in release notes.

### DIFF
--- a/backend/email-to-card/src/domain/server-side/emailServer.ts
+++ b/backend/email-to-card/src/domain/server-side/emailServer.ts
@@ -10,8 +10,10 @@
 import {EmailServerInterface} from './emailServerInterface';
 import {Email} from '../application/email';
 import imaps from 'imap-simple';
-import {simpleParser} from 'mailparser';
+import {AddressObject, simpleParser} from 'mailparser';
 import {EmailServerConfig} from './emailServerConfig';
+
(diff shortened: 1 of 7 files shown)


backend/email-to-card/src/domain/server-side/emailServer.ts
```diff
@@ -28,63 +30,89 @@ export default class EmailServer implements EmailServerInterface {
...
+            parsed.text?.trim() || '<empty>',
+            this.extractTextAttachments(parsed.attachments ?? [])
+        );
+    }
 
-                if (!bodyPart) continue;
+    private extractTextAttachments(attachments: any[]): {filename: string; content: string}[] {
+        return attachments
+            .filter((att) => att.contentType === 'text/plain')
+            .filter((att) => {
```
freddidierRTE: In fact i want to extract all attachement as  string , it is not necessart text/plain  (could be for example xml , json ) 

So name the method  extractAttachment and remove filter contentType 


For now we suppose all attachement can be converted as string

vlo-rte: Done.

freddidierRTE (approved):

PR #9931, opened 2026-04-17
Enforce type in custom screen definition (#9930)

### LINKED ISSUE #9930: Enforce type in custom screen definition
When registering a custom screen , user need to define the type of custom screen (filed "type") , for example  : 
 
```
   const customScreenExample = {
        id: 'testId',
        name: 'testName',
        type: 'CARD_LIST',
        initialBusinessPeriod: 'FROM_TODAY_TO_YEAR_END',
        processIds: [
            'api_test',
            'cypress',
            'defaultProcess',
```

Today it is possible to load the custom screen without this value , but in this case the bubble display does not work on custom tiles in the dashboard

The issue is to enforce the setting of this value . 

If the value is not set , the customscreen shall not be loaded and an error should appear on the console  

Code should be add in CustomScreenService.ts

Add unit test 

Add in migration guide for next release 4.13

freddidierRTE:
- In release notes :
  -  In chapter : Features 
  -  Text : `#9930  Enforce type in custom screen definition `

### DIFF
--- a/frontend/src/app/services/customScreen/CustomScreenService.spec.ts
+++ b/frontend/src/app/services/customScreen/CustomScreenService.spec.ts
@@ -1,0 +1,88 @@
+/* Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * See AUTHORS.txt
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of the OperatorFabric project.
+ */
+
+import {CustomScreenService} from './CustomScreenService';
+import {ScreenDefinition, ScreenType} from './ScreenDefinition';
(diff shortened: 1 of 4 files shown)


frontend/src/app/services/customScreen/CustomScreenService.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+        );
+    });
+
+    it('should not add a custom screen definition when type is invalid and log an error', () => {
+        const errorSpy = spyOn(LoggerService, 'error');
+        const screenDef = {
+            id: 'screen3',
+            name: 'Screen Three',
+            type: 'INVALID_TYPE'
+        } as unknown as ScreenDefinition;
```
vlo-rte: What do you think about this solution : 

const invalidScreenDef: any = {
    id: 'screen3',
    name: 'Screen Three',
    type: 'INVALID_TYPE'
};

It would avoid the double cast.

vlo-rte (commented):
Maybe we could add a test to verify the behavior in case of duplicate screen ids (to avoid future regressions) ?

vlo-rte (approved):

Issue #10123, opened 2026-06-05
Dashboard : dashboard shall only filter on start and end date (and not publish date)


freddidierRTE:
The dashboard date filter take either card where chosen period overlap  [startDate, endDate] or when publish date is in chosen period 

It shoud NOT take into account publish Date

PR #10125, opened 2026-06-08
Dashboard : dashboard shall only filter on start and end date (and not publish date) (#10123)

freddidierRTE:
Not in release note yet 

Will be in a release note for a patch version


vlo-rte (commented):
To improve readability, I propose to create a constant : 

`const ONE_HOUR = 3600 * 1000;`

And then : 

```
const now = Date.now();
const periodStart = now - ONE_HOUR;
const periodEnd = now + ONE_HOUR;

```

vlo-rte:
This code is repeated several times : 

```
filteredLightCardStore.updateFilter(
    FilterType.BUSINESSANDPUBLISHDATE_FILTER,
    true,
    {
        start: periodStart,
        end: periodEnd
    }
);

```

I propose to create a function : 

```
function activateBusinessAndPublishDateFilter(start: number, end: number): void {
    filteredLightCardStore.updateFilter(
        FilterType.BUSINESSANDPUBLISHDATE_FILTER,
        true,
        { start, end }
    );
}
```

and then to call this new function.

vlo-rte:
This code is repeated several times : 

```
const result = await firstValueFrom(
    dashboardView.getDashboardPage().pipe(skip(1))
);

```
so I propose to create a function : 

```
async function getDashboardResult() {
    return firstValueFrom(
        dashboardView.getDashboardPage().pipe(skip(1))
    );
}
```

vlo-rte (changes requested):

vlo-rte (approved):

